PR #15, opened 2015-04-13
Fixes for nightly, switch to liquid

johannhof:
Basically the whole thing rewritten, since unfortunately for us Rust's io module got completely overhauled.

Cobalt now uses our very own liquid template engine, which is still very unstable but will do its job of filling in attributes in templates just fine. Working on adding more tags like a proper ``for`` loop.

Still much TODO, ideas are at the bottom of the readme for now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,4 @@
 /target
+/build
+
+.DS_Store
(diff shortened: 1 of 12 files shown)


tak1n:
Awesome :+1:

PR #25, opened 2015-10-21
Drop nightly requirement in README

skade:
cobalt builts quite fine with Rust 1.3.0, the documentation seems outdated.


tak1n:
thx

PR #29, opened 2015-11-06
Cleaning up

johannhof:
- Removes the old util copy_recursive_filter function in favor of a nice
  iterator approach using walkdir.
- Replaces some instances of to_string with to_owned for performance.
- Some more try!
- deny warnings :sunglasses:


tak1n:
awesome :zap: :grinning:

PR #32, opened 2015-11-07
replace unwrap() with expect(format!(...))

durka:
`unwrap()` produces panics with no explanation, and on some platforms there are no line numbers in the backtrace even in debug builds (rust-lang/rust#24346). `expect()` with an informative message is a much better user experience.

before:

    $ ag -Q 'unwrap(' src | wc -l
      15

after:

    $ ag -Q 'unwrap(' src | wc -l
       0

I made my best guess for an explanation at each error location, but they should be audited by someone who knows how the code works!


johannhof:
This looks useful in the short term, but most of these errors should be handled properly and converted to warnings instead.

For anything that is so critical that a crash is completely inevitable, we should create our own `CobaltError` and return that to the build function.

So +1 from me, but let's make an issue on revisiting `expect`/`unwrap` handling.

Thanks for the contribution :heart:

PR #33, opened 2015-11-07
Skip layout dotfiles

durka:
Don't try to read dotfiles in the layout directory. In my case these were vim swap files which don't contain valid UTF-8.

This PR is based on #32, not master, so I'll rebase after (if) that is merged.

### DIFF
--- a/src/cobalt.rs
+++ b/src/cobalt.rs
@@ -11,6 +11,25 @@ use walkdir::WalkDir;
 use document::Document;
 use error::Result;
 
+macro_rules! walker {
+    ($dir:expr) => {
+        WalkDir::new($dir)
+            .into_iter()
+            .filter_map(|e| e.ok())
+            .filter(|f| {
+                // skip directories
+                f.file_type().is_file()
+                &&
+                // don't copy hidden files
+                !f.path()
+                  .file_name()
+                  .and_then(|name| name.to_str())
+                  .unwrap_or(".")
+                  .starts_with(".")
+            })
+    }
+}
+
 /// The primary build function that tranforms a directory into a site
 pub fn build(source: &Path, dest: &Path, layout_str: &str, posts_str: &str) -> Result<()> {
     // TODO make configurable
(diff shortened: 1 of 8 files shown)


durka:
Rebased.

src/cobalt.rs
```diff
@@ -23,14 +23,16 @@ pub fn build(source: &Path, dest: &Path, layout_str: &str, posts_str: &str) -> i
     // filename -> text content to the layout map
     for entry in walker.filter_map(|e| e.ok()).filter(|e| e.file_type().is_file()) {
```
johannhof: we already apply a filter on the files here, maybe you can create a filter function at the top level that checks if it's a file and if it doesn't start with ".", instead. We could also use that function in other places then.

durka: Yeah, seems like there should be a utility function for creating `WalkDir`s with various filtering options.

johannhof:
@durka are you still working on this? :)

I'd love to get this merged soon, as it's prerequisite to fixing the problems from https://github.com/edunham/cobaltsite. I think we should just ignore dotfiles globally, in all walkdir filter functions.

If I can assist you in any way, let me know.

durka:
Yes. Currently thinking about how best to abstract this out in current Rust where you can't return unboxed iterators. Maybe a macro?

durka:
@johannhof see new commits. I did add a test so not sure about the coveralls error.

johannhof:
Interesting solution with the macro, I would've just defined a function that you can pass into `filter`, but that works for me as well.

Test coverage seems to be plummeting because no additional files are copied over into the target directory. If you'd put a random file into your test it should get copied and then that code will be triggered, it's also nice to test that this actually works. Could you take care of that?

Thanks! :+1:

durka:
Hmm, that didn't fix it...

durka:
Oh, it did fix it, but Github didn't show the new result! Ready to merge, then.

johannhof:
:+1: thanks so much for working on this

PR #38, opened 2015-11-07
Include test, example of special post.name attribute in tests

nathanross:
Commits:

* in ```tests/target/examples.html``` change test data to require each post title to be wrapped in a ```<a>``` tag linking to the post.

* in ```tests/fixtures/example/index.tpl```, in the code's map loop over posts, wrap each post title outputted in an ```<a>``` tag that makes use of the special ```{{post.name}}``` attribute in the value for its href attribute.

Reason:

* tests that post.name attribute works

* demonstrates to new users the usage of ```{{post.name}}``` for linking to posts in an index page.

### DIFF
--- a/tests/fixtures/example/index.tpl
+++ b/tests/fixtures/example/index.tpl
@@ -3,5 +3,5 @@
 This is my Index page!
 
 {% for post in posts %}
- {{ post.title }}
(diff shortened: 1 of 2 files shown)


johannhof:
neat :+1:

Issue #37, opened 2015-11-07
Support a configuration file


johannhof:
It would be cool to have a `cobalt.yml` or `cobalt.toml` (maybe yml is easier because our front-matter is yaml already) to set all options that can currently only be passed through the command line.

johannhof:
Closed by #43

PR #43, opened 2015-12-04
Added support for a config file

freiguy1:
This comes from a feature request in issue #37.  The config file the code looks for now is `config.yml` instead of the suggested `cobalt.yml` in the issue.  This was just an oversight on my part.  I'm not terribly experienced in the github fork/PR lifecycle and rust itself, but I'm learning! Any positive/negative criticism is encouraged.


johannhof:
Hey @freiguy1, thank you very much for the pull request! I'm currently traveling and will only be able to take a look at it mid next week. First impression of this is :+1: , but I'd like to give it a closer look before merging.

freiguy1:
Alright, no problem. A couple things since I've noticed since writing this:

- The `config.yml` file gets copied into the target directory which it shouldn't.
- At the moment the `config.yml` file needs to be in the current directory. There is no command line argument to change this.  Also, you can  set source directory in the `config.yml`. That means that your `config.yml` can live completely separately from your source directory. I think a better option would be to *not* have source set-able in `config.yml` - only in the command line arguments.  Then the config file should exist in the source directory.

Perhaps this can be discussion for later.

johannhof:
Finally got around looking at this, sorry.

I don't see anything wrong with setting `src` in the config.yml, I'd rather allow the `cobalt` executable to be passed a `--config` option, but that's for another PR.

One (cheap) way to solve the copying would be to make the file name` .cobalt.yml` (with a leading dot), which will be ignored by the generator.

We should go for `.cobalt.yml` instead of config.yml as default name, would you be up for changing that? :)

freiguy1:
Alright, I made that change. I'm not sure making the config a *hidden* file is appropriate, but it does get us up and running now.  It seems inconsistent with how `_layouts` and `_posts` are named where they aren't hidden, but instead the program knows to avoid copying these key folders wholesale.

johannhof:
Yeah you make a good point about the inconsistency. I think it's still ok though, I consider `_layouts` and `_posts` to be actual content that is "transformed" and `.config` is just meta-information for the engine.

I guess the biggest problem is just implicitly copying all static content from the build folder into the target folder. Not sure if we want to keep that in the long run, it's just what has worked for me so far.

johannhof:
Anyway, thanks for the contribution :heart:

johannhof:
@freiguy1 After merging the PR I actually noticed two small things I wanted to change, you can check out the changes at https://github.com/cobalt-org/cobalt.rs/compare/e586ac6fc0af5ffe3d15ce82bdf735367b8badb1...b3e6dd46cdab60af75392fbdb742c42da2c729ae

freiguy1:
Awesome, my pleasure! This is now my first PR and first merged PR.  I'm hoping to contribute more in the future.

PR #48, opened 2016-01-16
Sort posts and parse front matter as yaml

johannhof:
The primary goal of this is to sort posts by date (latest first). On that occasion I also changed the front-matter parsing to use yaml-rust instead of our own hacky parsing function.

This is a backwards-incompatible change, as `@extends` is an invalid
identifier in yaml (specifically the @ part). I changed it to `extends`,
which doesn't seem bad.

Posts are now sorted by date by default. Dates have to have the format
`%d %B %Y %H:%M:%S %z`, e.g. `29 May 2016 23:00:00 +0100` to be
successfully parsed. If no dates are provided or the parsing fails, it
falls back to 1970.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ liquid = "0.2"
 walkdir = "0.1"
 crossbeam = "0.1.5"
 yaml-rust = "0.2.2"
+chrono = "0.2"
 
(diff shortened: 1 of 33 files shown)


johannhof:
Let's go!

PR #52, opened 2016-02-02
Implement better logging (closes #39)

### LINKED ISSUE #39: Better logging
Maybe using [log](https://doc.rust-lang.org/log/log/index.html) or something even more sophisticated.

johannhof:
This commit implements logging using the log and env_logger crates. The
lib part of cobalt strictly relies on using only log! macros to be
friendly to usage by other consumers.

main.rs uses env_logger to set different logging levels based on command
line flags, not environment variables.

The default log level is info. Optional log levels are debug and trace.
Logging can be turned of completely by passing --silent.

Feel free to litter the code with trace! and occasional debug! calls.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,8 @@ walkdir = "0.1"
 crossbeam = "0.1.5"
 yaml-rust = "0.2.2"
 chrono = "0.2"
+log = "0.3"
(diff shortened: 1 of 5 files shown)


johannhof:
:rocket:

Issue #58, opened 2016-03-13
String-related panic when processing valid line of Markdown


ebkalderon:
> Note: I posted this first on the johannhof/markdown.rs#13. Not sure whether the bug is in cobalt or in the Markdown library, so reposting here.

The error message reported by the panic, when looked up on Google, led me to this [Rust book page on string slicing](https://doc.rust-lang.org/book/strings.html#slicing). Could you please look into this? This is a pretty major bug. Thanks!

**Update**: I'm using [cobalt 0.2.0](https://github.com/cobalt-org/cobalt.rs) with markdown 0.1.1 on rustc 1.9.0-nightly (eabfc160f) on 64-bit Arch Linux.

### Output

```
[info]   Building from blog into web
[info]   Created web/index.html
[info]   Created web/_posts/week0.html
thread '<unnamed>' panicked at 'index 42 and/or 43 in `strictly read-only operations, so they don’t fit into either *views* or` do not lie on character boundary', ../src/libcore/str/mod.rs:1670
note: Run with `RUST_BACKTRACE=1` for a backtrace.
thread '<main>' panicked at 'called `Result::unwrap()` on an `Err` value: Any', ../src/libcore/result.rs:746
[info]   Created web/_posts/week6.html
```

### Backtrace

```
stack backtrace:
   1:     0x55b8389646c0 - sys::backtrace::tracing::imp::write::h3e252ef592ef6f37cHv
   2:     0x55b83896757b - panicking::default_handler::_$u7b$$u7b$closure$u7d$$u7d$::closure.44999
   3:     0x55b8389671e8 - panicking::default_handler::h5a98c984225a759d1nA
   4:     0x55b838958bbc - sys_common::unwind::begin_unwind_inner::ha2071187f9cf0076vvu
   5:     0x55b838959008 - sys_common::unwind::begin_unwind_fmt::h65e28ed738ec1ab9Buu
   6:     0x55b838963b31 - rust_begin_unwind
   7:     0x55b83899e48f - panicking::panic_fmt::h98dcb863ad76e04b7mM
   8:     0x55b8389a5584 - str::slice_error_fail::h98e1db52ab0be957SIT
   9:     0x55b83890bbed - parser::span::parse_spans::hd76b1aa268612ed8qja
  10:     0x55b838907404 - parser::block::parse_blocks::h40207e955dc8e399kza
  11:     0x55b838902e54 - to_html::h178a80c20d1918a2kcb
[...]

PR #64, opened 2016-03-22
Switch to pulldown-cmark for markdown parsing

johannhof:
This commit removes the dependency on markdown.rs for now, as
pulldown_cmark offers more compatibility and stability. However,
depending on how the two libraries turn out, we might want to consider
switching back again. None of the alternatives for Rust markdown parsing
is in a great state, currently.


johannhof:
This fixes #58, cc @ebkalderon, thanks for reporting!

ebkalderon:
@johannhof Thank you so much! The blog articles build completely, with no panics at all. See below:

screenshot from 2016-03-22 19-46-51

I think this should be merged.

PR #68, opened 2016-03-30
Add basic serve command

### LINKED ISSUE #50: Add a serve(r) command
We should enable people to run a local website server with a command like `cobalt serve` or `cobalt server` or `cobalt start`

tak1n:
Unfortunately I'm not that confident in rust yet so no tests for that.

Also do we need a test for just a basic command like this?
We are just using nickel therefore nickel should have a good test suite and done :D

This closes #50

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ log = "0.3"
 env_logger = "0.3"
 rss = "0.3.1"
 pulldown-cmark = "0.0.7"
-nickel = "*"
+nickel = "0.7.3"
 
 [dev-dependencies]
 difference = "0.4"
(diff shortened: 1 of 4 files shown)


johannhof:
This should be in main.rs and not in cobalt.rs, as it's more a convenience functionality provided by our "frontend" and not part of the generator.

I'm good with no tests for this

tak1n:
Done

johannhof:
So we can make followup bugs for this, but there are two things that come to mind right now:

- `serve` isn't very useful as it is without `watch`, we should implement that
- the user should be able to specify the port to serve on

But yeah, let's make some followup issues for that

:+1:

PR #69, opened 2016-03-31
Deprecate tpl extension

tak1n:
Not sure what I thought when I started out with that .tpl templates.
We use liquid templates therefore the extension should be .liquid

In future maybe we can adapt more template languages where the extension would indicate which template engine to use. But for now liquid-rust is fine.


johannhof:
Technically, deprecate would mean we retain support but discourage usage. But considering people can set their own extensions in the config and that we're not stable yet I'm good with this I think.

johannhof:
Oh, please also update the README (and take a grep through the project for "tpl") to find if you missed anything.

tak1n:
@johannhof thought missed stuff breaks the test, which was not the case, missed some addtional tpl templates in fixtures folder.

Also removed it from the readme.

johannhof:
:+1:

PR #77, opened 2016-04-05
Added is_post gobal to allow users to detect if they are a post or not

LucioFranco:
Example usage:
```
{% if is_post %}
          {% include src/_layouts/post.liquid %}
{% else %}
          {{ content }}
{% endif %}
```

### DIFF
--- a/src/document.rs
+++ b/src/document.rs
@@ -124,6 +124,9 @@ impl Document {
         }
 
         data.set_val("content", Value::Str(html));
+        if self.is_post {
+            data.set_val("is_post", Value::Bool(true));
+        }
 
         let options: LiquidOptions = Default::default();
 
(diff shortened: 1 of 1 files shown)


johannhof:
Great idea! Two minor things:

- can you move this into the `get_attributes` function (line 58) so that we assign all attributes in a central place?
- please set the value to `false` if it's not a post :)

LucioFranco:
@johannhof should be fixed. Sorry about that must have been very tired last night.

tak1n:
@LucioFranco I think you forgot to add a else branch here: https://github.com/cobalt-org/cobalt.rs/blob/08245bf9e12c3f398f19bbeb5bfe87883a263f99/src/cobalt.rs#L72

If its not a post we need to set is_post to false.

LucioFranco:
@tak1n it is already set to `false` by default. https://github.com/cobalt-org/cobalt.rs/blob/08245bf9e12c3f398f19bbeb5bfe87883a263f99/src/cobalt.rs#L282

tak1n:
@LucioFranco another good point :smile: 
totally missed that

Issue #76, opened 2016-04-02
Add a --port option to cobalt serve


johannhof:
Where one can specify the port to serve on

LucioFranco:
Closed with #80

Issue #51, opened 2016-01-31
Add a watch option


johannhof:
It would be cool if Cobalt could watch the build directory for changes and rebuild if the `--watch` flag is passed.

LucioFranco:
Closed with #80

PR #80, opened 2016-04-06
Add Watch command

LucioFranco:
Adds watch command and other things:

TODO:
- [x] `Serve` command now builds the site before serving it
- [x] `Watch` command builds the site then serves it then waits to see if a file has changed in the source dir and will rebuild on change #51 
- [x] Add `--port` flag to pass a port to the server #76 
- [x] Simple usage message for `build`, `watch`, and `serve`

Let me know if there is anything else that should be added/changed.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -3,6 +3,7 @@
 extern crate cobalt;
 extern crate getopts;
 extern crate env_logger;
+extern crate notify;
 
 #[macro_use]
 extern crate nickel;
@@ -17,6 +18,10 @@ use cobalt::Config;
 use log::{LogRecord, LogLevelFilter};
 use env_logger::LogBuilder;
 use nickel::{Nickel, StaticFilesHandler};
+
+use notify::{RecommendedWatcher, Error, Watcher};
+use std::sync::mpsc::channel;
+use std::thread;
 
 fn print_version() {
     println!("0.2.0");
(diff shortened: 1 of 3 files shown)


LucioFranco:
@johannhof @tak1n would love if one of you could review this code. It's not fully done and I would love some suggestions for improvement. Thanks!

johannhof:
I'll take a look later today! :)

src/main.rs
```diff
@@ -117,35 +136,78 @@ fn main() {
...
+                    loop {
+                        match rx.recv() {
+                            _ => {
+                                info!("Rebuilding cobalt site...");
+                                build(&config);
+                            }
+                        }
+                    }
+                }
+                Err(e) => error!("[Notify Error]: {}", e),
```
johannhof: we should `std::process::exit(1)` here

src/main.rs
```diff
@@ -117,35 +136,78 @@ fn main() {
...
+}
+
+fn serve(dest: &str, port: &str) {
+    info!("Serving {} through static file server", dest);
+    let mut server = Nickel::new();
+
+    server.utilize(StaticFilesHandler::new(dest));
+
+    let ip = "127.0.0.1:".to_owned() + port;
+    server.listen(&*ip);
```
johannhof: is the `&*` necessary here?

src/main.rs
```diff
@@ -117,35 +136,78 @@ fn main() {
...
+        Err(e) => {
+            error!("{}", e);
+            error!("Build not successful");
+            std::process::exit(1);
+        }
+    };
+}
+
+fn serve(dest: &str, port: &str) {
+    info!("Serving {} through static file server", dest);
```
johannhof: maybe we should mention the port that it's being served on. I think you can also remove the "through static file server"

src/main.rs
```diff
@@ -117,35 +136,78 @@ fn main() {
...
+            thread::spawn(move || {
+                serve(&dest, &port);
+            });
+
+            let (tx, rx) = channel();
+            let w: Result<RecommendedWatcher, Error> = Watcher::new(tx);
+
+            match w {
+                Ok(mut watcher) => {
+                    // TODO: clean up this unwrap
```
johannhof: we should make a follow-up bug if you don't want to fix it right now (which is fine by me)

johannhof:
Luckily cobalt is quite quick right now so the just-rebuild-all approach will work for now, but we should definitely split up the different steps in cobalt.rs into public functions so that we can have at least a tiny bit of "incremental" compilation.

johannhof:
@LucioFranco this is awesome! It's definitely mergeable for me, I left a few minor nits for you anyway :)

src/main.rs
LucioFranco: I am going to make an issue for this but to also go into notify deeper and find the best way to use it. We may want to have a delay for example so that if you change a bunch of files you are not recompiling everytime.

src/main.rs
LucioFranco: Since we need to add the port the ip must be a string that is then converted back into a str because str has the ToSocketAddr trait. The other solution I found is:
```
let ip: &str = &("127.0.0.1:".to_owned() + port);
server.listen(ip);
```

There might be a better solution that uses the actual Ipv4 type but I think the current way is cleaner. Let me know what you think.

johannhof:
:+1:

PR #82, opened 2016-04-09
file_system is now passed to liquid

### LINKED ISSUE #74: Proper Relative Paths
We should handle relative paths better. @tak1n suggested that we use the `build` dir to be the base of the relative path. I think this would be good and allow us to pass that value to `liquid-rust` so that the template paths do not need to be very long.

Example:
@tak1n 
`cobalt build -s src/ -d /wherever/I/want -c .my_config.yml`
It should respect the source directory and search for the config file in src/
currently as far as I know you have to do following
`cobalt build -s src/ -d /what/what -c src/.my_config.yml`

LucioFranco:
This resolves #74.


johannhof:
👍

Issue #62, opened 2016-03-22
Invalid URLs when building blog posts on MSYS2


ebkalderon:
It seems that pages generated and placed in the `_posts` folder on Windows systems running the [MSYS2](https://github.com/msys2) toolchain are linked to using `\` instead of `/`.

See this example [website](http://aceeri.github.io/) by @Aceeri. As of 2016-03-21, the link to *This Week in Amethyst 6* is generated as:
```
http://aceeri.github.io/_posts\week6.html
```
when it should be:
```
http://aceeri.github.io/_posts/week6.html
```

johannhof:
That's probably because we use the Rust `Path` module for building relative paths and then convert them to URLs :confused:

Maybe the quickest solution would simply be to replace \ with / before passing the url to the template.

johannhof:
@ebkalderon Assuming we didn't miss anything this should be fixed through #83 now. Please reopen if any other problems arise.

ebkalderon:
@johannhof @LucioFranco Will do. Thank you both!

natejms:
It appears as though this issue might be affecting version 0.7.2 as well. Using Cobalt under a Linux environment runs flawlessly. However, every time I try to do so on Windows (Win10 in my case), a file named `index.html` in the `docs` directory, for an example, would become a file named `docs\index.html` (accessible from `docs%5Cindex.html`) at the root of the site.

epage:
@natejms Do you mind trying it against whats currently in git?  I've major changes in this area with some (hopefully) improved test coverage.  I'm curious what your results are.

Also is Rust building with msys or Visual Studio?

Some things that'd be helpful
- cobalt's output when run with `--verbose`
- Any relevant files from `--dump DocLinkObject`, and `--dump Document`.

natejms:
@epage Still no luck when building with what I got from Git ... I'm building with Visual Studio as far as I can tell.

I'm not sure which DocLinkObject and Document files would be relevant, but I do have everything on Github [at this link](https://github.com/natejms/cobalt-org.github.io/tree/master) (I've been testing with the cobalt-org website).

Upon further inspection, I can't reproduce the bug when running `cobalt build`. It seems like this only occurs when importing a build to another branch using `cobalt import` or `cobalt build --import`. Hopefully that helps a bit more.

epage:
[...]

PR #83, opened 2016-04-09
Fix for Windows path issue

LucioFranco:
This is a temporary fix for #62.


johannhof:
LGTM, but I'd appreciate a code comment in case anyone stumbles upon this in the future and doesn't know why we do this.

LucioFranco:
@johannhof added that comment. Will hopefully get to the RSS issue soon and then we can release `0.3`  🎉

johannhof:
Yay!

PR #85, opened 2016-04-13
Temp Fix for watched folders and Added RSS author attribute.

### LINKED ISSUE #63: Support all fields from the RSS spec
The RSS spec defines a lot of optional fields that we do not support right now. We should accept more arguments through both the config file (for page-wide fields) and in the article front-matter (for article fields).

See http://cyber.law.harvard.edu/rss/rss.html

### LINKED ISSUE #84: Watch command ignore destination dir
If you use the watch command in the same directory that you build to then it will recursively call itself forever.

LucioFranco:
Small temp fixes for path issues when the build folder is in the same dir as the source folder. 

Fixes #84 and Partly Fixes #63.

### DIFF
--- a/src/cobalt.rs
+++ b/src/cobalt.rs
@@ -128,8 +128,10 @@ pub fn build(config: &Config) -> Result<()> {
             !template_extensions.contains(&f.path()
                                             .extension()
                                             .unwrap_or(OsStr::new(""))) &&
-            f.path() != dest && f.path() != layouts_path.as_path()
+            f.path() != dest && !f.path().starts_with(Path::new("./").join(dest)) &&
+            f.path() != layouts_path.as_path()
         }) {
+            trace!("copying entry {:?}", entry);
             let entry_path = try!(entry.path()
                                        .to_str()
                                        .ok_or(format!("Cannot convert pathname {:?} to UTF-8",
--- a/src/main.rs
+++ b/src/main.rs
@@ -22,6 +22,7 @@ use nickel::{Nickel, Options as NickelOptions, StaticFilesHandler};
 use notify::{RecommendedWatcher, Error, Watcher};
 use std::sync::mpsc::channel;
 use std::thread;
+use std::path::PathBuf;
 
 fn print_version() {
     println!("0.2.0");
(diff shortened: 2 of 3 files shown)


src/cobalt.rs
```diff
@@ -128,8 +128,10 @@ pub fn build(config: &Config) -> Result<()> {
             !template_extensions.contains(&f.path()
                                             .extension()
                                             .unwrap_or(OsStr::new(""))) &&
-            f.path() != dest && f.path() != layouts_path.as_path()
+            f.path() != dest && !f.path().starts_with(Path::new("./").join(dest)) &&
```
johannhof: Not sure if this is solving the problem cleanly enough. From what I can gather you want to prevent copying things that are inside the /dest folder. I think a better way to do this is to replace the `filter` above with `filter_entry` from walkdir (see http://burntsushi.net/rustdoc/walkdir/#example-skip-hidden-files-and-directories-efficiently-on-unix)

src/main.rs
```diff
@@ -174,8 +175,27 @@ fn main() {
...
+
+                                if let Some(path) = val.path {
+                                    if path.is_absolute() {
+                                        // get where process was run from
+                                        let cwd = std::env::current_dir().unwrap_or(PathBuf::new());
+                                        // strip absolute path
+                                        let rel_path = path.strip_prefix(&cwd).unwrap_or(&cwd);
+
+                                        // check if path starts with the build folder.
+                                        if !rel_path.starts_with(&config.dest) {
```
johannhof: what if config.dest is an absolute path?

johannhof: also, the comment should be updated, you want to check if we're _not_ in the dest folder

src/main.rs
```diff
@@ -174,8 +175,27 @@ fn main() {
...
+                                            build(&config);
+                                        }
+
+                                    } else {
+                                        // check if path starts with build folder.
+                                        // TODO: may want to check if it starts `./`
+                                        if path.to_str() != Some(&config.dest) {
```
johannhof: does that work? isn't `val.path` something like `my_blog/dest/file.txt` and `config.dest` just `my_blog_/dest`?

johannhof:
I'm a bit unsure about the way you check if the watched file is not in the dest directory. An alternative would be to walk over `val.path` components and check any matches the dest.

What do you think? :)

LucioFranco:
@johannhof I'm thinking of using the stuff on #86 instead of the stuff in this PR. What do you think?

LucioFranco:
Closing this so that it can be implemented properly.

Issue #36, opened 2015-11-07
Add an option to set filetypes/directories to ignore


johannhof:
See #31, this would be actually be quite easy to implement.

We need to

- add an additional option flag in main.rs
- pass that to cobalt::build
- in cobalt::build extend the walkdir filter with a check for the filename

achernyak:
I can take this on. Should have a working solution in the next couple days.

LucioFranco:
@achernyak have you made any progress? 😄

achernyak:
@LucioFranco working through it now. Apologize for the delay, work got really crazy out of nowhere.

I will have a pr submitted this week.

johannhof:
Closed by #96

PR #86, opened 2016-04-13
Ignore Filepath

achernyak:
I am creating this PR as the initial implementation for #36 and to further the discussion. I have a few implementation detail questions when going through this.

1. Should we accept multiple arguments on the command line for the filepath?
1. Should we create a separate command for ignored directories, or handle them together? A separate command would be my choice since the implementation is pretty different for them.

Please also provide any code review you can, and I will incorporate it as I continue work on the directory structure.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ rss = "0.3.1"
 pulldown-cmark = "0.0.7"
 nickel = "0.8"
 notify = "^2.5.0"
+ghp = "0.1"
 
 [dev-dependencies]
 difference = "0.4"
--- a/src/main.rs
+++ b/src/main.rs
@@ -4,6 +4,7 @@ extern crate cobalt;
 extern crate getopts;
 extern crate env_logger;
 extern crate notify;
+extern crate ghp;
 
 #[macro_use]
 extern crate nickel;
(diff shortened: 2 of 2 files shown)


johannhof:
Thanks for the PR!

1. Yeah, I was thinking of a comma separated list maybe, and in the config an array.

like `--ignore .jpg,private_folder/`

2. We could consolidate them by simplifying the implementation to just making a simple `contains` check on each passed string. I think this is not an uncommon way for build systems to do it.

What do you think? :)

achernyak:
That sounds good to me. However, I think we need to force a specific way of inputting parameters. Otherwise, we could end up ignoring unintended things.

For example, someone wants to ignore the `jpg` folder and all they put in is `jpg`. If we were just checking if it is contained in the list, this would ignore the `jpg` folder but also all `jpg` files even outside the folder.

I think we should expose a single interface to the user with the restriction that all extensions must start with a `.`. Otherwise, they will be treated as directories. Then internally when we split it up into a config, we would have one config for extensions and another for directories.

Let me know your thoughts on this approach, or if I'm over engineering this :)

LucioFranco:
@achernyak I like that approach. We would need to make sure that this is documented very well. We should also figure out if there are folders we should ignore by default. For example we should ignore any `.git` directories and we should also ignore the build dest folder.

Great work!

PR #88, opened 2016-04-18
Wrap shared Vecs in Arc during document creation

johannhof:
No idea when they got lost, but cloning both Vecs for each iteration
obviously isn't a good idea.


tak1n:
https://github.com/cobalt-org/cobalt.rs/commit/2a5f41687c17d40562dd206668ca83fe1218b89e

johannhof:
Oh, I must've missed that, sorry. Anyway, cloning isn't a real solution, so I'd rather use an Arc until we figure how to do this properly.

The coverage failures must be some coveralls problem, not an issue with the PR...

tak1n:
No problem :)

PR #93, opened 2016-04-27
added upload command with ghp-rs

LucioFranco:
Last week I wrote a small gh-pages import library similar to ghp-import that was written in python. I have added a subcommand for `cobalt`. Let me know what you think @johannhof . I was also thinking it may be better to have it `cobalt build --upload` rather than `cobalt upload`.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -30,7 +30,7 @@ fn print_version() {
 }
 
 fn print_usage(opts: Options) {
-    let usage = concat!("\n\tbuild -- build the cobalt project at the source dir, -i --import will import the site to the gh-pages branch",
+    let usage = concat!("\n\tbuild -- build the cobalt project at the source dir",
                         "\n\tserve -- build and serve the cobalt project at the source dir",
                         "\n\twatch -- build, serve, and watch the project at the source dir",
                         "\n\timport -- moves the contents of the dest folder to the gh-pages branch");
--- a/src/main.rs
+++ b/src/main.rs
@@ -4,6 +4,7 @@ extern crate cobalt;
 extern crate getopts;
 extern crate env_logger;
 extern crate notify;
+extern crate ghp;
 
 #[macro_use]
 extern crate nickel;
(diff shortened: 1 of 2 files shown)


johannhof:
Really nice! But the name `upload` isn't quite fitting since it will not actually push the changes to the server, right?

LucioFranco:
@johannhof I agree. I am very open to any suggestions.

johannhof:
Mmmmh maybe `--gh-pages` or `--push-to-branch` or `--push-branch` or `--commit <branchname>`

florianjacob:
I've seen this being called `publish` (pelican) or `deploy` (nikola) in other static site generators.

LucioFranco:
@johannhof added some fixes. I went with `import` and `build --import` because this felt more natural to what it actually does. I want to let the user push to the correct repository for the moment. This also allows space for further expansion on the `import` command. Users can also use `-m <commit-message>` and `--message <commit-message>`. The same can be done for the branch option.

src/main.rs
```diff
@@ -28,9 +30,10 @@ fn print_version() {
 }
 
 fn print_usage(opts: Options) {
-    let usage = concat!("\n\tbuild -- build the cobalt project at the source dir",
+    let usage = concat!("\n\tbuild -- build the cobalt project at the source dir, -i --import will import the site to the gh-pages branch",
```
johannhof: why add the `--import` instruction here as well?

johannhof:
@LucioFranco looks great, I had one comment but good to merge otherwise!

src/main.rs
LucioFranco: @johannhof so you can run `cobalt build -b src -d build --import` and it will build then import. It would be useful for ci tools to do it all in one command.

johannhof: sorry, I meant you already documented it below, I think documenting it in that line is unnecessary :)

LucioFranco: hmmm I see what you are saying I will remove it then merge.

johannhof:
🎉

PR #95, opened 2016-05-07
Migrate from getopt to clap

jespino:
I propose this change because is easier to improve the command subcommands parameters using clap.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,7 +1,7 @@
 #![deny(warnings)]
 
 extern crate cobalt;
-extern crate getopts;
+extern crate clap;
 extern crate env_logger;
 extern crate notify;
 extern crate glob;
@@ -12,8 +12,7 @@ extern crate hyper;
 #[macro_use]
 extern crate log;
 
-use getopts::Options;
-use std::env;
+use clap::{Arg, App, SubCommand, AppSettings};
 use std::fs;
 use cobalt::Config;
 use log::{LogRecord, LogLevelFilter};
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -123,52 +171,40 @@ fn main() {
...
-        return;
-    };
-
-    // Check for port and set port variable to it
-    let port = matches.opt_str("port").unwrap_or("3000".to_owned());
-    let branch = matches.opt_str("branch").unwrap_or("gh-pages".to_owned());
-    let message = matches.opt_str("message").unwrap_or("cobalt site import".to_owned());
-    let should_import = matches.opt_present("import");
+    if matches.is_present("source") {
+        config.source = matches.value_of("source").unwrap().to_string();
```
LucioFranco: Could these `unwrap`'s be cleaned up?

LucioFranco:
I'm not too opposed. Might make creating a complex cli tool easier than getopts.

johannhof:
Mmh give me some more time to look at this. At first sight the two version seem equally complex, though that might be because I'm familiar with the original.

Thanks for the proposal anyway :)

jespino:
@LucioFranco: I fixed the unwraps.

@johannhof: Of course, take your time to eval clap, but the main reasons that made me prefer clap are, more high-level API, more robust checking and features integrated in the options (default values, multiple values, subcommands, automatic help generation, params errors management...).

LucioFranco:
@jespino You should be able to rebase to get the coverage fixed.

jespino:
@LucioFranco Thanks, done! :smiley:

LucioFranco:
@johannhof did we decide to not use clap? or should we take another look at it?

jespino:
Any news about it, I can rebase and fix conflicts if you decide to finally use clap.

flosse:
clap seems to be a good choice IMHO :)

LucioFranco:
I will leave this one up too @johannhof

johannhof:
Yeah I'd be fine with migrating to clap. @jespino you can update the PR and I'll take another look at it :)

jespino:
Of course, I will take a look as soon as possible :smile:

jespino:
Now is ready, the failing tests are because master is failing.

src/main.rs
```diff
@@ -135,52 +183,59 @@ fn main() {
...
-    let command = if !matches.free.is_empty() {
-        matches.free[0].clone()
-    } else {
-        print_usage(opts);
-        return;
-    };
+    match matches.subcommand() {
+        ("new", Some(submatches)) => {
+            match submatches.value_of("DIRECTORY") {
+                Some(directory) => config.dest = directory.to_string(),
```
LucioFranco: It is a bit confusing to use the `config.dest` variable for the destination of the new project. I would create a new variable like in the old code with `getopts`.

src/main.rs
```diff
@@ -135,52 +183,59 @@ fn main() {
...
+            }
+        },
 
-    match command.as_ref() {
-        "build" => {
+        ("build", Some(submatches)) => {
             build(&config);
-            if should_import {
+            if submatches.is_present("import") {
+                let branch = submatches.value_of("branch").unwrap().to_string();
```
LucioFranco: Will there be undefined behavior with these `unwrap`'s or is this something we can handle with a `unwrap_or`?

LucioFranco:
Left you some comments. The only issue I can see at the moment is the use of the `unwrap`'s. Great work @jespino!

src/main.rs
jespino: Yes, have sense, I can change it and put the default in the clap argument.

src/main.rs
jespino: This unwrap is because branch always have a value, the provided by the user, or the default one, so the unwrap must never panic.

LucioFranco: Okay then we can keep it. If it were to panic people can report it as an issue which is what it should be.

LucioFranco:
@jespino Looks like travis is failing because it requires a `rustfmt` of the code. If you can do that then I think we can merge with the consent of @johannhof 😄

jespino:
Passed the rustfmt to the code, the travis must pass now.

The comment about use own variable to destination on new subcomand is fixed too.

src/main.rs
```diff
@@ -32,63 +31,121 @@ use std::io::prelude::*;
...
-        return;
-    }
-
-    if matches.opt_present("version") {
-        println!("0.3.0");
-        return;
-    }
+    let matches = App::new("Cobalt")
+        .version("0.3.0")
+        .author("Benny Klotz <r3qnbenni@gmail.com>, Johann Hofmann")
```
johannhof: Not sure if I need the author information in `--help` tbh :)

johannhof:
So currently this throws the following error when I run e.g. `cobalt build --trace`:
```
error: Found argument '--trace' which wasn't expected, or isn't valid in this context
```

This applies to all other options as well.

Running `cobalt --trace build` works, but I'd love to be able to run `cobalt build --trace`. Is that fixable somehow? :)

Thanks! ❤️

jespino:
Can be done, but requires a bit of reestructure because the options can be "globally available" but in this case, the option is managed by the subcommand matches.

I can take a look to it today and try to solve it :smile:

jespino:
I think now works correctly, I have made the changes in other pull request to make easy to review the changes.

src/main.rs
```diff
@@ -135,52 +214,60 @@ fn main() {
...
+    config.source = matches.value_of("source")
+        .or(global_matches.value_of("source"))
+        .map(str::to_string)
+        .unwrap_or(config.source);
 
-    if let Some(dest) = matches.opt_str("d") {
-        config.dest = dest;
-    };
+    config.dest = matches.value_of("destination")
+        .or(global_matches.value_of("destionation"))
```
johannhof: destination

src/main.rs
```diff
@@ -32,91 +31,171 @@ use std::io::prelude::*;
...
-    if matches.opt_present("trace") {
-        builder.filter(None, LogLevelFilter::Trace);
-    }
 
-    if matches.opt_present("silent") {
-        builder.filter(None, LogLevelFilter::Off);
+    match matches.value_of("log-level").or(global_matches.value_of("log-level")) {
+        Some("error") => {
+            builder.filter(None, LogLevelFilter::Error);
+        }
```
johannhof: Nit: can you leave out the braces in this match block? Those are a bit cluttering for tons of single-line statements. (Unless this is because of rustfmt of course)

johannhof:
Thanks! This looks good apart from the two comments :)

src/main.rs
jespino: It's a mix of rustfmt and rust compiler, when I remove the braces, the compiler understand that the match branch returns a value, so y have to included before the match a "let _ = " (this is a valid approach), if I put the braces in the same line, the rustfmt reformat it to three lines. Which option you prefer? the current option or the "let _ =" option? Any other option?

jespino: Ok, not really needed, with a ; at the end of the match block is enough, I have changed it :smile:

johannhof:
🚀

LucioFranco:
👍

PR #96, opened 2016-05-07
Add ignore config

jespino:
I have added an extra config parameter to ignore files in the files walker
(this helps to exclude files to not publish things accidentally). This can help
to fix the watch infinite loop adding to ignore config the "*.html" glob (#84).
In the future I can add --ignore parameter or --ignore-file parameter to add
ignore patterns in the current run.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ pulldown-cmark = "0.0.7"
 nickel = "0.8"
 notify = "^2.5.0"
 ghp = "0.1"
+glob = "0.2.11"
 
 [dev-dependencies]
 difference = "0.4"
--- a/src/cobalt.rs
+++ b/src/cobalt.rs
@@ -15,9 +15,10 @@ use chrono::{DateTime, UTC, FixedOffset};
 use chrono::offset::TimeZone;
 use rss::{Channel, Rss};
 use std::sync::Arc;
+use glob::Pattern;
 
 macro_rules! walker {
-    ($dir:expr) => {
+    ($dir:expr, $ignore:expr) => {
         WalkDir::new($dir)
             .into_iter()
             .filter_map(|e| e.ok())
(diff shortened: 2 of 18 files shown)


LucioFranco:
This is awesome! I actually got it working with my skeleton site and fixed the issues with copying the build and .git folder. @johannhof I think we can somewhat merge this and add some default ignores like .git etc.

johannhof:
Looks great, thank you! We should think of a set of sensible defaults, such as hidden files + underscored folders now.

jespino:
I'm not sure which behavior is expected for default values. Now the dot files are ignored always, this behavior must be configurable? If I pass a possible future --ignore parameter with *.html override the current ignore or add to the default ignores? and for a --ignore-file parameter?

johannhof:
@jespino I would actually merge this right away, but we should probably also add an integration test in the `tests` folder. Could you do that? It's pretty easy. You add a cobalt source including config file to `fixtures` and the desired output to `target`. Then you add your test case to `mod.rs`.

Thank you!

jespino:
@johannhof of course, I will add the integration tests. :smile:

johannhof:
> I'm not sure which behavior is expected for default values. Now the dot files are ignored always, this behavior must be configurable? If I pass a possible future --ignore parameter with *.html override the current ignore or add to the default ignores? and for a --ignore-file parameter?

Yes that is a very good point. I'd like to defer that to a different PR/issue, but I would not ignore things like *.html by default and try to solve the infinite recursive watch issue in another way.

LucioFranco:
@johannhof which watch issue. #97 should fix the issue with the watch command. For building where the source is `.` we just need to ignore `./dest/*` and `.*`.

jespino:
I have added the integration test.

I agree to not include the *.html pattern to fix the recursive watch, was only a example of possible usage.

johannhof:
Great, thank you!

PR #97, opened 2016-05-09
Fix for watch command recursive call

LucioFranco:
This should fix the watch command recursively calling itself on mac. We will have to check with the notify crate to check how the other file notifiers notify what file is changed. This is a fix for #84.


LucioFranco:
@johannhof do you understand why all the PR's have a huge decrease in coverage? I think it may be a bug with kcov.

johannhof:
Cool, thanks for looking into this and the kcov thing! 👍

PR #105, opened 2016-05-12
sass support [Proposed]

LucioFranco:
This is just a proposal for sass support. It uses `sass-rs`. There are also tests. The biggest thing that I can think of that we need to think about is if we want to create this with a preprocessor way and possibly use features to allow people to install either via `cargo install cobalt.rs --features sass` or `cargo install cobalt.rs --features no-sass`. Let me know what you guys think. This code is very temporary right now but I want to start thinking about a better way to handle extra folders that need compiling done on the files inside.


flosse:
> This is just a proposal for sass support.

awesome!
> `install cobalt.rs --features no-sass`

I like this one. But I don't know if I generally prefer to enable features by default or the other way around. More important is to be consistent ;-)

LucioFranco:
I am closing this because I want to look into using this in a asset manager that can be included with compiler features.

PR #106, opened 2016-05-14
Add deployment and content list to README

LucioFranco:
I added few updates to the readme tell me how you guys like it. The rendered readme can be seen on my fork. https://github.com/LucioFranco/cobalt.rs

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,12 +6,19 @@
 
 A static site generator written in [Rust](http://www.rust-lang.org/).
 
-# Content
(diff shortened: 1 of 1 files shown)


johannhof:
Looks awesome! Although we should really think about putting up a website :)

LucioFranco:
@johannhof agreed! Should we think of a website for a `0.3` release? and also possibly get us on crates.io 😄

PR #107, opened 2016-05-15
Implement custom document paths (permalinks)

johannhof:
This implements what Jekyll calls permalinks, although I really hate that name. It also fixes a very interesting problem with fs::create_dir_all not being thread-safe.

Some things still need to be finished, like proper slugification. But
that should be for follow up bugs.

Note that this also removes the automatically injected variable `name`
which used to be the file name without the extension. I think it never
really made sense to have it.


johannhof:
@LucioFranco Can you give this a look? :)

johannhof:
Oh something that's also TODO would be to add the ability to set a default path format in the config.

README.md
```diff
@@ -75,6 +75,39 @@ You can specify different template extensions by setting the `template_extension
 template_extensions: ['txt', 'lqd']
 ```
 
+#### Custom paths
```
LucioFranco: I was thinking of `permalinks` instead of `paths`. What are your thoughts?

src/cobalt.rs
```diff
@@ -231,73 +232,38 @@ fn create_rss(path: &str, dest: &Path, config: &Config, documents: &[Document])
...
 
-    let markdown = path.extension().unwrap_or(OsStr::new("")) == OsStr::new("md");
+    let file_path = file_path_buf.as_path();
 
-    let name = try!(path.file_stem()
-                        .and_then(|stem| stem.to_str())
-                        .ok_or(format!("Invalid UTF-8 in file stem for {:?}", path)));
+    // create target directories if any exist
+    if let Some(parent) = file_path.parent() {
+        try!(create_dir_all(parent).map_err(|e| format!("Could not create {:?}: {}", parent, e)));
```
LucioFranco: Would it not be better to print the actual `std::io` message?

src/cobalt.rs
```diff
@@ -231,73 +232,38 @@ fn create_rss(path: &str, dest: &Path, config: &Config, documents: &[Document])
...
-
-    // construct path
-    let mut path_buf = PathBuf::from(new_path);
-    path_buf.set_extension("html");
-
-    let path_str = try!(path_buf.to_str()
-                                .ok_or(format!("Cannot convert pathname {:?} to UTF-8", path_str)));
+fn create_document_file<T: AsRef<Path>>(content: String, path: T, dest: &Path) -> Result<()> {
+    // construct target path
+    let mut file_path_buf = PathBuf::new();
```
LucioFranco: Couldn't you just `let mut file_path_buf = dest.join(path);`? Probably not a big deal though.

LucioFranco:
@johannhof this looks really good! I only have a few comments. One thing that we might want to do is have a default custom path that is applied to all the posts. That way one can set the base path from not being `_posts` to something cleaner. It would be cool if the default path could be joined with a custom path specified per post. So the default path for all posts is `/:year/` then in the per post they could specify the extra path like `:title` and it would create a `/:year/:title` format. I think that may be what you already have but that could solve a lot of issues. The rest looks very good and once we figure these out we are ready for a merge!

README.md
johannhof: As mentioned, I'm not a big fan of the term "permalink", what does that even mean?. It's a custom path that is different from the file path, that's why I called it "custom paths",

src/cobalt.rs
johannhof: It does that, it appends the io message to the other text. This way the user gets some info on what file actually failed to create.

johannhof:
@LucioFranco generally speaking, I'm not sure I want to transform all posts by default. I would much rather completely get rid of `_posts` as the default posts folder and replace it with something like `blog` or just nothing at all.

In general I think it's a good idea to stay really simple (put a file in `blog/my-post.md` and have it show up as `example.com/blog/my-post.html`) and then add the option have all of them automatically placed under some different url with an attribute in .cobalt.yml (which we will implement).

README.md
LucioFranco: The only reason I think we may want to also include `permalinks` is for compatibility with jekyll. But I do like the name custom paths better. I think we should include the word permalinks in the docs.

src/cobalt.rs
LucioFranco: You are right....I don't know why I said that.

LucioFranco:
@johannhof Sounds good. I think this is good to merge and if there are any follow up issues we will create an issue for it and follow up PR's.

PR #110, opened 2016-05-19
Fix build not successful on a new site

thesavage-me:
`cobalt build` using fresh install in a site just created using `cobalt new` would throw a "Liquid error: Parsing error: Expected Identifier, found Some(StringLiteral("_layouts/post.liquid"))" and cause a build to fail.

### DIFF
--- a/src/new.rs
+++ b/src/new.rs
@@ -20,7 +20,7 @@ const default_liquid: &'static [u8] = b"<!DOCTYPE html>
     <body>
     <div>
       {% if is_post %}
-        {% include '_layouts/post.liquid' %}
+        {% include _layouts/post.liquid %}
       {% else %}
         {{ content }}
       {% endif %}


LucioFranco:
Thanks for the PR! We actually want it to be a string literal.
From the liquid include tag in the original liquid templating uses a string literal.
>   # Include allows templates to relate with other templates
  #
  # Simply include another template:
  #
  #   {% include 'product' %}

The fix is actually already in `liquid-rust` but has not been released yet. I think the best temporary solution is to have it not pull `liquid-rust` from crates.io and just pull it from git and lock it to a commit. Unless @johannhof wants to release a `0.7.1` but I would rather just push for a `1.0` release instead. If you want to submit a PR that locks `liquid-rust` down to commit `9e8bc536b2e0809508b54d89d190631d564ec910` we can merge that one. I am going to close this as it does not fix the issue. Thanks again for looking into this!

thesavage-me:
Thanks for the quick response. I suppose shall just set that locally and wait for release then.

PR #112, opened 2016-05-20
prettify error message for unknown cli options

flosse:


LucioFranco:
I don't quite understand why changing this to just print the error message to `stdout` and exit with code `0`. Is better than a panic? I may be missing something 😄

flosse:
`Unrecognized option: 'V'.` is much nicer than ``thread '<main>' panicked at 'Unrecognized option: 'V'.', /home/mk/.cargo/git/checkouts/cobalt.rs-a8b94e0cf18c540b/master/src/main.rs:74
note: Run with 'RUST_BACKTRACE=1' for a backtrace.` But of course the exit code should be `1`.

LucioFranco:
@flosse that makes sense. This looks good to me but before we merge could you change it to output to `stderr` and not `stdout`.

src/main.rs
```diff
@@ -60,7 +61,11 @@ fn main() {
 
     let matches = match opts.parse(&args[1..]) {
         Ok(m) => m,
-        Err(f) => panic!(f.to_string()),
+        Err(e) => {
+          writeln!(&mut io::stderr(), "{}", e)
+              .expect("failed printing to stderr");
+          std::process::exit(1);
```
johannhof: please use the `error!` macro here (see code below)

johannhof:
Good idea :)

src/main.rs
flosse: hmm....using `error!` doesn't show any error information to the user :-(
So should we use `println!` followed by a `error!` statement?

LucioFranco: you can do
```rust
error!("Failed to parse options: {}", e);
```

flosse: hm....that doesn't work :( It just exits.

    $ ./target/debug/cobalt -fooo
    $

johannhof: oh right, because we didn't create the LogBuilder yet, duh, sorry for that. I guess your version is ok then :)

LucioFranco:
@flosse if you fix the merge conflicts we can merge this.

LucioFranco:
I am gonna close this. @flosse feel free to reopen if you want to finish this.

flosse:
I think this will be obsolete anyway as soon as #95 is merged ;-)

Issue #109, opened 2016-05-17
Move from Nickel to Hyper


LucioFranco:
Nickel is a huge library and it would be nicer to move to a more lightweight web server. We just need to serve files so we could just open the file and dump the bytes into the response. Let me know if you guys see any issues with this.

johannhof:
Yeah, I agree. It's not really a matter of binary size as the compilation should eliminate the code we're not using but rather the installation experience which involves installing all the nickel dependencies, and that's a drag.

PR #113, opened 2016-06-07
Move from Nickel to Hyper

DonRyuDragoni:
As proposed in #109, I moved all Nickel-related code to Hyper. I wasn't sure of the best (or this project's) way of handling each error, so I unwrapped everything.

I also couldn't find something similar to `StaticFilesHandler` in hyper, so I wrote a very basic handler that could be improved upon.


LucioFranco:
Thanks for the PR @DonRyuDragoni!

There are a few ways you can solve the error handling.

You can do:
```rust
match Server::http(&*ip) {
    Ok(handle) => handle(StaticFileServer),
    Err(e) => error!("{}", e),
};
```

I would also move the handler function to a separate function and handle the errors and call the `error!` macro inside the handler when it errors. 

```rust
handle(move |req:Request, res: Response| {
    if let Err(e) = handler(req, res) {
         error!("{}", e);
         std::process::exit(1);
    }
})
```

Also you are not handling the `/blog/engineer/` which would point to `/blog/engineer/index.html`.

DonRyuDragoni:
Ok, I'm applying the changes now.

Sorry I completely forgot about `/blog/engineer`. I re-coded the handler so that, if the requested path does not correspond to a file (according to [`Path::is_file()`][path_is_file]), it will try to find an `index.html` in the same place. Is this behavior acceptable?

[path_is_file]: https://doc.rust-lang.org/std/path/struct.Path.html#method.is_file

src/main.rs
```diff
@@ -267,17 +268,77 @@ fn build(config: &Config) {
...
+    let http_server = match Server::http(&*ip) {
+        Ok(server) => server,
+        Err(e) => {
+            error!("{}", e);
+            std::process::exit(1);
+        }
+    };
+
+    // need a clone because of closure's lifetime
+    let dest_clone = dest.to_owned();
```
LucioFranco: This is a nitpick but could you change this to `.clone()`.

src/main.rs
```diff
@@ -267,17 +268,77 @@ fn build(config: &Config) {
     };
 }
 
+fn static_file_handler(dest: &str, req: Request, res: Response) {
+    let path = PathBuf::from(format!("{}{}", dest, req.uri));
```
LucioFranco: In this situation format is a pretty heavy call. You should change it to something like.

```rust
```

DonRyuDragoni: I was doing exactly this, until I realized it would not work on my machine because `join` was replacing the whole path with `req.uri` rather than appending it. Now that I think about it, my solution was really terrible too. 

I'll experiment a bit here to see how can I grab the string directly.

src/main.rs
DonRyuDragoni: Sure thing. I just didn't know `str`s had a clone method.

LucioFranco: You are right.

src/main.rs
```diff
@@ -267,17 +268,77 @@ fn build(config: &Config) {
...
+
+    let serve_path = if path.is_file() {
+        // try to point the serve path to `path` if it corresponds to a file
+        path.clone()
```
LucioFranco: I believe you do not need to clone here. You should be able just have it return `path` not `path.clone()`.

src/main.rs
```diff
@@ -267,17 +268,77 @@ fn build(config: &Config) {
...
+        };
+
+        match res.send(&buffer) {
+            Err(e) => {
+                error!("{}", e);
+                std::process::exit(1);
+            },
+            Ok(()) => {},
+        };
+    }
```
LucioFranco: You need to handle the `else` case here. I believe it should return a `404` error when it can not find the file.

src/main.rs
```diff
@@ -267,17 +268,77 @@ fn build(config: &Config) {
...
+            Ok(f) => f,
+            Err(e) => {
+                error!("{}", e);
+                std::process::exit(1);
+            }
+        };
+
+        let mut buffer: Vec<u8> = vec![];
+
+        match file.read_to_end(&mut buffer) {
```
LucioFranco: You can clean this up a bit with an `if let` as you are only checking one case.

src/main.rs
```diff
@@ -267,17 +268,77 @@ fn build(config: &Config) {
...
+            Err(e) => {
+                error!("{}", e);
+                std::process::exit(1);
+            },
+            Ok(_) => {},
+        };
+
+        match res.send(&buffer) {
```
LucioFranco: Again here you should be able to do the same. It will make the code a lot nicer! 😄

LucioFranco:
@DonRyuDragoni Left you a few comments. Once you get those fixed and we figure out this weird coverage issue we can merge! Thanks again!

src/main.rs
DonRyuDragoni: I thought I needed to clone, but, yes, it can just return `path`.

src/main.rs
DonRyuDragoni: I'm trying to set the response status to 404 (or [`StatusCode::NotFound`](http://hyper.rs/hyper/v0.9.1/hyper/status/enum.StatusCode.html)) with the `Response::set_status()`, but it appears to not exist, even if a quick search in the repo tells me [it is supposed to](https://github.com/hyperium/hyper/blob/master/examples/server.rs#L116) ([source](https://github.com/hyperium/hyper/blob/ffb499bdf90a7e7ca441e0f6186c0363070a2e5d/src/server/response.rs#L40)). It apparently is also not documented.

For now, I'm just sending a simple 404 page. I'll dig further into it after I finish these changes.

LucioFranco: Have you taken a look at the response page? http://hyper.rs/hyper/v0.9.1/hyper/server/response/struct.Response.html

There seems to be as function called `status_mut` which should return a mutable reference to the status code. You can then just reassign it.

src/main.rs
DonRyuDragoni: After some digging, I did this:

```rust
fn static_file_handler(dest: &str, req: Request, res: Response) {
    let req_path = match req.uri {
        RequestUri::AbsolutePath(p) => p,
        _ => {
            error!("Request is not of the format AbsolutePath(p)");
            std::process::exit(1);
        },
    };

    let path = PathBuf::from(dest).join(&req_path[1..]);

    // ...
}
```

I tried many times, with paths that should fail and with correct paths, but all of them fell into `AbsolutePath(p)`. I'm still unsure of how to deal with the other cases, though.

Also, instead of relying in `format!`, this rids `p` of its first character (in all the tests I made, `p` started with `/`, and that was causing `join` to think `p` was an absolute path, so, instead of appending, it replaced the contents of `path` with `p`), and relies on `join` to get the local path to the file.

johannhof:
You're sometimes using `println!` and sometimes `error!` to communicate error messages. What's keeping you from using only `error!`?

src/main.rs
```diff
@@ -267,17 +269,100 @@ fn build(config: &Config) {
...
+    let http_server = match Server::http(&*ip) {
+        Ok(server) => server,
+        Err(e) => {
+            error!("{}", e);
+            std::process::exit(1);
+        }
+    };
+
+    // need a clone because of closure's lifetime
+    // let dest_clone = dest.to_owned();
```
johannhof: nit: you can remove that comment :)

src/main.rs
```diff
@@ -267,17 +269,100 @@ fn build(config: &Config) {
...
+    let dest_clone = dest.to_owned();
+
+    // bind the handle function and start serving
+    match http_server.handle(move |req: Request, res: Response| {
+        static_file_handler(dest_clone.as_str(), req, res);
```
johannhof: can't you simply pass `&dest_clone` here?

src/main.rs
```diff
@@ -267,17 +269,100 @@ fn build(config: &Config) {
     };
 }
 
+fn static_file_handler(dest: &str, req: Request, mut res: Response) {
+    // grab the requested path
+    let req_path = match req.uri {
+        RequestUri::AbsolutePath(p) => p,
+        _ => {
+            println!("Requset is not of the format AbsolutePath(p)");
+            std::process::exit(1);
```
johannhof: we should return a friendly 400 here instead of crashing the server

src/main.rs
```diff
@@ -267,17 +269,100 @@ fn build(config: &Config) {
...
+        // return a 404 status
+        *res.status_mut() = hyper::status::StatusCode::NotFound;
+
+        // write a simple body for the 404 page
+        let body = b"<h1> <center> 404: Page not found </center> </h1>";
+
+        if let Err(e) = res.send(body) {
+            // error!("{}", e);
+            println!("{}", e);
+            std::process::exit(1);
```
johannhof: You do this quite a lot, it's getting pretty verbose. What you could do is just return a `Result<(), String>` from `static_file_handler` and handle the error case in your closure. Then you could simply do

```rust
try!(res.send(body));
```
or

```
try!(file.read_to_end(&mut buffer));
```

which reads much nicer.

johannhof:
Maybe try another rebase for getting the coverage right :)

johannhof:
@DonRyuDragoni btw, thanks so much for doing this. I just had to do another clean install and the amount of dependencies that come with Nickel drive me crazy. Great that someone is tackling this now. 👍

DonRyuDragoni:
Sorry, I messed up!

I built a small program to test some things from hyper and ended up coding the hole handler there. So, after I change the handler there, I move them to this project files, test, and commit them. I was in a hurry because I had a medical appointment ~10 minutes after that push and was nearly finished, so I completely forgot to change the `println!`s back to `error!`s. 

Again, I'm really sorry. I just got home and I'll have a calmer look at it.

src/main.rs
DonRyuDragoni: No idea how this ended there...

src/main.rs
DonRyuDragoni: Indeed I can. Fixing.

DonRyuDragoni:
@johannhof its nothing, really. I was looking for something I could do to try and exercise my rusty programming skills (really, no pun intended), and it just happened that I played a bit with Hyper a few days ago, so I though I could help a bit.

Anyway, I'm finishing revising the corrections here just to be sure I won't accidentally push some test code again.

I know that the error handling in the `serve` function looks a bit ugly, but, after googling for a while, I could not find a better way to do it. Is there anything that can be done there?

johannhof:
> Again, I'm really sorry.

Don't worry, everything's good. That's what code review is for. :)

> I know that the error handling in the serve function looks a bit ugly, but, after googling for a while, I could not find a better way to do it. Is there anything that can be done there?

Seems fine to me. It's straightforward and goes with what we did in the other parts of the main function.

The patch looks great now, thanks a lot!

LucioFranco:
:tada:  Good work!

PR #115, opened 2016-06-12
Liquid date filter

kstep:
The PR contains one feature and one fix. I can split it up into two PRs if you like, but I feel like the fix is too small for a separate PR.

The feature is a new `date` filter so that one can do `post.date | date:"%d/%m/%y"` in liquid templates (example usage: https://github.com/kstep/kstep.github.com/blob/master/src/index.liquid#L17). Although it seems like liquid parser contains another bug: it doesn't parse filter arguments correctly, so I can't use spaces in the `date` filter arguments (I worked it around by using `&nbsp;` entity instead).

The fix is the incorrect YAML header splitting from markdown files. It splits files by all instances of `---` separator, while it should use the first one only. The fix is pretty obvious and self-explanatory.

---
This change is [Reviewable](https://reviewable.io/reviews/cobalt-org/cobalt.rs/115)

### DIFF
--- a/src/document.rs
+++ b/src/document.rs
@@ -115,7 +115,7 @@ impl Document {
         // TODO: make this a regex to support lines of any length
         if content.contains("---") {
             let content2 = content.clone();
-            let mut content_splits = content2.split("---");
+            let mut content_splits = content2.splitn(2, "---");
 
             // above the split are the attributes
             let attribute_string = content_splits.next().unwrap_or("");


johannhof:
Hi there! Thank you for helping out! The `date` filter is actually part of the default set of filters in Liquid, we're just missing it in liquid-rust right now. It would be great if you could open another PR against https://github.com/cobalt-org/liquid-rust/, see also https://github.com/cobalt-org/liquid-rust/issues/11

Please remove the date filter from this PR as well, so that we can merge the small fix.

As for the filter parsing issue, can you please open an issue in liquid-rust with a really small example of what does not work?

Thanks 👍

kstep:
Ok, I removed the commit with the date filter. I will prepare an example for the bug and post a bug report later today. As for the date filter, I need it now, so I will keep on using my branch of cobalt for my blog. I will repeat the filter implementation for liquid repo later this week.

kstep:
See cobalt-org/liquid-rust#44 for the bug report.

kstep:
See cobalt-org/liquid-rust#42 for date filter implementation in liquid source tree.

johannhof:
👍

PR #118, opened 2016-06-21
Refactoring Document, Rustfmt

johannhof:
This solves some potential issues and code mess in document.rs, e.g.
having duplicated calls to get_attributes() and setting attributes in
different places.

I also did a Rustfmt.


johannhof:
@LucioFranco can you take a quick look and merge if it looks ok to you? :)

src/cobalt.rs
```diff
@@ -158,7 +154,7 @@ pub fn build(config: &Config) -> Result<()> {
                 }
 
                 try!(fs::copy(entry.path(), &dest.join(relative))
-                         .map_err(|_| format!("Could not copy {:?}", entry.path())));
+                    .map_err(|_| format!("Could not copy {:?}", entry.path())));
```
LucioFranco: I know you didn't change this in this PR but should we not tell the user what the error was. Like permissions?

johannhof: you're right, I'll take a look

johannhof:
@LucioFranco updated

PR #119, opened 2016-07-03
Ignore underscored directories by default & add a "post_path" attribute

johannhof:
**tl;dr** `_posts` are `posts` now and everything with an underscore is considered meta-data

This is something I've been meaning to do for a long time now.

I would like to add consistency in how we deal with files and folders
that have a leading dot or underscore. This commit makes it so that both
kinds are simply ignored by default when looking for documents to
generate or assets to copy. I like to see those filenames as
indication that the folders contain "meta" information.

My idea of a good site generator is that your source file structure is
not fundamentally different from the resulting file structure. You
basically end up with the same structure but "enhanced" by the site
generator (e.g. by turning .md and .liquid files into .html files).
It is very useful to have an easy indicator like an underscore to show
that this file will be used in the generation process but will not be
there in the resulting site. You can now simply have a file called _data.json
next to your template or blog file.

The default post path has been changed to "posts".
For compatibility it is still possible to set your default post path to
"_posts", however, you now have to do it explicitly in your .cobalt.yml.

This is good, as having _posts/blog-post.md generate to
_posts/blog-post.html was just laughable.

I will make PRs for that in all affected user pages.
(Most of them already specify a different posts directory, so nothing should change).

_layouts obviously stays the same. I changed the layout handling a
bit, though. Layouts are now read from disk every time they are
required. This needs a bit of optmization in the future but allows for
better and more flexible layout specification.

Furthermore, users can now specify a `post_path` in their .cobalt.yml, which is the global
fallback for the `path` attribute in post front matter (called permalink in Jekyll). I feel like these
two changes finally give us both an easy but also flexible post writing experience.


johannhof:
@LucioFranco @tak1n Wanna take a look? It looks more than it is, because all tests had to change. :)

tak1n:
@johannhof from just reading the code everything looks good for me :)

will test it a little bit when I'm home

:dancers:

LucioFranco:
@johannhof I reviewed the code and it looks good. I also built my personal site with it and it seems to work. I think we should merge.

johannhof:
Thanks folks!

LucioFranco:
Great work @johannhof. hopefully I will find sometime soon to help out a bit.

PR #132, opened 2016-08-22
Add basic CLI tests

johannhof:
This adds a CLI testing mechanism around [assert_cli](https://github.com/killercup/assert_cli) that we should use to add more CLI tests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,8 @@ version = "0.9"
 default-features = false
 
 [dev-dependencies]
+assert_cli = "0.3"
+lazy_static = "0.1"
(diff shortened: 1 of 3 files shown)


LucioFranco:
The tests are failing because there is a bug in rustc beta and nightly right now. Also it seems you pushed this commit already. https://github.com/cobalt-org/cobalt.rs/commit/5ffe5803eab04b299047cecfca3571b68032f009

johannhof:
Haha wtf I'll back this out and try to fix the PR, thanks for noticing!

johannhof:
Ok I assume I can merge this then :)

PR #135, opened 2016-09-09
Processor refactor [Not Ready]

LucioFranco:
As I started to implement the `sass` feature I kept running into a lot of code bloat. The code in `cobalt::build` kept growing and would soon get too big to handle. So I thought of this remedy. My idea is to create a `Processor` trait that defines how different components interact with the cobalt site.

The trait is:

``` Rust
pub trait Processor {
    fn match_dir(&mut self, entry: &DirEntry, config: &Config) -> bool;

    fn process(&mut self, entry: DirEntry, config: &Config) -> Result<()>;
}
```

@johannhof I am curious what you think of this. The tests will fail but it compiles. I want to know if I should keep going forward with this as it will be some work to get the tests to pass as its a huge refactor.


johannhof:
Hey @LucioFranco, sorry for the very delayed answer on this. I'm actually not sure if we should include more preprocessors via Rust code, or if we should rather create a tooling integration that could work like this:

.cobalt.yml:
```yaml
processors:
   - pattern: style/*.sass
     command: "sass --compass"
   - pattern: scripts/**/*.coffee
     command: "coffee --stdio"
```

Where command is any tool that takes stdin and writes to stdout.

We could also utilize something like hooks in addition to that:

.cobalt.yml:
```yaml
processors:
   - pattern: style/*.sass
     command: "sass --compass"
startup: "export PATH=$PATH:/bin/coffee",
after_build: "coffee scripts/**/*.coffee -c -o build/scripts/bundle.js"
```

**Pros:**
- Much more flexible
- Less code to maintain for us
- We suddenly support everyone's favorite tool
- We don't have to think about a processor trait ;)

**Cons:**
- Requires more setup + installing the right components
- We need to document this well
- Huge speed loss compared to integrating native Rust code

What do you think?

LucioFranco:
@johannhof that's an interesting idea. One thing that bothers me with this idea is the extra installing and the speed loss which are two things that are very important concepts to the project. To me, it would over complicate it and while is very very flexible I don't think it should be a major part. I think it would be a good idea to have it in the future but not for the plugin/processor system.

johannhof:
@LucioFranco I understand your point about the installing. I think there are certain much-wanted components we have to identify and include by default. One could be Sass. (It would be great if we could support a Rust-based Sass compiler).

PR #144, opened 2016-10-08
watch should never rebuild on changes that are ignored

nott:
There was a bug when `path` contained `./` as prefix (e.g. `./build/index.html`). Now `path::PathBuf` takes care of corner cases like this.

The bug described above meant there was an infinite rebuild loop in `cobalt watch` even when `build` directory was marked as ignored:

* on each rebuild some files in `build` directory were updated
* `inotify` started new rebuilds on these updates

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -290,16 +290,13 @@ fn main() {
                                     // get where process was run from
                                     let cwd = std::env::current_dir().unwrap_or(PathBuf::new());
 
-                                    // the final goal is to have a relative path
-                                    // without corner-case prefixes like "./"
-                                    let abs_path = match path.is_absolute() {
-                                        true => path.clone(),
-                                        false => {
-                                            // let path::Path handle corner cases like `./rel/path`
-                                            let mut abs_path = cwd.clone();
-                                            abs_path.push(path.clone());
-                                            abs_path
-                                        }
+                                    // The final goal is to have a relative path. If we already
+                                    // have a relative path, we still convert it to an abs path
+                                    // first to handle prefix "./" correctly.
+                                    let abs_path = if path.is_absolute() {
+                                        path.clone()
+                                    } else {
+                                        cwd.join(&path)
                                     };
                                     let rel_path = abs_path.strip_prefix(&cwd).unwrap_or(&path);
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -287,28 +287,29 @@ fn main() {
                             Ok(val) => {
                                 trace!("file changed {:?}", val);
                                 if let Some(path) = val.path {
-                                    if path.is_absolute() {
-                                        // get where process was run from
-                                        let cwd = std::env::current_dir().unwrap_or(PathBuf::new());
-                                        // strip absolute path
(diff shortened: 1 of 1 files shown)


nott:
I don't have ideas for proper unit tests here. Do we need to add some?

johannhof:
Hey there! Thanks for fixing this! I'll take a look later today. Testing sounds tricky, even with our new [CLI tests](https://github.com/cobalt-org/cobalt.rs/blob/master/tests/cli.rs). So I don't mind missing tests here :)

src/main.rs
```diff
@@ -287,28 +287,29 @@ fn main() {
...
-                                                Pattern::matches_path(pattern, rel_path)
-                                            });
-                                        if !path_starts_with_build {
-                                            build(&config);
+                                    // get where process was run from
+                                    let cwd = std::env::current_dir().unwrap_or(PathBuf::new());
+
+                                    // the final goal is to have a relative path
+                                    // without corner-case prefixes like "./"
+                                    let abs_path = match path.is_absolute() {
```
johannhof: You can probably replace this with

```rust
let abs_path = if path.is_absolute() {
  ...
} else {
  ...
}
```

src/main.rs
```diff
@@ -287,28 +287,29 @@ fn main() {
...
+                                        true => path.clone(),
```
johannhof: Why do you have to clone here?

src/main.rs
```diff
@@ -287,28 +287,29 @@ fn main() {
...
+                                        false => {
+                                            // let path::Path handle corner cases like `./rel/path`
+                                            let mut abs_path = cwd.clone();
+                                            abs_path.push(path.clone());
+                                            abs_path
```
johannhof: I think this whole block can be cut to `cwd.join(path)`

src/main.rs
```diff
@@ -287,28 +287,29 @@ fn main() {
...
-                                            .any(|pattern| Pattern::matches_path(pattern, &path)) {
-                                            build(&config);
-                                        }
+                                    // check whether this path has been marked as ignored in config
+                                    let rel_path_matches =
+                                        |pattern| Pattern::matches_path(pattern, rel_path);
+                                    let path_ignored = &config.ignore.iter().any(rel_path_matches);
+
+                                    if !path_ignored {
+                                        build(&config);
```
johannhof: Thanks, this section is much clearer now.

src/main.rs
nott: nice idea

src/main.rs
nott: `&path` is used in `unwrap_or` of `rel_path`, so I have to `path.clone()` here to have same datatype to `true` and `false branches.

nott:
@johannhof thank you for the feedback. Everything but one `path.clone()` has been addressed in 86307d6.

johannhof:
Looks good, thanks! We should probably also add the destination directory in there by default, to avoid infinite loops in case the destination is in the source folder, but that's for another pull request.

Merged 👍 ❤️

Issue #138, opened 2016-09-28
Get the content of a post


deontologician:
It would be nice to be able to get the individual contents of each post for some kind of main page feature. So like:

```liquid
{% for post in posts %}
<article>
  <h2>{{post.title}}</h2>
  {{post.content}}
</article>
{% endfor %}
```

johannhof:
Yup this is a good idea. I wonder if this is trivial to do though, since the post rendering is pretty late in the pipeline and all documents get rendered in parallel.

deontologician:
In jekyll I guess they have an extra "pagination" thing that enables this.
https://jekyllrb.com/docs/pagination/

johannhof:
Actually since we recently separated rendering from file creation this 
might be easier than I thought. I'll give it a try!

PR #145, opened 2016-10-16
Add Jekyll-like excerpts

nott:
This PR adds Jekyll-like excerpts. Please see [Jekyll docs](https://jekyllrb.com/docs/posts/#post-excerpts) for details.

What hasn't been implemented yet:

* site-wide `excerpt_separator` defined in `.cobalt.yml`
* disabling excerpts by setting `excerpt_separator` to `""` (either per post or globally in `.cobalt.yml`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ A static site generator written in [Rust](http://www.rust-lang.org/).
   - [Layouts](#layouts)
   - [Posts](#posts)
   - [Other Files](#other-files)
+  - [Excerpts](#excerpts)
   - [Attributes](#attributes)
   - [RSS](#rss)
   - [Import](#import)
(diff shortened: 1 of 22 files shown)


johannhof:
This looks fine but I don't really like that we need to render the post again using markdown. This exposes a larger issue that has been coming up a couple of times now. What if excerpts contain liquid markup? Should we render the liquid? What if that template contains the excerpt again? This is the same problem as in #138. Posts can't have access to their own content or other posts content via attributes or we end up with infinite recursion/deadlocks. Other files however frequently need access to post content, that is currently limited by the fact that they are rendering at the same time as posts.

I would suggest we change our internal implementation to do two separate render runs, one for posts (which don't get access to `content` and `excerpt`) and one for other templates (which get access to the full post data, including `excerpt` and `content`).

Sorry, I think this is a great PR and a good feature to implement but I would really like to solve the problem this is exposing properly. We can leave this open in the meantime. :)

I can look into solving that if you like. I'll try to be quick!

nott:
> What if excerpts contain liquid markup? Should we render the liquid?

I don't think it makes sense to extract excerpts automatically from non-markdown posts. "First paragraph" doesn't actually makes sense for liquid templates.

I don't think we ever render anything in attributes (e.g. title), so I'd treat `excerpts` provided via attributes in liquid templates as html.

Personally I'd like this to be merged before refactoring. The main use case for excerpts is markdown blog posts. Some oddity in other cases is acceptable.

nott:
I wonder what Jekyll does in cases like #138.

nott:
I'll reimplement this PR once #146 gets merged.

PR #146, opened 2016-10-21
Render posts before documents, remove threading

### LINKED ISSUE #138: Get the content of a post

johannhof:
In #138 and #145 I noticed that it might be beneficial to render posts
before documents, because documents might want to access post
information such as content or excerpt that requires all posts to be
rendered.

This also removes the admittedly naive approach of spawning a thread per
document which at least by my measurements adds way too much overhead to
be of any use. I will open a new issue for experimenting with ways to
add good parallelism.

Fixes #138 by adding a "content" attribute to posts.

### DIFF
--- a/src/cobalt.rs
+++ b/src/cobalt.rs
@@ -1,7 +1,5 @@
-use crossbeam;
-
 use std::fs::{self, File};
-use std::io::{self, Write, ErrorKind};
+use std::io::Write;
 use std::path::{Path, PathBuf};
 use std::ffi::OsStr;
 use liquid::Value;
@@ -12,7 +10,6 @@ use config::Config;
 use chrono::{UTC, FixedOffset};
 use chrono::offset::TimeZone;
 use rss::{Channel, Rss};
-use std::sync::Arc;
 use glob::Pattern;
 
 fn ignore_filter(entry: &DirEntry, source: &Path, ignore: &[Pattern]) -> bool {
@@ -51,11 +48,11 @@ pub fn build(config: &Config) -> Result<()> {
 
     let layouts = source.join(&config.layouts);
     let layouts = layouts.as_path();
-    let posts = source.join(&config.posts);
-    let posts = posts.as_path();
+    let posts_path = source.join(&config.posts);
+    let posts_path = posts_path.as_path();
 
     debug!("Layouts directory: {:?}", layouts);
-    debug!("Posts directory: {:?}", posts);
+    debug!("Posts directory: {:?}", posts_path);
     debug!("Draft mode enabled: {}", config.include_drafts);
     if config.include_drafts {
         debug!("Draft directory: {:?}", config.drafts);
@@ -66,7 +63,7 @@ pub fn build(config: &Config) -> Result<()> {
     let walker = WalkDir::new(&source)
         .into_iter()
         .filter_entry(|e| {
-            (ignore_filter(e, source, &config.ignore) || compare_paths(e.path(), posts)) &&
+            (ignore_filter(e, source, &config.ignore) || compare_paths(e.path(), posts_path)) &&
             !compare_paths(e.path(), dest)
         })
         .filter_map(|e| e.ok());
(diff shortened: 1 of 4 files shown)


johannhof:
@nott Would you like review this or tell me what you think? :)

If you check out the way I assign a new "content" attribute this could be how we could also add the excerpt in your PR without having to re-render the markdown.

It's still missing tests for the "content" part, I'll add those tomorrow.

src/cobalt.rs
```diff
@@ -115,44 +113,47 @@ pub fn build(config: &Config) -> Result<()> {
...
+
     // sort documents by date, if there's no date (none was provided or it couldn't be read) then
     // fall back to the default date
-    documents.sort_by(|a, b| b.date.unwrap_or(default_date).cmp(&a.date.unwrap_or(default_date)));
+    posts.sort_by(|a, b| b.date.unwrap_or(default_date).cmp(&a.date.unwrap_or(default_date)));
 
     // check if we should create an RSS file and create it!
     if let &Some(ref path) = &config.rss {
-        try!(create_rss(path, dest, &config, &documents));
+        try!(create_rss(path, dest, &config, &posts));
```
nott: i think it makes sense to `create_rss` after we insert content, so we can add content (or excerpt) to the feed

nott:
i did some naive performance testing and this pr actually improved performance on my very small blog

src/cobalt.rs
```diff
@@ -115,44 +113,47 @@ pub fn build(config: &Config) -> Result<()> {
...
     }
 
     // these are the attributes of all documents that are posts, so that they can be
     // passed to the renderer
-    // TODO: do we have to clone these?
-    let post_data: Vec<Value> = documents.iter()
-        .filter(|x| x.is_post)
+    let mut post_data: Vec<Value> = posts.iter()
```
nott: I understand the naming comes from previous revisions... however

name `post_data` is a little bit confusing. i'd call this at least `posts_data`

src/cobalt.rs
```diff
@@ -115,44 +113,47 @@ pub fn build(config: &Config) -> Result<()> {
...
 
-    // thread handles to join later
-    let mut handles = vec![];
+    let simple_post_data = post_data.clone();
 
-    // generate documents (in parallel)
-    crossbeam::scope(|scope| {
-        let post_data = Arc::new(post_data);
+    trace!("Generating posts");
+    for (post, data) in posts.iter().zip(post_data.iter_mut()) {
```
nott: we have 3 copies of posts attributes at this point (post_data, simple_post_data, posts)

what if we

* use 2 copies: `simple_post_data` and posts
* `post.attributes.insert` `content` in the posts loop
* and create `post_data` vector once again after the loop

imho that will simplify the code - we won't need the `unreachable!()` below

nott:
great pr

your scepsis about #145 was fair

src/cobalt.rs
nott: however this should def be a separate pr

src/cobalt.rs
johannhof: **Awesome idea**, I think we can even improve on top of this and get down to cloning posts only once!

src/cobalt.rs
johannhof: Good point

src/cobalt.rs
johannhof: Yes, I agree. Happy to take a PR if you want to take care of it :)

nott: :+1:

johannhof:
@nott Thanks for the great comments. I think we ended up with faster and much more readable and code.

> i did some naive performance testing and this pr actually improved performance on my very small blog

Yeah I implemented the threading at a time when I was just blindly parallelizing my Rust code because it's so easy. Turns out oftentimes the overhead (performance + cognitive) is larger than if you just do it sequentially.

nott:
LGTM

nott:
@johannhof

i've just realized there is a bug here: `content` contains html taken from layouts
that won't be useful for the case mentioned in #138

Issue #120, opened 2016-07-07
Add a layout cache


johannhof:
Since #119 we don't pre-load layouts anymore, instead they are read when generating documents. This causes a lot of duplicate fs calls and memory allocation for commonly used layouts. A simple shared cache Mutex could help.

PR #147, opened 2016-10-22
Refactor methods for rendering documents, add excerpts and fix content attributes

nott:
Note: the scope of this PR has changed.
- This PR adds Jekyll-like excerpts. Please see [Jekyll docs](https://jekyllrb.com/docs/posts/#post-excerpts) for details. The previous attempt to do it was #145 .
- Fixes content attribute added in #146 - this attribute shouldn't include extended layout, just the document.
- Changes the way we split attributes and content.
- Fixes bug with two sequential rendering when `extends` is not defined.
- Adds a layout cache (#120).

Method `as_html` was replaces with `render` which now has side-effects.

### DIFF
--- a/src/cobalt.rs
+++ b/src/cobalt.rs
@@ -136,6 +136,7 @@ pub fn build(config: &Config) -> Result<()> {
 
         let mut context = post.get_render_context(&simple_posts_data);
 
+        try!(post.render_excerpt(&mut context, &source, &config.excerpt_separator));
         let post_html = try!(post.render(&mut context, &source, &layouts));
         try!(create_document_file(&post_html, &post.path, dest));
     }
--- a/src/cobalt.rs
+++ b/src/cobalt.rs
@@ -134,10 +134,10 @@ pub fn build(config: &Config) -> Result<()> {
     for mut post in &mut posts {
         trace!("Generating {}", post.path);
 
-        let content = try!(post.as_html(&source, &simple_posts_data, &layouts));
-        try!(create_document_file(&content, &post.path, dest));
+        let mut context = post.get_render_context(&simple_posts_data);
 
-        post.attributes.insert("content".to_owned(), Value::Str(content));
+        let post_html = try!(post.render(&mut context, &source, &layouts));
+        try!(create_document_file(&post_html, &post.path, dest));
     }
 
     // during post rendering additional attributes such as content were
@@ -147,11 +147,12 @@ pub fn build(config: &Config) -> Result<()> {
         .collect();
 
     trace!("Generating other documents");
-    for doc in documents {
+    for mut doc in documents {
         trace!("Generating {}", doc.path);
 
-        let content = try!(doc.as_html(&source, &posts_data, &layouts));
-        try!(create_document_file(&content, &doc.path, dest));
+        let mut context = doc.get_render_context(&posts_data);
+        let doc_html = try!(doc.render(&mut context, &source, &layouts));
+        try!(create_document_file(&doc_html, &doc.path, dest));
     }
 
     // copy all remaining files in the source to the destination
--- a/src/cobalt.rs
+++ b/src/cobalt.rs
@@ -1,4 +1,5 @@
 use std::fs::{self, File};
+use std::collections::HashMap;
 use std::io::Write;
 use std::path::{Path, PathBuf};
 use std::ffi::OsStr;
@@ -48,6 +49,7 @@ pub fn build(config: &Config) -> Result<()> {
 
     let layouts = source.join(&config.layouts);
     let layouts = layouts.as_path();
+    let mut layouts_cache = HashMap::new();
     let posts_path = source.join(&config.posts);
     let posts_path = posts_path.as_path();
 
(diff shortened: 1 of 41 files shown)


nott:
General notice:

* on the one hand writing tests with fxitures/target
* however it seems that they it will be hard to support these tests in future
* it's not very convenient to cover `Document::as_html` with unit tests when this function accepts paths (for `source` and `layouts_path`) instead of file contents

We should probably move reading `source` and `layout` into separate functions and cover all the functions with unit tests.

nott:
I'll probably extend the scope of this PR in a day or two.

nott:
@johannhof could you please review this?

src/document.rs
```diff
@@ -264,17 +258,79 @@ impl Document {
...
+    /// When we say "content" we mean only this document without extended layout.
+    pub fn render(&mut self,
+                  context: &mut Context,
+                  source: &Path,
+                  layouts_dir: &Path,
+                  layouts_cache: &mut HashMap<String, String>)
+                  -> Result<String> {
+        let content_html = try!(self.render_html(&self.content, context, source));
+        self.attributes.insert("content".to_owned(), Value::Str(content_html.clone()));
+        context.set_val("content", Value::Str(content_html.to_owned()));
```
johannhof: please use `clone` instead of `to_owned` here, too

src/document.rs
```diff
@@ -264,17 +258,79 @@ impl Document {
...
+        Ok(html.to_owned())
+    }
 
-        let options = LiquidOptions { file_system: Some(source.to_owned()), ..Default::default() };
+    /// Renders excerpt and adds it to attributes of the document.
+    pub fn render_excerpt(&mut self,
+                          context: &mut Context,
+                          source: &Path,
+                          default_excerpt_separator: &str)
+                          -> Result<String> {
```
johannhof: This doesn't have to return a `String`

src/document.rs
```diff
@@ -264,17 +258,79 @@ impl Document {
...
+                .and_then(|attr| attr.as_str()) {
+                excerpt_str.to_string()
+            } else if excerpt_separator.is_empty() {
+                "".to_owned()
+            } else {
+                self.content.split(excerpt_separator)
+                    .filter(|chunk| !chunk.is_empty())
+                    .next() // first non-empty chunk
+                    .map(|s| s.to_string())
+                    .unwrap_or(self.content.to_string())
```
johannhof: It's not a huge thing but you can save on allocation and code if you make `excerpt` a `&str` instead of an owned String. You'd have to move it into a block for creating `excerpt_html` to avoid the dangling reference to `self.attributes`, though.

johannhof: Hm maybe that comment wasn't really clear, here's how I'd implement it:
```
    /// Renders excerpt and adds it to attributes of the document.
    pub fn render_excerpt(&mut self,
                          context: &mut Context,
                          source: &Path,
                          default_excerpt_separator: &str)
                          -> Result<()> {
        let excerpt_html = {
            let excerpt_attr = self.attributes
                .get("excerpt")
                .and_then(|attr| attr.as_str());

            let excerpt_separator = self.attributes
                .get("excerpt_separator")
                .and_then(|attr| attr.as_str())
                .unwrap_or(default_excerpt_separator);

            let excerpt = if let Some(excerpt_str) = excerpt_attr {
                excerpt_str
            } else if excerpt_separator.is_empty() {
                ""
            } else {
                self.content.split(excerpt_separator)
                        .filter(|chunk| !chunk.is_empty())
                        .next() // first non-empty chunk
                        .unwrap_or(&self.content)
            };

            try!(self.render_html(excerpt, context, source))
        };

        self.attributes.insert("excerpt".to_owned(), Value::Str(excerpt_html));
        Ok(())
    }
```

src/document.rs
```diff
@@ -264,17 +258,79 @@ impl Document {
...
-            try!(liquid::parse(&layout, options))
-        } else {
-            try!(liquid::parse(&html, options))
-        };
+    /// Renders the document to an HTML string.
+    ///
+    /// Side effects:
+    ///
+    /// * content is inserted to the attributes of the document
+    /// * content is inserted to context
```
johannhof: another side effect is that the layout is inserted into the `layouts_cache`

src/cobalt.rs
```diff
@@ -48,6 +49,7 @@ pub fn build(config: &Config) -> Result<()> {
 
     let layouts = source.join(&config.layouts);
     let layouts = layouts.as_path();
+    let mut layouts_cache = HashMap::new();
```
johannhof: thanks for adding this

nott:
@johannhof thank you for the feedback!
i'll address all your comments later today or tomorrow

johannhof:
Sounds great, thanks!

nott:
> Sorry for the delay, I had to think about this. As you know I'm not a big fan of rendering twice, but I think you're right, there's no other way to do this if we want excerpts.

We can't avoid rendering twice indeed. I was thinking about lazy rendering on first usage, but it complicates the code. I don't think it makes sense to add this feature having the existing infrastructure... But it's a good project for future that can improve both liquid-rust and cobalt.

src/document.rs
nott: nice catch

nott:
@johannhof all your comments have been addressed.

There was one more additional change in `render_excerpt`:

Instead of
```
                self.content.split(excerpt_separator)
                        .filter(|chunk| !chunk.is_empty())
                        .next() // first non-empty chunk
                        .unwrap_or(&self.content)
```
I used
```
self.content.split(excerpt_separator).next().unwrap_or(&self.content)
```

Skipping leading empty chunk was imported before 8af758b was merged. But as soon as 8af758b is merged, skipping empty chunks doesn't make much sense (jekyll also doesn't do it and it's contr-intuitive to skip them when using custom separators like `<!-- more -->`)

johannhof:
Great work, thank you!

nott:
i've just noticed one `clone` that should have been removed... fixed

PR #148, opened 2016-10-30
RSS: use content as a default description, add guid

nott:
This PR:
- uses excerpt or content as a default value for description in rss
- adds guid tag to rss feed (as suggested by [validator.w3.org](https://validator.w3.org/feed/docs/warning/MissingGuid.html))

### DIFF
--- a/src/document.rs
+++ b/src/document.rs
@@ -222,6 +222,7 @@ impl Document {
     pub fn to_rss(&self, root_url: &str) -> rss::Item {
         let description = self.attributes
             .get("description")
+            .or(self.attributes.get("excerpt"))
             .or(self.attributes.get("content"))
             .and_then(|s| s.as_str())
             .map(|s| s.to_owned());
--- a/src/document.rs
+++ b/src/document.rs
@@ -220,15 +220,23 @@ impl Document {
 
     /// Metadata for generating RSS feeds
     pub fn to_rss(&self, root_url: &str) -> rss::Item {
-        let description = self.attributes.get("description")
+        let description = self.attributes
+            .get("description")
             .or(self.attributes.get("content"))
             .and_then(|s| s.as_str())
             .map(|s| s.to_owned());
 
+        // Swap back slashes to forward slashes to ensure the URL's are valid on Windows
+        let link = root_url.to_owned() + &self.path.replace("\\", "/");
+        let guid = rss::Guid {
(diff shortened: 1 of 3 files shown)


src/document.rs
```diff
@@ -47,6 +47,38 @@ fn yaml_to_liquid(yaml: &Yaml) -> Option<Value> {
...
+            }
+            _ => {}
+        }
+    }
+
+    if last < html.len() {
+        escaped.push_str(&html[last..]);
+    }
+    escaped
+}
```
johannhof: We should consider publishing a crate with this functionality. We will also need it in liquid to add an `escape` filter and I've encountered multiple occasions where a project needed a simple HTML escape function. An alternative could be using e.g. https://github.com/nagisa/marksman_escape/

nott: @johannhof 

* publishing a single function crate doesn't make much sense to me
* marksman_escape doesn't seem to be actively developed/supported
* adding marksman_escape as a dep for one simple function looks like an overkill

if you need it in liquid-rust, does it make sense to add this function there and reuse it in cobalt?

nott:
Note: this is not ready for review yet.

src/document.rs
nott: i was confused by my rss reader when i was testing this locally - in fact rss crate escapes html for us: https://github.com/rust-syndication/rss/blob/0.3.1/src/item.rs#L55 (this fact is also covered by `tests/target/rss/rss.xml`)

so i removed `escape_html` function

nott:
@johannhof 
#147 is merged, so this is ready for review

src/document.rs
```diff
@@ -220,15 +220,26 @@ impl Document {
 
     /// Metadata for generating RSS feeds
     pub fn to_rss(&self, root_url: &str) -> rss::Item {
+        let description = self.attributes
+            .get("description")
+            .or_else(|| self.attributes.get("excerpt"))
+            .or_else(|| self.attributes.get("content"))
```
johannhof: doesn't `.or(self.attributes.get("excerpt"))` work as well?

nott: `.or_else(|| self.attributes.get("excerpt"))` is more lazy

it's hard to tell in advance which one is faster, so I'll take your approach as a more readable

johannhof:
Looks good, thanks!

PR #150, opened 2016-11-04
adding clean command

### LINKED ISSUE #102: Clean command
We should add a command that will recursively remove the `dest` directory.

rjgoldsborough:
fixes #102 

If there is a cleaner way to do this, please let me know.
I kept getting `expected (), found enum 'std::result::Result\'` but writing it this way worked.

I did not add a test for this just yet but trying to figure that out.

### DIFF
--- a/src/cobalt.rs
+++ b/src/cobalt.rs
@@ -42,13 +42,6 @@ pub fn build(config: &Config) -> Result<()> {
     let dest = PathBuf::from(&config.dest).join("");
     let dest = dest.as_path();
 
-    // try to create the supplied dest dir if it currently does not exist
-    if dest.is_dir() == false {
-        try!(fs::create_dir(dest));
-        info!("Created dest dir: {:?}", dest);
-    }
-
-
     let template_extensions: Vec<&OsStr> = config.template_extensions
         .iter()
         .map(OsStr::new)
--- a/src/cobalt.rs
+++ b/src/cobalt.rs
@@ -42,6 +42,13 @@ pub fn build(config: &Config) -> Result<()> {
     let dest = PathBuf::from(&config.dest).join("");
     let dest = dest.as_path();
 
+    // try to create the supplied dest dir if it currently does not exist
+    if dest.is_dir() == false {
+        try!(fs::create_dir(dest));
+        info!("Created dest dir: {:?}", dest);
+    }
+
+
     let template_extensions: Vec<&OsStr> = config.template_extensions
         .iter()
         .map(OsStr::new)
--- a/src/main.rs
+++ b/src/main.rs
@@ -122,8 +122,7 @@ fn main() {
                 .help("Commit message that will be used on import")
                 .default_value("cobalt site import")
                 .takes_value(true)))
-        .subcommand(SubCommand::with_name("clean")
-            .about("cleans directory set as destination"))
+        .subcommand(SubCommand::with_name("clean").about("cleans directory set as destination"))
         .subcommand(SubCommand::with_name("serve")
             .about("build and serve the cobalt project at the source dir")
             .arg(Arg::with_name("port")
@@ -262,8 +261,8 @@ fn main() {
 
         "clean" => {
             match fs::remove_dir_all(&config.dest) {
-              Ok(..) => println!("directory \"{}\" removed", &config.dest),
-              Err(err) => println!("Error: {}", err)
+                Ok(..) => info!("directory \"{}\" removed", &config.dest),
+                Err(err) => error!("Error: {}", err),
             }
         }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -122,6 +122,8 @@ fn main() {
                 .help("Commit message that will be used on import")
                 .default_value("cobalt site import")
                 .takes_value(true)))
+        .subcommand(SubCommand::with_name("clean")
+            .about("cleans directory set as destination"))
         .subcommand(SubCommand::with_name("serve")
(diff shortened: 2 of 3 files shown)


src/main.rs
```diff
@@ -258,6 +260,13 @@ fn main() {
             }
         }
 
+        "clean" => {
+            match fs::remove_dir_all(&config.dest) {
```
nott: `cobalt build` doesn't create `dest` directory:
```
$ cobalt build
[info]   Using config file ./.cobalt.yml
[info]   Building from . into build
[error]  IO error: No such file or directory (os error 2)
[error]  Build not successful
```

I think it meanse `cobalt clean` shouldn't remove it.

src/main.rs
```diff
@@ -258,6 +260,13 @@ fn main() {
...
+              Ok(..) => println!("directory \"{}\" removed", &config.dest),
+              Err(err) => println!("Error: {}", err)
```
nott: could you please use `error!` instead of `println!` here?

nott:
Travis build is failing due to issues with formatting.

You can fix them with [cargo fmt](https://github.com/rust-lang-nursery/rustfmt#quick-start)

rjgoldsborough:
Thanks for the feedback. (sorry for the sloppiness :))
I will incorporate these.

src/main.rs
nott: Makes sense indeed.

rjgoldsborough: I'll definitely give it a shot and come back with any questions or issues I run into.

johannhof:
Thanks for the PR! What you *could* do to test this is to take a look at [tests/cli.rs](https://github.com/cobalt-org/cobalt.rs/blob/master/tests/cli.rs) and add a new test for `clean` that runs

```
env::set_current_dir(CWD.join("tests/fixtures/example")).unwrap();
assert_cli!(&BIN, &["build", "-d", "./test_dest"] => Success).unwrap();
```
check if  `./test_dest` was created and then run 

```
let output = assert_cli!(&BIN, &["clean", "-d", "./test_dest"] => Success).unwrap();
```

and check if `./test_dest` was removed. You should also check if `output` has the correct messages.

Let me know how that works for you :)

Thanks!

src/main.rs
rjgoldsborough: hmm weird. cobalt *does* create the dir for me.

https://gist.github.com/rjgoldsborough/08fc7700a847b9318a7ebd1da10ade89

rjgoldsborough:
@johannhof, thanks for the suggestions.

It seems to have worked for the most part. I was trying to check the output but was getting errors about `PartialEq` and `&str` so I pushed the rest of the test until I get that sorted.

rjgoldsborough:
One other question I had with that last commit was if we should use `create_dir_all` for the case of a user supplying a multi level dest dir?

src/main.rs
nott: @rjgoldsborough 

my comment was misleading, sorry for that
i tested on version 0.3, latest master creates dest dir automatically indeed

src/cobalt.rs
```diff
@@ -42,6 +42,13 @@ pub fn build(config: &Config) -> Result<()> {
     let dest = PathBuf::from(&config.dest).join("");
     let dest = dest.as_path();
 
+    // try to create the supplied dest dir if it currently does not exist
+    if dest.is_dir() == false {
+        try!(fs::create_dir(dest));
+        info!("Created dest dir: {:?}", dest);
+    }
```
johannhof: I'm pretty sure you can remove this again. With your new test it should be easy to verify. Sorry for the confusion. :)

tests/cli.rs
```diff
@@ -65,3 +65,12 @@ pub fn log_levels() {
...
 }
 
+#[test]
+pub fn clean() {
+    env::set_current_dir(CWD.join("tests/fixtures/example")).unwrap();
+    assert_cli!(&BIN, &["build", "-d", "./test_dest"] => Success).unwrap();
+    assert_eq!(Path::new("./test_dest/").is_dir(), true);
+
+    assert_cli!(&BIN, &["clean", "-d", "./test_dest"] => Success).unwrap();
+    assert_eq!(Path::new("./test_dest").is_dir(), false);
```
johannhof: why can't you use `assert_contains` like in the above examples here? E.g.:

```rust
assert_contains!(&output.stderr, "directory \"./test_dest/\" removed");
```

rjgoldsborough:
I removed that code to add the dest dir but had an issue trying to test the output.

It seems like there is no default config file when testing so the output when running clean was:

```
running 3 tests
test invalid_calls ... ok
test log_levels ... ok
test clean ... FAILED

failures:

---- clean stdout ----
	[warn]   No .cobalt.yml file found in current directory, using default config.
[info]   directory "./test_dest" removed

thread 'clean' panicked at 'assertion failed: text.contains("[info]    directory \"./test_dest\" removed")', tests/cli.rs:77
```

so I think maybe that warn is getting tested as well?

Any ideas on how to handle that?

johannhof:
Hm maybe it's because of the whitespace between `[info]` and the rest?

For me adding

```rust
assert_contains!(&output.stderr, "directory \"./test_dest\" removed");
```

passes all tests :)

rjgoldsborough:
Huh yeah, I dunno. I reran the tests this AM and they seem fine.

johannhof:
Great, thanks!

nott:
Awesome

PR #151, opened 2016-11-04
Add Cargo.lock to repository

nott:
The best practice is to keep this file in gitignore for libraries and in repo for applications.

Committing Cargo.lock makes the builds much more reproducible.

Note: I've called `cargo update` just before creating this PR, so I hope all the dependencies are now up-to-date.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,4 @@
 /build
 
 .DS_Store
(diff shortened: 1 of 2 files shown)


johannhof:
:+1:

PR #153, opened 2016-11-07
Add release notes for v0.4

nott:


nott:
I expect the tests to be red due to an issue with crates.io certificate

src/main.rs
```diff
@@ -33,7 +33,7 @@ use std::fs::File;
 
 fn main() {
     let global_matches = App::new("Cobalt")
-        .version("0.3.0")
+        .version(crate_version!())
```
johannhof: nice!

nott:
@johannhof do you mind if i use crate_version macro so we have a single version number in the source code?

johannhof:
Not at all.

PR #156, opened 2016-11-19
error-chain for error handling

benaryorg:
I thought using [*error-chain*](https://crates.io/crates/error-chain) might be nice because it removes a lot of overhead code.

It also enables easy addition of errors. This in turn means that one can use more detailed `ErrorKind`s than a `String` *so* easy that people actually add verbose error handling.

Furthermore it makes it possible to use *error-chain* in projects that depend on *cobalt.rs* to "include" it's error chain directly.


johannhof:
Thanks!!!

PR #157, opened 2016-11-19
ErrorKind error instead of String

benaryorg:
Depends on #156.

Replaces the error generated on incomplete RSS configs with an `ErrorKind` instead of a `String`.


benaryorg:
After messing with git I managed to make it right. It works. You can merge this.

johannhof:
Nice! Thanks again.

PR #159, opened 2016-11-19
Path only in build()

benaryorg:
Strip all non-`Path` things from `build()`.

Exceptions: `OsStr` for *dirwalk*.


johannhof:
Cautiously optimistic, let's see if it breaks anything.

PR #165, opened 2016-11-20
Adding syntax highlighting

gnunicorn:
This work adds syntax highlighting to cobalt using the awesome syntect library. This still work in progress, not ready for merging yet!

- [x] add syntax highligting behind a (default on) feature flag
- [x] implement [highlight template tag](http://jekyllrb.com/docs/templates/#code-snippet-highlighting)
- [x] implement for markdown-inline-code-fences (_three ticks_)
- [x] include default set of syntax parsers into the binary (comes built-in with syntect! - whooot!?!)
- [x] add feature to configure hl-theme (how?)
- [x] move syntax loading into singleton

### DIFF
--- a/src/syntax_highlight.rs
+++ b/src/syntax_highlight.rs
@@ -99,8 +99,11 @@ impl<'a> Iterator for DecoratedParser<'a> {
                             .next()
                             .and_then(|lang| SETUP.syntax_set.find_syntax_by_token(lang))
                             .unwrap_or_else(|| SETUP.syntax_set.find_syntax_plain_text());
-                        self.h = Some(HighlightLines::new(&cur_syntax, &SETUP.theme_set.themes[THEME_NAME]));
-                        return Some(Html(Owned(start_coloured_html_snippet(&SETUP.theme_set.themes[THEME_NAME]))));
+                        self.h = Some(HighlightLines::new(&cur_syntax,
+                                                          &SETUP.theme_set.themes[THEME_NAME]));
+                        let snippet =
+                            start_coloured_html_snippet(&SETUP.theme_set.themes[THEME_NAME]);
+                        return Some(Html(Owned(snippet)));
                     }
                     if let End(cmarkTag::CodeBlock(_)) = item {
                         // reset highlighter
--- a/src/syntax_highlight.rs
+++ b/src/syntax_highlight.rs
@@ -10,9 +10,10 @@ use liquid::Token::{self, Identifier};
 use liquid::lexer::Element::{self, Expression, Tag, Raw};
 use liquid::Error;
 
-use syntect::parsing::{SyntaxSet};
+use syntect::parsing::SyntaxSet;
 use syntect::highlighting::ThemeSet;
-use syntect::html::{IncludeBackground, highlighted_snippet_for_string, styles_to_coloured_html, start_coloured_html_snippet};
+use syntect::html::{IncludeBackground, highlighted_snippet_for_string, styles_to_coloured_html,
+                    start_coloured_html_snippet};
 use syntect::easy::HighlightLines;
 
 use std::borrow::Cow::Owned;
--- a/src/syntax_highlight.rs
+++ b/src/syntax_highlight.rs
@@ -1,4 +1,3 @@
-
 extern crate syntect;
 extern crate liquid;
 extern crate pulldown_cmark as cmark;
@@ -29,10 +28,8 @@ struct Setup {
     theme_set: ThemeSet,
 }
 
-
 unsafe impl Send for Setup {}
 unsafe impl Sync for Setup {}
-
 
 lazy_static!{
     static ref SETUP: Setup = Setup {
@@ -40,7 +37,6 @@ lazy_static!{
         theme_set: ThemeSet::load_defaults()
     };
 }
-
 
 struct CodeBlock {
     lang: Option<String>,
@@ -62,7 +58,6 @@ impl Renderable for CodeBlock {
     }
 }
 
-
 pub struct DecoratedParser<'a> {
     h: Option<HighlightLines<'a>>,
     parser: Parser<'a>,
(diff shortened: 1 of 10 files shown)


gnunicorn:
@johannhof We are feature complete. Including [markdown-code fences](https://github.com/cobalt-org/cobalt.rs/pull/165/commits/b8a777782bd921180aad46649b8e05476f1e8c7b). However, it currently loads the syntax dump and themes on every render, slowing down the exection time _by magnitudes_. That's quite unacceptable. So I am wondering, how we want to deal with that. Is there any nice way to only load on the first time it is needed and then keep it around until we clean up? Some form of singleton?

----

Secondly, how do we want manage theme configuration? Should be put into the confguration file, so do it globally? I guess that would be best, as it means we could discard all other themes. If we allowed configuration per code-block, it would mean keeping all themes around for the entire execution.

gnunicorn:
@johannhof I am using `lazy_static!` now to make the loading only happen once and stuff is fast and smooth again (at least running tests). However, I had to wrap it into a struct that I had to implement unsafe Sync and Send for. The entire part doesn't look very clean to me :( . So if you have suggestions how this could be solved instead, I am all ears.

gnunicorn:
Regarding the them configuration: we aren't passing down the config object into the document as of now, nor are we passing it into liquid (needed for the code-fences). I've defined it as a const as of now, we should change that. But I'd leave that to a future PR.

gnunicorn:
P.S. I ran `cargo fmt` yet travis is complaining about things in the format *sigh* ... not sure how you want to handle this.

nott:
@gnunicorn what version of rustfmt are you using?

gnunicorn:
@nott not sure and `cargo fmt --version` doesn't help figuring that out. But installed it the first time yesterday (via `cargo install`), so I suppose very current ;) . Running both on latest stable via rustup on arch linux.

nott:
@gnunicorn 

You can check version of rustfmt with `rustfmt -V`.

By the way, you can install rustfmt from `community` repo via `pacman` on Arch. I actually use it and it works:

```
$ git co 2066198
HEAD is now at 2066198... roundabout cargo format to make travis happy
$ git diff
$ cargo fmt
$ git diff 
diff --git a/src/document.rs b/src/document.rs
index 9f1545d..dbff3bd 100644
--- a/src/document.rs
+++ b/src/document.rs
@@ -166,9 +166,8 @@ impl Document {
             content
         };
 
-        if let &mut Value::Bool(val) =
-            attributes.entry("is_post".to_owned())
-                .or_insert(Value::Bool(is_post)) {
+        if let &mut Value::Bool(val) = attributes.entry("is_post".to_owned())
+            .or_insert(Value::Bool(is_post)) {
             is_post = val;
         }
 
```

gnunicorn:
I am on `0.6.3` and it can't find any changes to be done:

```bash
[ben:~/dev/cobalt-org/cobalt.rs] syntax-hl 10s 130 ± git branch
  master
* syntax-hl
[ben:~/dev/cobalt-org/cobalt.rs] syntax-hl ± git push
Everything up-to-date
[ben:~/dev/cobalt-org/cobalt.rs] syntax-hl 8s ± rustfmt --version
0.6.3 ()
[ben:~/dev/cobalt-org/cobalt.rs] syntax-hl ± rustfmt --write-mode=diff src/*.rs tests/*.rs
[ben:~/dev/cobalt-org/cobalt.rs] syntax-hl ± git st
[ben:~/dev/cobalt-org/cobalt.rs] syntax-hl ± 
```

ducks:
It looks like the issue here is the version difference.

The travis build is running rustfmt 0.6.2.
0.6.2 seems to make changes while 0.6.3 doesn't.

You can install 0.6.2 with `cargo install --force --vers =0.6.2 rustfmt` and see the changes.
Dunno if this is a bug or just a new formatting rule?

Maybe we can just make travis run `rustfmt` vs checking for it and erroring?

johannhof:
@gnunicorn thanks for the awesome PR. I re-triggered Travis and the build seems to pass. It was definitely Travis having the outdated rustfmt, not you. Before getting to code review, this is how the test page looks for me:

screen shot 2016-11-24 at 00 25 46

As you can see the tag-highlighted block looks fine (:thumbsup::thumbsup::thumbsup:!), but there seems to be a bug with the markdown one.

johannhof:
@gnunicorn I fixed the markdown problem in the latest commit on your branch. Moar review tomorrow :)

src/syntax_highlight.rs
```diff
@@ -0,0 +1,195 @@
...
+            }
+            .to_owned() + &a
+    });
+
+    let lang = match arguments.iter().next() {
+        Some(&Identifier(ref x)) => Some(x.clone()),
+        _ => None,
+    };
+
+    // FIXME: add language declarion support
```
gnunicorn: that's an outdated comment. already fixed it.

gnunicorn:
@johannhof anything else for me to get this merged? (Wanted to ask at colearning, but you aren't here ;) )

johannhof:
@gnunicorn sorry for the delay. Travels, the cold and a baby. Anyway:

- There are still rustfmt issues apparently:
  https://travis-ci.org/cobalt-org/cobalt.rs/jobs/178467790
- It seems that syntect does not build on Windows because [rust-onig](https://github.com/rust-onig/rust-onig) does not compile there without the `static-libonig` flag. This should be solved on syntect (ideally by adding Windows CI to that project), but for now I suggest we just disable syntax highlighting on Windows and make a new issue.

gnunicorn:
I fix the rust formatting problems.

I also guarded the feature behind a windows-platform check, and as it isn't always available for now (and we no way to communicate that) I made it non-default. Until we have support for windows, you need to explicitly tell people to compile it in and that it won't work on windows.

johannhof:
Alright, looks like we can merge this. Thanks for the patience!

PR #166, opened 2016-11-20
Add warning message if the current directory is being deleted.

### LINKED ISSUE #161: Add a warning if "cobalt clean" would delete the current working directory
Because that just happened to me.

kracekumar:
Fixes #161.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,6 +19,7 @@ dependencies = [
  "pulldown-cmark 0.0.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "regex 0.1.80 (registry+https://github.com/rust-lang/crates.io-index)",
  "rss 0.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "syntect 1.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "tempdir 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "walkdir 0.1.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "yaml-rust 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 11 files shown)


kracekumar:
I am happy to modify if there is a different expectation for the warning message.

LucioFranco:
@kracekumar thanks for contributing! The tests on master on travis seem to fail because of a nightly bug so don't worry about that. It looks good. I do not have time now to test but I would write a cli test and assert that this warning is displayed.

kracekumar:
@LucioFranco I have added the test case for asserting the warning. Can you give your feedback? Apologies for last few commits where I had hiccup figuring out running integration tests. Do you want me to squash commits?

johannhof:
This looks good but I realize I didn't quite word #161 correctly. Instead of warning, I would like to cancel the operation if the current directory would be deleted. I can not come up with scenario where a user would like to do that.

So it'd be great if you could change the `warn` into `error` and add a `std::process::exit(1);` :)

kracekumar:
@johannhof I have fixed the PR as per latest comment.

johannhof:
Ok, thanks! For some reason this is failing on Windows https://ci.appveyor.com/project/tak1n/cobalt-rs/build/1.0.117/job/8xrdoa78i0hbl5a2#L266

I would suggest to not create a new directory in fixtures, but use e.g. `tests/fixtures/example` instead. Maybe that fixes the Windows error, too.

johannhof:
Hum, very strange. I'm pretty clueless why these failures on Windows are happening, I think I'd try to change the folders the tests are using to random other ones from the `fixtures` folder and see how that turns out.

johannhof:
Alright, this has really been stale for long enough now. Sorry for that. I don't really know what the issue on Windows is, so I disabled the test on Windows for now. I'll make a new issue and whoever feels like it can investigate.

Thanks for the contribution!

PR #173, opened 2017-01-06
Bump clippy to 0.0.106 to fix nightly (1.16) travis builds

nott:
While working on another PR I've noticed that nightly travis builds are broken due to an old version of clippy (0.0.98).

This PR bumps clippy to the recent available version and fixes or ignores found clippy violations.

Red job example: [jobs/189656668](https://travis-ci.org/cobalt-org/cobalt.rs/jobs/189656668)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5,7 +5,7 @@ dependencies = [
  "assert_cli 0.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "chrono 0.2.25 (registry+https://github.com/rust-lang/crates.io-index)",
  "clap 2.18.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


johannhof:
Oh, right, I was wondering what's up with Clippy lately, forgot that we added a Cargo.lock file. Thanks!

nott:
We can also try to remove clippy from `Cargo.{toml,lock}` and install it via `cargo install clippy`. But that will require updating `.travis.yml` to skip this step on stable and beta.

johannhof:
I'm generally unhappy with this way of running Clippy, but I don't know of a better alternative.

nott:
It does.

Please see the note about cargo-clippy feature here: https://github.com/Manishearth/rust-clippy/#allowingdenying-lints

johannhof:
Ah-ha, they seem to have added that. That would make things much easier! Awesome!

PR #176, opened 2017-01-12
Le Big CI Overhaul

johannhof:
This switches our CI to a system that

- [x] Automatically builds and uploads binaries for Linux, OSX and Windows that can be easily installed through a shell script
- [x] Downloads rustfmt binaries instead of installing them from source
- [x] Runs cargo-clippy which allows us to remove most clippy rules from our lib.rs and will keep us up-to date on clippy (it also makes it way easier to run clippy locally)
- [x] Fixes code coverage

The system is originally based on https://github.com/japaric/trust but I made some pretty big changes along the way, including throwing out cross for now. We can easily drop it in once we need support for more exotic compilation targets.

cc @nott @gnunicorn

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,9 +1,12 @@
-# This is based on https://github.com/rust-lang-nursery/rustfmt/blob/master/appveyor.yml
-# which is in turn based on https://github.com/japaric/rust-everywhere/blob/master/appveyor.yml
+# Based on the "trust" template v0.1.1
+# https://github.com/japaric/trust/tree/v0.1.1
 
 environment:
   global:
-    PROJECT_NAME: cobalt.rs
+    RUST_VERSION: stable
+
+    CRATE_NAME: cobalt
+
   matrix:
     # Stable channel
     - TARGET: x86_64-pc-windows-gnu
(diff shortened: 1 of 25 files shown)


johannhof:
The deploy keys are still local to my repo, I'll update them as soon as I'm done with code coverage

johannhof:
A nice side effect is that CI builds run much faster now, but that's mostly diminished by Travis OSX machines being so scarce. It can take up to an hour during US business times to get one. Unfortunately it's not possible to only run OSX build on tag releases.

ci/before_deploy.sh
```diff
@@ -0,0 +1,29 @@
...
+    case $TRAVIS_OS_NAME in
+        linux)
+            stage=$(mktemp -d)
+            ;;
+        osx)
+            stage=$(mktemp -d -t tmp)
+            ;;
+    esac
+
+    test -f Cargo.lock || cargo generate-lockfile
```
nott: how can Cargo.lock be missing?

tests/cli.rs
```diff
@@ -11,6 +11,7 @@ static EMPTY: &'static [&'static str] = &[];
 lazy_static! {
     static ref _CWD: PathBuf = env::current_dir().unwrap();
     static ref CWD: &'static Path = _CWD.as_path();
+    // TODO test on release
```
nott: let's create a github issue so this doesnt't get lost

appveyor.yml
```diff
@@ -1,9 +1,12 @@
...
-# which is in turn based on https://github.com/japaric/rust-everywhere/blob/master/appveyor.yml
+# Based on the "trust" template v0.1.1
+# https://github.com/japaric/trust/tree/v0.1.1
 
 environment:
   global:
-    PROJECT_NAME: cobalt.rs
+    RUST_VERSION: stable
+
+    CRATE_NAME: cobalt
```
johannhof: Cobalt, cobalt-bin is only the crate name, the binary is called cobalt.

ci/before_deploy.sh
johannhof: Good question, that's still from trust, I can probably just remove it. I also think it'd be auto-generated otherwise anyway.

tests/cli.rs
johannhof: Yup I'll do that once this merges so that I can link to the line.

johannhof:
Ugh so it seems like cargo-clippy is currently breaking in Nightly, unfortunately there's no way around that even with the refactor. I'll go ahead and merge anyway, I'm sure it will go away soon.

LucioFranco:
@johannhof sorry, I never got to reviewing the PR. Been quite busy lately.

PR #182, opened 2017-01-18
Enable syntax-highlighting by default

johannhof:
The new version of syntect allows us to statically link libonig, so let's see what our CI thinks of that.


johannhof:
Clippy is borked but this doesn't touch code so yay

PR #184, opened 2017-01-26
Changing build path handling to use `std::path::Path`.

### LINKED ISSUE #183: Static files placed in incorrect location when `.` used as source
When `.cobalt.yml`'s `source` variable is equal to `.`, all static files are copied into the root of the output folder with their name being their extension.

To reproduce:
```
mkdir cobalt-test
cd cobalt-test
cobalt new
touch myimg.png
cobalt build
```
`build/myimg.png` should now exist; however it does not, and `build/png` exists.

Version information:
```
$ cobalt --version
Cobalt 0.5.0
```

leesdolphin:
Fixes #183.

Tests are pending a discussion in [Gitter](https://gitter.im/cobalt-org/cobalt.rs?at=5889ebe3dcb66e4f76b40e26)


johannhof:
Hey, thanks so much! This is an awesome change.

As mentioned on Gitter, if `cargo test -- --test-threads 1` solves your problem with setting the current dir, please go ahead and change that.

Also please rebase on master to get green from Travis :)

nott:
@jmatraszek also confirmed this fixes the issue for him - see #189

what needs to be done here?

johannhof:
@leesdolphin I think it would be cleaner to implement the test you are adding in the [cli tests](https://github.com/cobalt-org/cobalt.rs/blob/master/tests/cli.rs) instead.

Also note that you would have to add `-- --test-threads 1` to the appveyor.yml file as well, which is probably why Windows tests are failing.

Thanks!

epage:
#227 ended up resolving the primary aim of this PR.  I've noted the value of the tests here in #284 to remind us to get back to them..

PR #134, opened 2016-08-26
includes posts recursively into `post_data'

jianingy:
Hi, all

Currently, posts in subdirectories of `./posts' are missing in the post list in index.html. This patch helps fix this problem.

I've just tried cobalt a few days. Hope it's a valid fixing.

Thanks guys

PR #185, opened 2017-02-03
Posts in subdirs

uwearzt:
This superseds pull request https://github.com/cobalt-org/cobalt.rs/pull/134 because I have completely started over the test.

After applying, pull request 134 can be dropped.

Additional Infos:

1. cargo fmt also changes src/main.rs I have not touched
2. did not run clippy, because it doesn't compile at the moment

### DIFF
--- a/src/cobalt.rs
+++ b/src/cobalt.rs
@@ -33,6 +33,13 @@ fn compare_paths(a: &Path, b: &Path) -> bool {
     }
 }
 
+fn starts_with_path(complete_path: &Path, post_path: &Path) -> bool {
+    match (fs::canonicalize(complete_path), fs::canonicalize(post_path)) {
+        (Ok(p), Ok(p2)) => p.starts_with(p2),
+        _ => false,
+    }
+}
+
 /// The primary build function that transforms a directory into a site
 pub fn build(config: &Config) -> Result<()> {
     trace!("Build configuration: {:?}", config);
@@ -72,7 +79,7 @@ pub fn build(config: &Config) -> Result<()> {
         if template_extensions.contains(extension) {
             // if the document is in the posts folder it's considered a post
             let is_post =
-                entry_path.parent().map(|p| compare_paths(p, &posts_path)).unwrap_or(false);
+                entry_path.parent().map(|p| starts_with_path(p, &posts_path)).unwrap_or(false);
 
             let new_path = entry_path.strip_prefix(source).expect("Entry not in source folder");
 
--- a/src/cobalt.rs
+++ b/src/cobalt.rs
@@ -33,8 +33,21 @@ fn compare_paths(a: &Path, b: &Path) -> bool {
     }
 }
 
-fn starts_with_path(complete_path: &Path, post_path: &Path) -> bool {
-    match (fs::canonicalize(complete_path), fs::canonicalize(post_path)) {
+/// Checks if one path is the starting point of another path. The examples are taken
+/// from tests/fixtures/posts_in_subfolder/
+///
+/// # Examples 
+///   * starts_with_path(this, starts_with) -> bool
+/// ".../tests/fixtures/posts_in_subfolder" ".../tests/fixtures/posts_in_subfolder/posts"
+///   -> false
+/// ".../tests/fixtures/posts_in_subfolder/posts" ".../tests/fixtures/posts_in_subfolder/posts"
+///   -> true
+/// ".../tests/fixtures/posts_in_subfolder/posts/20170103" ".../tests/fixtures/posts_in_subfolder/posts"
+///   -> true
+/// ".../tests/fixtures/posts_in_subfolder/posts/2017/01/08" ".../tests/fixtures/posts_in_subfolder/posts"
+///   -> true
+fn starts_with_path(this: &Path, starts_with: &Path) -> bool {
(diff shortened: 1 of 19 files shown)


johannhof:
Thanks! I'll take a closer look at this tomorrow. I will push an update to rustfmt so that your build doesn't fail because of that anymore. :)

uwearzt:
FYI, my version of rustfmt

```
cargo fmt -- --version
0.7.1 ()
```

uwearzt:
Sorry clicked on the wrong button :(

tests/fixtures/posts_in_subfolder/_layouts/default.liquid
```diff
@@ -0,0 +1,21 @@
...
+  <body>
+    <div>
+      {% if is_post %}
+        {% include '_layouts/post.liquid' %}
+      {% else %}
+        {{ content }}
+      {% endif %}
+    </div>
+  </body>
+  <!-- start default.liquid -->
```
johannhof: end?

tests/fixtures/posts_in_subfolder/README.md
```diff
@@ -0,0 +1,20 @@
+extends: default.liquid
+
+title: README for differnent directory layouts
```
johannhof: typo: different

tests/fixtures/posts_in_subfolder/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+All blog posts reside inside a single directory.
+
+## posts inside some folders
+
+For example there can be one folder per year. This makes managing of blog posts easier.
+
+## one folder per post
+
+There is one folder per blog post. Especially usefull if there are many associated files
```
johannhof: typo: useful

tests/fixtures/posts_in_subfolder/_layouts/default.liquid
uwearzt: That fix changes all html files in the directory tests/target/posts_in_subfolder too.

uwearzt:
Great catches during the review :) Thanks. Pushed the fixes to the branch.

src/cobalt.rs
```diff
@@ -33,6 +33,13 @@ fn compare_paths(a: &Path, b: &Path) -> bool {
     }
 }
 
+fn starts_with_path(complete_path: &Path, post_path: &Path) -> bool {
```
nott: this function signature looks slightly confusing to me: argument name `post_path` sounds like this is a path to a single post (`posts_path` makes more sense here).
it also may make sense to avoid `post_path`/`posts_path` completely - the function can be reused then. what about `starts_with_path(this: &Path, path: &Path)`?

could you please also add a docstring to improve readability?

uwearzt: What do you think about:

```
/// Checks if one path is the starting point of another path. The examples are taken
/// from tests/fixtures/posts_in_subfolder/
///
/// # Examples 
///   * starts_with_path(this, starts_with) -> bool
/// ".../tests/fixtures/posts_in_subfolder" ".../tests/fixtures/posts_in_subfolder/posts"
///   -> false
/// ".../tests/fixtures/posts_in_subfolder/posts" ".../tests/fixtures/posts_in_subfolder/posts"
///   -> true
/// ".../tests/fixtures/posts_in_subfolder/posts/20170103" ".../tests/fixtures/posts_in_subfolder/posts"
///   -> true
/// ".../tests/fixtures/posts_in_subfolder/posts/2017/01/08" ".../tests/fixtures/posts_in_subfolder/posts"
///   -> true
fn starts_with_path(this: &Path, starts_with: &Path) -> bool {
    match (fs::canonicalize(this), fs::canonicalize(starts_with)) {
        (Ok(p), Ok(p2)) => p.starts_with(p2),
        _ => false,
    }
}
```

uwearzt:
The failing clippy error is not from my changes. Should I try to fix them?

src/cobalt.rs
```diff
@@ -33,6 +33,31 @@ fn compare_paths(a: &Path, b: &Path) -> bool {
     }
 }
 
+/// Checks if one path is the starting point of another path.
+/// The tests are taken from tests/fixtures/`posts_in_subfolder`/
```
johannhof: I don't think tests need comments in the doc-style (`///`)

src/cobalt.rs
```diff
@@ -33,6 +33,31 @@ fn compare_paths(a: &Path, b: &Path) -> bool {
...
+#[test]
+fn test_starts_with_path() {
```
johannhof: Thanks for adding a test! Please move this test to the bottom of the file

src/cobalt.rs
```diff
@@ -33,6 +33,31 @@ fn compare_paths(a: &Path, b: &Path) -> bool {
...
+    let posts_folder = Path::new("tests/fixtures/posts_in_subfolder/posts");
+
+    assert_eq!(starts_with_path(Path::new("tests/fixtures/posts_in_subfolder"), posts_folder),
+               false);
```
johannhof: You can leave out true and false in this and the following calls and replace `assert_eq!` with `assert!`

johannhof:
Oh, I didn't see this wasn't merged yet. Here we go!

Issue #186, opened 2017-02-05
Markdown reference-style links are not resolved in excerpts (in most cases)


djwf:
In Cobalt v.0.5.0 when the excerpt is not explicitly defined in a post, any Markdown reference-style links do not resolve in excerpts unless the reference itself is immediately below or above the initial paragraph.

To be clear, the following example allows reference-style links to resolve in the first paragraph.

    title=Markdown reference-style link test
    ---
    [Markdown]: https://daringfireball.net/projects/markdown/
    Test paragraph using [Markdown][] reference-style link.

The following three examples, however, result in unresolved references.

In the first case, the excerpt ends up being completely blank.

    title=Markdown reference-style link test
    ---
    [Markdown]: https://daringfireball.net/projects/markdown/
    
    Test paragraph using [Markdown][] reference-style link.

In the second case and third case, the excerpt is displayed without the references being rendered.

    title=Markdown reference-style link test
    ---
    Test paragraph using [Markdown][] reference-style link.
    [Markdown]: https://daringfireball.net/projects/markdown/

Note that it doesn't matter if there's a line between the paragraph and reference or not, in either case references in the excerpt are not resolved.

    title=Markdown reference-style link test
    ---
    Test paragraph using [Markdown][] reference-style link.
    
    [Markdown]: https://daringfireball.net/projects/markdown/

A workaround is to move all references used in the first paragraph up to immediately precede the first paragraph, but ideally the references should work regardless of where they are placed within the file.

nott:
Makes sense

This is how the feature is implemented in jekyll: https://github.com/jekyll/jekyll/pull/5212b

PR #188, opened 2017-02-11
Ref style links in excerpts

djwf:
This should address issue #186: feedback requested, as I'm still learning Rust.

I used the [regex from Jekyll](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/excerpt.rb#L120) to match the reference-style links.

Also, instead of creating a new set of tests, I added the tests to the existing excerpts test set: let me know if that's the preferred method.

### DIFF
--- a/src/document.rs
+++ b/src/document.rs
@@ -298,7 +298,7 @@ impl Document {
     }
 
     /// Extracts references iff markdown content.
-    pub fn extract_markdown_references(&self) -> String {
+    pub fn extract_markdown_references(&self, excerpt_separator: &str) -> String {
         let mut trail = String::new();
         let re = Regex::new(r"(?m:^ {0,3}\[[^\]]+\]:.+$)").unwrap();
 
@@ -308,7 +308,11 @@ impl Document {
                 trail.push('\n');
             }
         }
-        trail
+        trail +
+        self.content
+            .split(excerpt_separator)
+            .next()
+            .unwrap_or(&self.content)
     }
 
     /// Renders excerpt and adds it to attributes of the document.
@@ -327,19 +331,15 @@ impl Document {
                 .and_then(|attr| attr.as_str())
                 .unwrap_or(default_excerpt_separator);
 
-            let excerpt = if let Some(excerpt_str) = excerpt_attr {
-                excerpt_str.to_string()
+            if let Some(excerpt_str) = excerpt_attr {
+                try!(self.render_html(excerpt_str, context, source))
             } else if excerpt_separator.is_empty() {
-                "".to_string()
+                try!(self.render_html("", context, source))
             } else {
-                self.extract_markdown_references() +
-                self.content
-                    .split(excerpt_separator)
-                    .next()
-                    .unwrap_or(&self.content)
-            };
-
-            try!(self.render_html(&excerpt, context, source))
+                try!(self.render_html(&self.extract_markdown_references(excerpt_separator),
+                                      context,
+                                      source))
+            }
         };
 
         self.attributes.insert("excerpt".to_owned(), Value::Str(excerpt_html));
--- a/src/document.rs
+++ b/src/document.rs
@@ -297,6 +297,20 @@ impl Document {
         Ok(html.to_owned())
     }
 
+    /// Extracts references from markdown content.
+    pub fn extract_references(&self) -> String {
+        let mut trail = String::new();
+        let re = Regex::new(r"(?m:^ {0,3}\[[^\]]+\]:.+$)").unwrap();
+
+        if re.is_match(&self.content) {
+            for mat in re.find_iter(&self.content) {
+                trail.push_str(&self.content[mat.0..mat.1]);
+                trail.push('\n');
+            }
+        }
+        trail
+    }
+
     /// Renders excerpt and adds it to attributes of the document.
     pub fn render_excerpt(&mut self,
                           context: &mut Context,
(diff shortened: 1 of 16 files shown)


src/document.rs
```diff
@@ -314,14 +328,18 @@ impl Document {
...
 
             let excerpt = if let Some(excerpt_str) = excerpt_attr {
-                excerpt_str
+                excerpt_str.to_string()
             } else if excerpt_separator.is_empty() {
-                ""
+                String::new()
             } else {
-                self.content.split(excerpt_separator).next().unwrap_or(&self.content)
+                self.extract_references() +
```
nott: some posts are not markdown

we can probably extract references only if the document is markdown (`self.markdown` is truthful in this case)

nott: and personally I'd like references to be added to the end of excerpt (feel free to ignore me here though)

src/document.rs
```diff
@@ -297,6 +297,20 @@ impl Document {
...
+    /// Extracts references from markdown content.
+    pub fn extract_references(&self) -> String {
+        let mut trail = String::new();
+        let re = Regex::new(r"(?m:^ {0,3}\[[^\]]+\]:.+$)").unwrap();
+
+        if re.is_match(&self.content) {
+            for mat in re.find_iter(&self.content) {
+                trail.push_str(&self.content[mat.0..mat.1]);
+                trail.push('\n');
+            }
```
nott: we can also use [itertools::join](https://docs.rs/itertools/0.5.9/x86_64-unknown-linux-gnu/itertools/fn.join.html) here

this will require adding `itertools` to `Cago.{toml,lock}`
@johannhof any objections?

nott:
@djwf thank you for working on this

i think this is gtg once you take into account non-markdown posts

src/document.rs
djwf: I forgot about the potential for liquid posts (or HTML for that matter), so I've fixed that.  Regarding the references before the excerpt (rather than after), that's how I originally had things arranged... but that required changing the type of the returned String from extract references from String to &str (outside the function, because otherwise it doesn't live long enough), changing the type of the extracted excerpt (shown below) to a String, and adding two newlines to separate the references from the excerpt.

    self.content
        .split(excerpt_separator)
        .next()
        .unwrap_or(&self.content)

My original idea was to have the extract references return the type &str, but I don't think I understand the lifetime system well enough, because I was never able to get it to work.

I would like to not have to change the type returned by the "if let" arms in this area: originally it returned &str, and I had to change to Strings to be able to concatenate the extracted references with the excerpt.  Suggestions as to how that would be done are welcome!

src/document.rs
djwf: That would be more elegant: in fact, it might not require an added function if we used join.

djwf:
@nott You're welcome!  Changes pushed to make sure references are only extracted in markdown posts.  If @johannhof decides itertools should be included, then I'll revise the extraction function.

src/document.rs
```diff
@@ -297,12 +297,12 @@ impl Document {
         Ok(html.to_owned())
     }
 
-    /// Extracts references from markdown content.
-    pub fn extract_references(&self) -> String {
+    /// Extracts references iff markdown content.
```
nott: typo: `if` instead of `iff`

johannhof: Or maybe it's https://en.wikipedia.org/wiki/If_and_only_if :)

src/document.rs
```diff
@@ -314,14 +328,18 @@ impl Document {
                 .unwrap_or(default_excerpt_separator);
 
             let excerpt = if let Some(excerpt_str) = excerpt_attr {
-                excerpt_str
+                excerpt_str.to_string()
             } else if excerpt_separator.is_empty() {
-                ""
+                String::new()
```
johannhof: It's a small thing but you could avoid the allocation here. You could either declare `let mut excerpt` and assign it in the if clauses or you call `try!(self.render_html(..., context, source))` inside each clause. Does that make sense?

src/document.rs
```diff
@@ -297,6 +297,20 @@ impl Document {
...
+    /// Extracts references iff markdown content.
+    pub fn extract_markdown_references(&self) -> String {
+        let mut trail = String::new();
+        let re = Regex::new(r"(?m:^ {0,3}\[[^\]]+\]:.+$)").unwrap();
+
+        if self.markdown && re.is_match(&self.content) {
+            for mat in re.find_iter(&self.content) {
+                trail.push_str(&self.content[mat.0..mat.1]);
+                trail.push('\n');
+            }
```
johannhof: I'm pretty indifferent to the usage of `itertools` here, you can use it if you want. We may have other places in the code where it would be useful.

src/document.rs
djwf: Yep: I actually meant `iff`, not `if`. :)

src/document.rs
djwf: As I already have it working I'll leave it be for now.  If and when itertools becomes a dependency, this section can be revisited.

src/document.rs
djwf: I tried with a number of different options in each clause, but I am not fluent enough with the vagaries of the rust type/borrowing/lifetime system, so I'll go for the simple fix.  Instead of allocating a new string, I'll just convert the `""` to a string using `"".to_string()`.

johannhof: `"".to_string()` also allocates a new string :)

What error do you get if you do `try!(self.render_html(..., context, source))` in every clause instead?

djwf: Curses! :)

Your question about the errors if I use the `try!(...)` in each clause led me to the solution.  Basically: my issue was I was leaving the semicolon terminating the `if let Some(excerpt_str) = ...` expression causing a `mismatched types` error.

Once I fixed that everything seems to work!  I'll push the new changes for review.

johannhof:
Alright, this looks good now. Thank you very much!

PR #192, opened 2017-02-25
feat(attributes): Add input-file based attributes

epage:
Add document attributes "title", "slug", and "ext" if needed based on the
input-file.

The primary motivation for this change is to allow a custom "path" while
retaining the original :slug. A more surgical approach was taken rather
than trying to achieve 100% Jekyll-compatibility though Jekyll does serve
as an inspiration for how these features work.

The "slug" attribute defaults to a lower-case slugified version of the
document's filename.  The primary use is for specifying a custom "path"
attribute that still includes the original file name (via :slug).

Jekyll compatibility:
- Precedence: Allows a user to override "slug" in their front matter.
- Template variable for path: The default slug attempts to be formatted
  like Jekyll
Jekyll incompatibility:
- General: Jekyll excludes "YYYY-MM-DD-" from the slug. Cobalt would need
  additional work done to make "date" more Jekyll compatible before this
  is worth implementing.
- Template variable for path: An overridden slug is not slugified in
  Cobalt. This is in part for simplicity and in part trusting the user.
- Attribute: Jekyll does not sluggify the attribute but instead expects
  the user to slugify it with a filter.

The "title" attribute is now defaulted to a title-cased, slugified version
of the document's filename. The primary use is to minimize user's having
to repeat themselves in their liquid document.

Jekyll compatibility:
- Precedence: Allows a user to override "title" in their front matter.
  Does not use the overridden "slug" attribute if provided.
Jekyll incompatibility:
- General: Jekyll excludes "YYYY-MM-DD-" from the title.
- Template variable for path: :title is supposed to be a "pretty" mixed
  case version of :slug
- Attribute: Jekyll does not slugify the file name before title-casing it.

The "ext" attribute defaults to the file extension of the liquid document
("md", "liquid", etc)

Jekyll compatibility:
- Precedence: Allows a user to override "ext" in their front matter.
Jekyll incompatibility:
- Attribute: Jekyll includes the ".".

The :name template variable for path (slugified file_stem, not overriding
allowed) was also going to be added but for some reason it caused a test
to fail. Since it wasn't needed for the primary purpose of this change, it
was skipped rather than time being spent on investigating the failure.

### DIFF
--- a/src/document.rs
+++ b/src/document.rs
@@ -20,7 +20,8 @@ use pulldown_cmark as cmark;
 use liquid;
 
 lazy_static!{
-    static ref DATE_VARIABLES: Regex = Regex::new(":(year|month|i_month|day|i_day|short_year|hour|minute|second)").unwrap();
+    static ref DATE_VARIABLES: Regex =
+        Regex::new(":(year|month|i_month|day|i_day|short_year|hour|minute|second)").unwrap();
     static ref SLUG_INVALID_CHARS: Regex = Regex::new(r"([^a-zA-Z0-9]+)").unwrap();
     static ref FRONT_MATTER_DIVIDE: Regex = Regex::new(r"---\s*\r?\n").unwrap();
     static ref MARKDOWN_REF: Regex = Regex::new(r"(?m:^ {0,3}\[[^\]]+\]:.+$)").unwrap();
--- a/src/document.rs
+++ b/src/document.rs
@@ -1,7 +1,6 @@
 use std::fs::File;
 use std::collections::HashMap;
 use std::collections::hash_map::Entry;
-use std::ffi::OsStr;
 use std::path::{Path, PathBuf};
 use std::default::Default;
 use error::Result;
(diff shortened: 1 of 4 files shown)


epage:
Checklist
- [x] Run "cargo test"
- [x] Unit tests added but not integration tests
- [x] Run clippy
- [x] Run rustfmt

src/document.rs
```diff
@@ -104,6 +102,32 @@ fn format_path(p: &str,
...
 }
 
+/// The base-name without an extension.  Correlates to Jekyll's :name path tag
+fn file_stem(p: &Path) -> String {
+    p.file_stem().map(|os| os.to_string_lossy().into_owned()).unwrap_or_else(|| "".to_owned())
+}
+
+/// Create a slug for a given file.  Correlates to Jekyll's :slug path tag
+fn slugify(name: &str) -> String {
+    let invalid_re = Regex::new(r"([^a-zA-Z0-9]+)").unwrap();
```
johannhof: Can you create the Regex lazily using `lazy_static`? :)

epage:
*sigh* Nightly is failing to compile clippy

johannhof:
Great, thanks, no worries about the clippy failure, we should just bump the version I guess. :)

PR #193, opened 2017-02-26
Add support for YAML->Liquid conversion of dicts

silven:
YAML dictionaries are BTreeMap<yaml::Yaml::Value, yaml::Yaml::Value> 
and Liquid dictionaries are HashMap<String, liquid::Value>.

This commit adds support for converting from the yaml definition to
the liquid definition of a dictionary data type.

### DIFF
--- a/src/document.rs
+++ b/src/document.rs
@@ -1,5 +1,5 @@
 use std::fs::File;
-use std::collections::HashMap;
+use std::collections::{HashMap, BTreeMap};
 use std::collections::hash_map::Entry;
 use std::ffi::OsStr;
 use std::path::{Path, PathBuf};
@@ -39,6 +39,19 @@ fn read_file<P: AsRef<Path>>(path: P) -> Result<String> {
     Ok(text)
 }
 
+fn yaml_btreemap_to_hashmap(input: &BTreeMap<Yaml, Yaml>) -> HashMap<String, Value> {
+    let mut result = HashMap::new();
+    for (yaml_key, yaml_value) in input {
+        if let Yaml::String(ref s) = *yaml_key {
+            let liquid_value = yaml_to_liquid(yaml_value).expect("Value must be liquid data");
+            result.insert(s.to_owned(), liquid_value);
+        } else {
+            panic!("Key in yaml dictionary must be string");
+        }
+    }
+    result
+}
+
 fn yaml_to_liquid(yaml: &Yaml) -> Option<Value> {
     match *yaml {
         Yaml::Real(ref s) |
@@ -46,8 +59,48 @@ fn yaml_to_liquid(yaml: &Yaml) -> Option<Value> {
         Yaml::Integer(i) => Some(Value::Num(i as f32)),
         Yaml::Boolean(b) => Some(Value::Bool(b)),
         Yaml::Array(ref a) => Some(Value::Array(a.iter().filter_map(yaml_to_liquid).collect())),
+        Yaml::Hash(ref dict) => Some(Value::Object(yaml_btreemap_to_hashmap(dict))),
         Yaml::BadValue | Yaml::Null => None,
         _ => panic!("Not implemented yet"),
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::collections::{HashMap, BTreeMap};
+    use yaml_rust as yaml;
+    use liquid;
+
+    #[test]
+    fn test_yaml_to_liquid() {
+        let key_str = "key".to_owned();
+        let input = {
+            let mut temp = BTreeMap::new();
+            temp.insert(yaml::Yaml::String(key_str.clone()), yaml::Yaml::Integer(42));
+            yaml::Yaml::Hash(temp)
+        };
+        let expected = {
+            let mut temp = HashMap::new();
+            temp.insert(key_str.clone(), liquid::Value::Num(42.0));
+            Some(liquid::Value::Object(temp))
+        };
+        assert_eq!(super::yaml_to_liquid(&input), expected);
+    }
+
+    #[test]
+    fn test_dictionary_conversion() {
+        let key_str = "key".to_owned();
+        let input = {
+            let mut temp = BTreeMap::new();
+            temp.insert(yaml::Yaml::String(key_str.clone()), yaml::Yaml::Integer(42));
+            temp
+        };
+        let expected = {
+            let mut temp = HashMap::new();
+            temp.insert(key_str.clone(), liquid::Value::Num(42.0));
+            temp
+        };
+        assert_eq!(super::yaml_btreemap_to_hashmap(&input), expected);
     }
 }
 
--- a/tests/fixtures/yaml_dicts/_layouts/index.liquid
+++ b/tests/fixtures/yaml_dicts/_layouts/index.liquid
@@ -1,0 +1,8 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>My blog - {{ title }}</title>
+</head>
+<body>
+{{ content }}</body>
+</html>
--- a/tests/fixtures/yaml_dicts/index.md
+++ b/tests/fixtures/yaml_dicts/index.md
@@ -1,0 +1,14 @@
+extends: index.liquid
+
+title:   My first Blogpost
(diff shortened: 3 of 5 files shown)


src/document.rs
```diff
@@ -39,18 +39,71 @@ fn read_file<P: AsRef<Path>>(path: P) -> Result<String> {
     Ok(text)
 }
 
+fn yaml_btreemap_to_hashmap(input: &BTreeMap<Yaml, Yaml>) -> HashMap<String, Value> {
```
silven: I'm not sure of the naming convention here. Please suggest improvements.

src/document.rs
```diff
@@ -39,18 +39,71 @@ fn read_file<P: AsRef<Path>>(path: P) -> Result<String> {
...
+    let mut result = HashMap::new();
+    for (yaml_key, yaml_value) in input {
+        if let Yaml::String(ref s) = *yaml_key {
+            let liquid_value = yaml_to_liquid(yaml_value).expect("Value must be liquid data");
+            result.insert(s.to_owned(), liquid_value);
+        } else {
+            panic!("Key in yaml dictionary must be string");
+        }
+    }
+    result
```
silven: I prefer "return result" in this situation, but Clippy suggested against it. And the contributors file said to use Clippy.

src/document.rs
```diff
@@ -39,18 +39,71 @@ fn read_file<P: AsRef<Path>>(path: P) -> Result<String> {
...
+        };
+        let expected = {
+            let mut temp = HashMap::new();
+            temp.insert(key_str.clone(), liquid::Value::Num(42.0));
+            temp
+        };
+        assert_eq!(super::yaml_btreemap_to_hashmap(&input), expected);
+    }
+}
+
```
silven: The simplest test cases I could come up with.

tests/fixtures/yaml_dicts/_layouts/index.liquid
```diff
@@ -0,0 +1,8 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>My blog - {{ title }}</title>
+</head>
+<body>
+{{ content }}</body>
+</html>
```
silven: Indentation is kind of messy when doing these kinds of tests. It took like half an hour just to get all the spaces to match.

tests/fixtures/yaml_dicts/index.md
```diff
@@ -0,0 +1,14 @@
+extends: index.liquid
+
+title:   My first Blogpost
+date:    24/08/2014 at 15:36
+some-data:
+  - nested:
+      dicts: Works
+  - nested:
+      dicts: just fine
+---
```
silven: This is the functionality I wish to have when using Cobalt. Structuring data in yaml like this is a great tool. If there is a "better" (more idiomatic) way of doing this, please let me know.

silven:
1) rustup and nightly is broken at the moment.
2) Does this warrant a bump in version number? Should this commit do that, or do we do that later?

epage:
I'm sorry you've only gotten radio silence on this so far.  I appreciate you taking the time to do this.  I didn't even realize cobalt doesn't support this yet.

> My first addition to an open source project in a long, long time. I hope more than I find this readable and useful.

I think its great.  I've only recently gotten back into open source development.

One suggestion is to try to focus a bit more on the user impact in the commit message.  I misunderstood the intent of the change until I looked at the diff.  For example "Extend frontmatter to support nested YAML structures".

> rustup and nightly is broken at the moment.

Yeah, intermittent CI is not ideal.  We should probably make nightly advisory-only.

tests/fixtures/yaml_dicts/index.md
epage: I think this is a reasonable feature for some specific use cases.  When a user's data becomes more complex or needs to be shared across files, we'd probably want to implement something like Jekyll's [data files](https://jekyllrb.com/docs/datafiles/) feature.

tests/fixtures/yaml_dicts/_layouts/index.liquid
epage: Agreed that the testing is less than ideal right now.  We'd like to move more towards unit testing rather than end-to-end testing but we need to decouple some of the pieces so we can unit test them.

src/document.rs
```diff
@@ -39,18 +39,71 @@ fn read_file<P: AsRef<Path>>(path: P) -> Result<String> {
...
+    fn test_yaml_to_liquid() {
+        let key_str = "key".to_owned();
+        let input = {
+            let mut temp = BTreeMap::new();
+            temp.insert(yaml::Yaml::String(key_str.clone()), yaml::Yaml::Integer(42));
+            yaml::Yaml::Hash(temp)
+        };
+        let expected = {
+            let mut temp = HashMap::new();
+            temp.insert(key_str.clone(), liquid::Value::Num(42.0));
```
epage: Tip: you can avoid an allocation by having `key_str`'s type be a `&str`
```rust
let key_str = "key";
...
temp.insert(yaml::Yaml::String(key_str.to_owned()), yaml::Yaml::Integer(42));
...
temp.insert(key_str.to_owned(), liquid::Value::Num(42.0));
```

src/document.rs
```diff
@@ -39,18 +39,71 @@ fn read_file<P: AsRef<Path>>(path: P) -> Result<String> {
...
+            Some(liquid::Value::Object(temp))
+        };
+        assert_eq!(super::yaml_to_liquid(&input), expected);
+    }
+
+    #[test]
+    fn test_dictionary_conversion() {
+        let key_str = "key".to_owned();
```
epage: Tip: Ditto about avoiding an allocation

src/document.rs
epage: I can understand this feeling weird.  The Rust community's coding conventions are very terse and while I don't agree with it in some cases like this, I'm slowly getting used to it.

epage:
> Does this warrant a bump in version number? Should this commit do that, or do we do that later?

My guess is to not worry about it.  I don't know what cobalt's lead's release process looks like but so far he hasn't had me bump the version for any of my changes.

src/document.rs
epage: Preface: I'm still experimenting with Rust by pushing the language conventions to the limit so I can learn their strengths and weaknesses.  This might actually be a bad idea :)

You might not even need a function here.  In my experimenting with TOML support (mostly for debug dumps), I found doing the conversion the "functional" way felt clean to me (inmates running the asylum?).  Longer term, we should probably move this stuff to liquid-rs itself and take advantage of serde to do so.

For an idea of how this might look, here is my toml code:
```rust
        liquid::Value::Object(ref t) => {
            let toml_object: Result<BTreeMap<String, toml::Value>> = t.iter()
                .map(|(k, v)| convert_value(v).map(|v| (k.to_string(), v)))
                .collect();
            let toml_object = toml_object?;
            Ok(toml::Value::Table(toml_object))
        }
```

A simpler way for handling yaml's flexible keys is
```rust
yaml_key.as_str().expect("Keys must be strings").to_owned()
```

Ideally, we'd switch this function to returning a Result so we could avoid the panic/expect.

src/document.rs
```diff
@@ -39,18 +39,71 @@ fn read_file<P: AsRef<Path>>(path: P) -> Result<String> {
...
+        let expected = {
+            let mut temp = HashMap::new();
+            temp.insert(key_str.clone(), liquid::Value::Num(42.0));
+            Some(liquid::Value::Object(temp))
+        };
+        assert_eq!(super::yaml_to_liquid(&input), expected);
+    }
+
+    #[test]
+    fn test_dictionary_conversion() {
```
epage: Personally, I'd recommend testing dictionary conversion at the  `yaml_to_liquid` level to make this test more refactor-proof.

src/document.rs
epage: Thanks for adding some coverage of this code.  We're relying too much on end-to-end testing currently and its causing maintenance problems.

epage:
Liquid now has serde support, see cobalt-org/liquid-rust#91.  Once that gets released and cobalt gets updated to rely on it, we should get this "for free".

PR #203, opened 2017-03-21
fix(frontmatter): Support empty frontmatters

badboy:
On trying to port my site over, I encoutered crashes due to invalid index access with files that have a front matter such as:

```
---
name: ...
---
```

This gets parsed as an empty front matter due to beginning with dashes, which lead to the index error.

I'm not sure if empty frontmatter = error is the right thing or if we just want to treat it without problems.

### DIFF
--- a/src/document.rs
+++ b/src/document.rs
@@ -177,21 +177,20 @@ impl Document {
 
             let yaml_result = try!(YamlLoader::load_from_str(attribute_split));
 
-            if yaml_result.is_empty() {
-                return Err(format!("Incorrect front matter format in {:?}", file_path).into());
-            }
+            if !yaml_result.is_empty() {
+                let yaml_attributes = try!(yaml_result[0]
+                    .as_hash()
+                    .ok_or_else(|| format!("Incorrect front matter format in {:?}", file_path)));
 
-            let yaml_attributes = try!(yaml_result[0]
-                .as_hash()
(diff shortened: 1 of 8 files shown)


epage:
Sorry for not getting back to you on this.

Thank you for taking the time to address this.  

> I'm not sure if empty frontmatter = error is the right thing or if we just want to treat it without problems.

I'm leaning towards gracefully handling it because Cobalt can or "soon" will be able to handle all required attributes being defaulted.

Also, would you be willing to add a test case for this to prevent us from breaking this again in the future?  If not, thats ok,  I'll try to get to it.

For now, the easiest way to test this is adding to [mod.rs](https://github.com/cobalt-org/cobalt.rs/blob/master/tests/mod.rs#L83)
```
#[test]
pub fn empty_frontmatter() {
    run_test("empty_frontmatter").expect("Build error");
}
```
Put a cobalt project in `tests/fixtures/empty_frontmatter` that has an empty front matter.  Put an expected output in `tests/target/empty_frontmatter`.

epage:
As for the CI failure, that was due to changes in Appveyor that have been fixed in master.  I'd recommend pulling down the latest changes from master so your branch can build without problems.

badboy:
Rebased against master, made it handle no frontmatter gracefully (simply not trying to extract any data from it) and added the test as advised.

epage:
Thanks for contributing!

PR #204, opened 2017-03-24
chore: Upgrade dependencies

epage:
Upgrade notes:
- notify: Debouncing was added but have not yet switched to it because of
  the extra work to extract the relevant path

Dependencies that were held back:
- assert_cli: See killercup/assert_cli#30 and killercup/assert_cli#31
- rss: Seems they turned off XML escaping.  Didn't have time to research
  this.


johannhof:
Cool, thanks again!

PR #205, opened 2017-03-25
fix(appveyor): Get a proper path for curl

epage:


johannhof:
Great! Builds passing! Thanks!

Issue #208, opened 2017-04-07
Create a new post or new page using a CLI command.


thowenzel:
It would be nice if the user can create a new post using a CLI command. Something like this:
`cobalt new post <title>`

This could create a new markdown-file withe the given name. 
I would like an handling similar to Hexo (https://hexo.io/docs/writing.html).

Are there any plans to implement something like this?

bddenhartog:
I'd like a feature like this too, and would be happy to work on it myself. Right now, however, the `new` sub-command is used to generate a new cobalt project, not a component of a project (like a post, or page, or what-have-you).

Thought will have to be given to how to handle this; one idea I have is to "rename" the new-project-generation subcommand from `new` to one of `gen`, `generate`, or (my preferred) `init`. The `init` command follows the pattern of other tools, like `git`, `hg`, `npm`, etc. The other option, of course, is to use a name akin to `create` for the new subcommand for handling components of the cobalt project, e.g.:

    cobalt create post <title>
    cobalt create page <title>
    cobalt create page --template <template-name> <title>

This approach may be preferred, as it doesn't change existing workflows (people may have `cobalt new` tied in with certain automation tasks).

thowenzel:
@bddenhartog Thanks, I like the approach your described. For me, the best solution iIn the sense of correct semantics) would be a rename of the new-project-generation subcommand to `init`. I think in this early stage of cobalt.rs is such a fundamental change still possible.

epage:
Something that might be interesting for this is to open up cobalt to have subcommand plugins like cargo.  It might even be worth creating a new crate for subcommand plugins.  My main concern is how well it plays with clap.

See https://github.com/rust-lang/cargo/blob/master/src/bin/cargo.rs

epage:
A related and useful command would be `cobalt publish`.  It could make the file no longer a draft and update the timestamp.

epage:
#211 proposes we have a "skel" directory for users to place their skeleton files for creating new documents
- Should we initialize "skel" on "cobalt init" with default files or have a fallback to built-in versions if the name isn't overriden in "skel"?
  - Initializing it is more discoverable of what a user can customize
  - Users overriding our built-in skeletons is better for version migration. They just suddenly become available rather than having to `cobalt init` and copy the files over or creating a new migration command/option (though maybe we should do that anyways.

epage:
If I do `cobalt new` from a "posts" folder, should it detect that and default to the "post" skeleton?

[...]

PR #209, opened 2017-04-08
Refactor default templates directory to '_templates'

bddenhartog:
The name `_layouts` is a bit confusing, as what really exists in the
`_layouts` directory is file templates, to be used in pages or posts.
This patch refactors `Config.layouts` to `Config.templates`, and all
reference values from `_layouts` to `_templates`.

Additionally, extraneous backticks (in the README) for inline
preformatted text were fixed up to use the minimal requirement of a
single backtick.


README.md
```diff
@@ -1,6 +1,6 @@
-# ![Cobalt](https://raw.githubusercontent.com/cobalt-org/logos/master/cobald.logo.02.resize.png)  
+# ![Cobalt](https://raw.githubusercontent.com/cobalt-org/logos/master/cobald.logo.02.resize.png)
```
bddenhartog: Automatic whitespace removal from my editor.

README.md
```diff
@@ -1,6 +1,6 @@
...
 [![](https://img.shields.io/crates/v/cobalt-bin.svg?maxAge=25920)](https://crates.io/crates/cobalt-bin)
-[![](https://travis-ci.org/cobalt-org/cobalt.rs.svg?branch=master)](https://travis-ci.org/cobalt-org/cobalt.rs) [![](https://ci.appveyor.com/api/projects/status/gp2mmvk8dpe8wsmi/branch/master?svg=true)](https://ci.appveyor.com/project/johannhof/cobalt-rs/branch/master) 
+[![](https://travis-ci.org/cobalt-org/cobalt.rs.svg?branch=master)](https://travis-ci.org/cobalt-org/cobalt.rs) [![](https://ci.appveyor.com/api/projects/status/gp2mmvk8dpe8wsmi/branch/master?svg=true)](https://ci.appveyor.com/project/johannhof/cobalt-rs/branch/master)
```
bddenhartog: Automatic whitespace removal from my editor.

bddenhartog:
There are a couple of compiler issues I'm aware of, currently traveling but will be working through them.

src/config.rs
```diff
@@ -29,8 +29,8 @@ impl Default for Config {
         Config {
             source: "./".to_owned(),
             dest: "./".to_owned(),
-            layouts: "_layouts".to_owned(),
-            drafts: "_drafts".to_owned(),
+            templates: "templates".to_owned(),
+            drafts: "drafts".to_owned(),
```
epage: Reason you are removing the leading "_"?  I believe those tell cobalt to ignore these folders so they don't get processed or copied over as assets?

epage:
I think the rename makes sense.  My only concern is with how this might play our with other features.

`layouts` was probably copied from Jekyll, describing how to layout the content.  It gives a little bit more "why" to the "what" of templates but as you mentioned, it can be confusing.

My concern is how this might play out with a fix for #208.  We might want to allow people to create their own templates for a `cargo new`.  So how do we distinguish between source document templates and layout templates?

src/config.rs
bddenhartog: Ah, I see that you are correct - [`/src/cobalt.rs L23-L25`](https://github.com/cobalt-org/cobalt.rs/blob/master/src/cobalt.rs#L23-L25). I'll add an underscore to `templates`, but -- off-topic -- I wonder if this is really the best way to handle this? Using a pattern for ignoring could cause problems for the user; usually things like this are handled via an explicit list or from a configuration value, like the `ignored` property in `.cobalt.yml`. What are your thoughts? This conversation should probably be in its own issue.

`_drafts` actually shouldn't be changed here, as this PR is strictly for `_layouts -> templates`. I'll remove this change.

bddenhartog:
> My concern is how this might play out with a fix for #208. We might want to allow people to create their own templates for a cargo new. So how do we distinguish between source document templates and layout templates?

I'm not sure I'm following you on this, but perhaps some conversation in regards to that issue has happened outside of the issue page - I don't see discussion about so-called "document templates", rather just creating some default items (pages, posts, template files) with some default boilerplate added in.

_Although I do like the idea of allowing user-defined document templates to be used as the source for `cobalt new <page|post|template|etc...>`, so..._

Surely these "document templates" would exist in the repository, in their own subdirectory. They are functionally different from `templates`, and files that serve the same purpose are canonically known as _skeleton files_. If we are to support these _skeleton files_, they should probably live (in the scope of the project) within `{.,_}skel`, or even in `.cobalt/skel` (where `.cobalt/` is a single subdirectory containing build/cli-related files).

epage:
> I'm not sure I'm following you on this, but perhaps some conversation in regards to that issue has happened outside of the issue page - I don't see discussion about so-called "document templates", rather just creating some default items (pages, posts, template files) with some default boilerplate added in.

Sorry if I wasn't clear.  There wasn't any extra conversation.  My comment was an extrapolation from #208 especially in the context of that issue pointing to [Hexo](https://hexo.io/docs/writing.html#Scaffolds) which includes a "scaffolds" features.

As an aside, I feel Hexo takes a very distinct approach from Jekyll.  I need to set aside some time to read up more on it to see what we can learn from it.

>  They are functionally different from templates, and files that serve the same purpose are canonically known as skeleton files. 

"Template" is a very board word and I've also seen it applied to what you define as a skeleton file.  I like the idea of calling them "skeletons" for #208 since it avoids an overly broad word (and seems clearer than Hexo's "scaffold").

The board nature of "template" is one mark against moving away from a more specific name like "layouts" but I'm not a web dev.  If there is a more specific meaning to the word "template" in the web community then it makes sense.

src/config.rs
epage: I think thats a good idea to split this into another issue.

> Personally, I think moving all build/cli-related files into a single .cobalt subdirectory allows us to add new tooling and features - like skeleton files - without having to worry about polluting the project space. It makes it clear what the user is responsible for maintaining within their project, and where they can go to modify how cobalt works in regards to their project.

Just to check, is your comment about a ".cobalt/" meant to be related to avoiding hiding files with a "_" prefix?

epage:
@johannhof could you comment on what direction you think is right here?  I don't want to overstep especially since I'm still trying to calibrate myself on what direction you all want to take this project.

src/config.rs
bddenhartog: I've made a new issue to continue this discussion: #211

bddenhartog:
> "Template" is a very board word and I've also seen it applied to what you define as a skeleton file. I like the idea of calling them "skeletons" for #208 since it avoids an overly broad word (and seems clearer than Hexo's "scaffold").

Right, I was talking specifically about "templates" in the context of this PR, which I suppose is to say "`build` templates". Glad we agree that "skeleton" files makes sense 👍

bddenhartog:
@epage @johannhof I hate to do this, but could either of you look at the `posts_in_subfolder` and `excerpts` tests and see if you can spot why they're failing? This is a pretty straightforward change, but from what I can see it looks like the fixtures are being rendered incorrectly.

epage:
Something strange is going on.  I've tried to clone your branch on my machine but git can only see your master branch.  I'm still getting used to using git/github in a collaborative environment, so maybe I'm missing something.

```
git clone https://github.com/bddenhartog/cobalt.rs.git
git branch
master
```
and
```
git clone https://github.com/bddenhartog/cobalt.rs.git --branch chore/refactor-layouts-dir-to-template --single-branch
Cloning into 'cobalt.rs'...
warning: Could not find remote branch chore/refactor-layouts-dir-to-template to clone.
fatal: Remote branch chore/refactor-layouts-dir-to-template not found in upstream origin
```

bddenhartog:
> Something strange is going on. I've tried to clone your branch on my machine but git can only see your master branch. I'm still getting used to using git/github in a collaborative environment, so maybe I'm missing something.
> 
> git clone https://github.com/bddenhartog/cobalt.rs.git
> git branch
> master
> and
> 
> git clone https://github.com/bddenhartog/cobalt.rs.git --branch chore/refactor-layouts-dir-to-template --single-branch
> Cloning into 'cobalt.rs'...
> warning: Could not find remote branch chore/refactor-layouts-dir-to-template to clone.
> fatal: Remote branch chore/refactor-layouts-dir-to-template not found in upstream origin

1. When you first clone a repository, you don't automatically check out all of the remote branches locally. In your first attempt, you could have _seen_ the branch by running `git branch --all`, and you could have _checked it out_ with tracking set up by executing `git checkout chore/refactor-layouts-dir-to-templates`.

2. You have a typo in the branch name; `chore/refactor-layouts-dir-to-template` should be `chore-refactor-layouts-dir-to-template**s**`.

Irregardless of these two things, cloning `cobalt-org/cobalt.rs` and executing `cargo test` on `master` results in the 15 failures I listed above.

epage:
Weird, I only see one failure on Windows 10 (posts_in_subfolder).  The failure I'm seeing makes it seem like its a problem in liquid (randomly inserting or re-arranging spacing) and almost acts as if Liquid has some kind of undefined behavior.

bddenhartog:
> Weird, I only see one failure on Windows 10 (posts_in_subfolder). The failure I'm seeing makes it seem like its a problem in liquid (randomly inserting or re-arranging spacing) and almost acts as if Liquid has some kind of undefined behavior.

Is this from a new clone of `cobalt-org/cobalt.rs`, using the `cobalt` version `0.5.0`?

epage:
Sorry, not reading close enough.

The error I reported was with your branch.  I see no errors on master, even after doing a `cargo update`.

> Quite honestly, the way tests are written for this project are a bit funky. I think we could vastly improve upon this by simply managing the entire test within the test function, instead of relying on file-based fixtures. Wouldn't it be enough to call cobalt::build, walk the tree, assert that the tree matches an expected tree (built inside the test function)?

Let me see if I understand.  Currently, most testing is end-to-end testing by running the the CLI against some fixed directories and checking the results against some expected data.  One area where I've been impacted by the focus on end-to-end testing is when upgrading deps and syntect makes changes to its output, causing me to needlessly update tests.

What you are proposing is to call the cobalt library (specifically `cobalt::build`).  As the code stands now, it seems like that would still require the filesystem input / expected folders just like now.

My collections refactor (not been getting enough time for it) includes making `cobalt::build`'s functionality more modular which should help us in unit test some of the features rather than trying to get every feature with end-to-end tests.  Depending on which is easier, it might still involve some filesystem stuff but less of a pain to keep up-to-date.

epage:
Sorry for being quiet.  To improve the debugability of things, something I'm playing with is cobalt dumping its liquid context and template and creating a command line program to process it.

I've started on the command line tool.  I hope to have enough time to finish it soon.  I'd love to add it to liquid-rs directly but I don't want to add clap et al as dependencies and so far I've not seen a way for separate bin/lib dependencies.

epage:
Alright, finished the liquid-rs command line.  Haven't tested it too much but then again, its Rust :)

https://github.com/epage/liquidate

bddenhartog:
> What you are proposing is to call the cobalt library (specifically cobalt::build). As the code stands now, it seems like that would still require the filesystem input / expected folders just like now.

Yes, from my viewpoint, we don't have a need for E2E tests when testing the builder - these become cumbersome and confusing to maintain. The only real E2E tests we need are to test `cobalt::serve`, to ensure certain endpoints are reachable post-file-parsing.

`liquidate` seems great, but I'm unsure of its' application here given that the larger issue is the inability to have test consistency between platforms. I think I'll work on refactoring the tests and get that merged in prior to the open PRs.

epage:
> `liquidate` seems great, but I'm unsure of its' application here given that the larger issue is the inability to have test consistency between platforms. I think I'll work on refactoring the tests and get that merged in prior to the open PRs.

`liquidate` is more to help us figure out the immediate failure to unblock you.  Refactoring the code to make it more testable is a little bigger of an endeavor which I'm hoping my other changes will help with.

epage:
Ha! I went through all the effort to create my debug tools, update your branch to the latest master, and start debugging just to realize the failure I'm seeing (`posts_in_subfolder`) is because the proposed change stripped trailing whitespace from `tests/target/posts_in_subfolder/README.html`.

The diff from the test gives misleading output.  To be fair, diffing trailing whitespace is annoying.

src/cobalt.rs
```diff
@@ -53,11 +53,11 @@ pub fn build(config: &Config) -> Result<()> {
         .map(OsStr::new)
         .collect();
 
-    let layouts = source.join(&config.layouts);
-    let mut layouts_cache = HashMap::new();
+    let templates_dir = source.join(&config.templates);
+    let mut templates_cache_dir = HashMap::new();
```
epage: Shouldn't this either be `templates_cache` or `templates_dir_cache`?

bddenhartog: I renamed this variable from `layouts_cache` to `templates_cache_dir`, because, canonically, it is the "cache directory for templates". Although if you think `templates_cache` makes more sense and is more in line with the rest of the codebase, this is obviously a simple thing to change.

bddenhartog:
> Ha! I went through all the effort to create my debug tools, update your branch to the latest master, and start debugging just to realize the failure I'm seeing (posts_in_subfolder) is because the proposed change stripped trailing whitespace from tests/target/posts_in_subfolder/README.html.
> 
> The diff from the test gives misleading output. To be fair, diffing trailing whitespace is annoying.

Yikes, that seems like something that shouldn't be compared. I'll get those fixed shortly.

epage:
> Yikes, that seems like something that shouldn't be compared. I'll get those fixed shortly.

On the other hand., by not comparing them means that we aren't guaranteeing that we maintain the user's input exactly.   I'm mixed.

My thought was for an editorconfig file to be created to say not to modify files in fixtures/targets.  That means I'd have to actually get around to configuring editorconfig...

src/cobalt.rs
epage: >  canonically, it is the "template cache directory" 

I think I'm missing something.  Which part are you referring to as a directory?

On disk, there is a "templates" directory.  We load individual templates from that and cache them in memory in this variable.  So that'd make the variable a cache of templates (`templates_cache`) or a cache of the templates directory (`templates_dir_cache`).

bddenhartog:
@epage to add to the wonderful world of our tests, it seems that executing `cargo test` on the `master` branch fails on 64-bit linux (kernel v `4.10.12-200`), too: https://gist.github.com/bddenhartog/aa8054d4e4ec25f5a64754c9237eedb4

epage:
> I really think refactoring our tests should be a priority.

I agree.  I think one of the important steps to improving the tests is refactoring the code to make it more testable.  My hope is my (slow) work to refactor us into supporting collections will decouple the code enough to make it easier to test.

For example on the benefits of refactoring, consolidating our directory walking logic into `Files` made it so I could write [unit tests for it without even touching the file system](https://github.com/cobalt-org/cobalt.rs/blob/master/src/files.rs#L120) though I did [add an integration test](https://github.com/cobalt-org/cobalt.rs/tree/master/tests/fixtures/hidden_files) and modified several others.

bddenhartog:
I've deleted the branch from my fork and will be resubmitting this patch shortly.

The tests are such an issue that a user on Windows who has [`core.autocrlf`](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#__code_core_autocrlf_code) set to `true` (a very common practice for Windows users who work with developers on *nix systems) has failing tests immediately after cloning the repository.

I've been pretty busy lately, but may have time to do a large-scale overhaul of our tests in the next month or so.

epage:
> The tests are such an issue that a user on Windows who has core.autocrlf set to true (a very common practice for Windows users who work with developers on *nix systems) has failing tests immediately after cloning the repository.

Initially I wanted to jump in and solve this with `.gitattributes` but now I'm curious why this fails.  In theory, autocrlf would modify the fixture and target, making it so it'd still pass.  This implies we are manipulating line endings or inserting line endings which I doubt is intended.

Could you open an issue on this so we can avoid losing track of this conversation?  Please also open an issue on other problems with tests being brittle.

epage:
From #211 

> RE: layouts -> templates
> I'm leaning against this. I feel like layout is more descriptive and template would be better suited for describing the templating done (ie Liquid), see #331.
> ...
> While I am going to go ahead and close this, I won't say its the final word. If someone wants to contend any of these points, you are welcome to do so. I am trying to get the major breaking changes out of the way soon (see milestones), so time is relatively short (weeks ideally).

PR #212, opened 2017-04-20
 Upgrade pulldown-cmark to 0.0.14

iovxw:
The old test failed because it was wrong

the old test:
old_test

new test:
new_test


epage:
The travis failure is because of clippy failures.  I'm assuming these are from a new version of clippy since you didn't change that code.  The reason for the inconsistency with appveyor is it appears we don't run clippy on appveyor.

epage:
Started on #213 to unblock you.

epage:
Drat, kicking travis off again didn't try to run your patch against the latest trunk.  Barring a dummy commit on your part, I'm not sure how to get it to re-run.  I'm pretty sure all the failures are addressed and unrelated to your change, so I'm going to go ahead and submit this.

Thank you for taking the time to upgrade this.  I'm sorry for our CI being a hassle.

PR #214, opened 2017-04-23
feat(debug): Dump intermediate state

epage:
This adds a new command line flag --dump. 

Currently the only option for what you can dump is `Liquid`.  This will cause `_<file>.toml` and `_<file>.liquid` to be created in the `dest` folder next to `<file>.html`.  A user can see what variables are available.  A developer can see if cobalt is feeding liquid the right data and start debugging liquid issues with the tool from cobalt-org/liquid-rust#83.


epage:
Replaced by #238

Issue #207, opened 2017-03-31
Server can't handle querystrings in CSS link


whostolemyhat:
I added basic [cachebusting](https://css-tricks.com/strategies-for-cache-busting-css/) to my CSS file in the `default.liquid` template:

```<link rel="stylesheet" href="/css/main.css?v=1">```

but when running `cobalt watch` or `cobalt serve` the CSS file returns a 404.

Removing the querystring `?v=1` from the CSS URL means `cobalt watch` correctly serves the file.

In addition, with the cachebusting in place, building with Cobalt then serving the static files from `/build` with another static server (in this case [http-server](https://www.npmjs.com/package/http-server)) also behaves correctly.

[Example repo](https://github.com/dontexpectanythingsensible/blog) - `_layouts/default.liquid`

epage:
Fixed by #217

PR #217, opened 2017-05-09
Ignore querystrings

whostolemyhat:
Fix for #207 - strips querystrings from urls when serving files locally

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -375,7 +375,7 @@ fn build(config: &Config) {
 
 fn static_file_handler(dest: &str, req: Request, mut res: Response) -> IoResult<()> {
     // grab the requested path
-    let req_path = match req.uri {
+    let mut req_path = match req.uri {
         RequestUri::AbsolutePath(p) => p,
         _ => {
             // return a 400 and exit from this request
@@ -389,12 +389,15 @@ fn static_file_handler(dest: &str, req: Request, mut res: Response) -> IoResult<
     // strip off any querystrings so path.is_file() matches
     // and doesn't stick index.html on the end of the path
     // (querystrings often used for cachebusting)
-    let stripped_path = &req_path.split('?').collect::<Vec<&str>>()[0];
+    match req_path.rfind('?') {
+        Some(position) => req_path.truncate(position),
+        None => (),
+    }
 
     // find the path of the file in the local system
     // (this gets rid of the '/' in `p`, so the `join()` will not replace the
     // path)
-    let path = PathBuf::from(dest).join(&stripped_path[1..]);
+    let path = PathBuf::from(dest).join(&req_path[1..]);
 
     let serve_path = if path.is_file() {
         // try to point the serve path to `path` if it corresponds to a file
--- a/src/main.rs
+++ b/src/main.rs
@@ -386,10 +386,15 @@ fn static_file_handler(dest: &str, req: Request, mut res: Response) -> IoResult<
         }
     };
 
+    // strip off any querystrings so path.is_file() matches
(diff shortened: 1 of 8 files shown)


whostolemyhat:
Travis is failing on `cargo fmt` - I ran the formatter on my code but this produced more errors, so not sure where to go from here.

epage:
Thanks for contributing!  I probably won't get a chance to review this for a day or two.

> Travis is failing on cargo fmt - I ran the formatter on my code but this produced more errors, so not sure where to go from here.

I just ran into the same problem with liquid.  The CI code uses a fixed version of rustfmt which is easy to get out of sync with.

https://github.com/cobalt-org/cobalt.rs/blob/master/ci/install.sh#L7

One option is to include updating the version of rustfmt as part of your change.  I took the quick and dirty route in liquid and am using cargo to install rustfmt but that slows down the CI.  The problem with that is that I want to be respectful of travis/appveyor providing these services for us.

src/main.rs
```diff
@@ -386,10 +386,17 @@ fn static_file_handler(dest: &str, req: Request, mut res: Response) -> IoResult<
...
+    // and doesn't stick index.html on the end of the path
+    // (querystrings often used for cachebusting)
+    let stripped_path = &req_path.split("?").collect::<Vec<&str>>()[0];
+    
     // find the path of the file in the local system
     // (this gets rid of the '/' in `p`, so the `join()` will not replace the
     // path)
-    let path = PathBuf::from(dest).join(&req_path[1..]);
+    let path = PathBuf::from(dest).join(&stripped_path[1..]);
+
```
epage: Rather than `split`/`collect`, what if we
- `rfind` + `truncate`
- `rsplitn(2, "?").last().expect("always at least one element")`

src/main.rs
```diff
@@ -386,10 +397,15 @@ fn static_file_handler(dest: &str, req: Request, mut res: Response) -> IoResult<
...
+    // strip off any querystrings so path.is_file() matches
+    // and doesn't stick index.html on the end of the path
+    // (querystrings often used for cachebusting)
+    let stripped_path = &req_path.split('?').collect::<Vec<&str>>()[0];
+
     // find the path of the file in the local system
     // (this gets rid of the '/' in `p`, so the `join()` will not replace the
     // path)
-    let path = PathBuf::from(dest).join(&req_path[1..]);
+    let path = PathBuf::from(dest).join(&stripped_path[1..]);
```
epage: Rather than `split`/`collect`, what if we
- `rfind` + `truncate`
- `rsplitn(2, "?").last().expect("always at least one element")`

epage:
So what to do about the CI issue.

Personally, I'd recommend stripping off the last of your commits and rebasing against upstream master.  At that point it should be resolved if you use the same version of rust fmt (0.8.3 I believe).  This will get the CI passing and will keep history clean.  I know "rebase + push -f" is generally considered bad form but no else should really be broken because of it and not sure what else would be recommended to do in this case.

If you don't feel up to that, we could probably commit as is (once the issue above gets cleaned up) or I could clone your branch and clean it up if you wish.

src/main.rs
whostolemyhat: Cool, I'll have a look into this

whostolemyhat:
I've removed the `split().collect()` and replaced it with `rfind` - if there's no match then the original `req_uri` is used.

epage:
> I've removed the split().collect() and replaced it with rfind - if there's no match then the original req_uri is used.

Looks good.  Once the clippy issue is resolved, this is good to go in.

PR #163, opened 2016-11-20
Add next and previous posts to liquid context

nott:
Fixes #139

The idea how to use this feature in Jekyll has been described in [this blog post by David Elbe](http://david.elbe.me/jekyll/2015/06/20/how-to-link-to-next-and-previous-post-with-jekyll.html).

Same in cobalt:
```
<div>
  {% if previous.path %}
    <a class="prev" href="{{previous.path}}">&laquo; {{previous.title}}</a>
  {% endif %}
  {% if next.path %}
    <a class="next" href="{{next.path}}">{{next.title}} &raquo;</a>
  {% endif %}
</div>
```

Issue #139, opened 2016-09-29
Get next and previous posts


ebkalderon:
This is an important feature to have for any blog. See [this blog post by David Elbe](http://david.elbe.me/jekyll/2015/06/20/how-to-link-to-next-and-previous-post-with-jekyll.html) for its usage in Jekyll.

johannhof:
Yup, would be nice.

nott:
i'm woking on this

ebkalderon:
@nott Any progress with this lately?

epage:
Fixed by #218

PR #218, opened 2017-05-10
Add previous and next attributes to posts

whostolemyhat:
Fix for #139. Based on [this PR](https://github.com/cobalt-org/cobalt.rs/pull/163) but with just the changes to the post attributes.

### DIFF
--- a/src/cobalt.rs
+++ b/src/cobalt.rs
@@ -131,8 +131,18 @@ pub fn build(config: &Config) -> Result<()> {
         .collect();
 
     trace!("Generating posts");
-    for mut post in &mut posts {
+    for (i, mut post) in &mut posts.iter_mut().enumerate() {
         trace!("Generating {}", post.path);
+
+        // posts are in reverse date order, so previous post is the next in the list (+1)
+        if let Some(previous) = simple_posts_data.get(i + 1) {
+            post.attributes.insert("previous".to_owned(), previous.clone());
+        }
+        if i >= 1 {
+            if let Some(next) = simple_posts_data.get(i - 1) {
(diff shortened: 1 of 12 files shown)


src/cobalt.rs
```diff
@@ -131,9 +131,19 @@ pub fn build(config: &Config) -> Result<()> {
         .collect();
 
     trace!("Generating posts");
-    for mut post in &mut posts {
+    for (i, mut post) in &mut posts.iter_mut().enumerate() {
         trace!("Generating {}", post.path);
 
+        // posts are in reverse date order, so previous post is the next in the list (+1)
+        if let Some(previous) = simple_posts_data.get(i + 1) {
```
epage: Drat, I was hoping [Iter::window](https://doc.rust-lang.org/std/primitive.slice.html#method.windows) would solve this problem for us. Ideally, we'd use an iterator for this but this is fine for now.

src/cobalt.rs
```diff
@@ -131,9 +131,19 @@ pub fn build(config: &Config) -> Result<()> {
...
+            post.attributes.insert("previous".to_owned(), previous.clone());
+        }
+        if i >= 1 {
+            if let Some(next) = simple_posts_data.get(i - 1) {
```
epage: I had memory and processing performance concerns but in thinking about them, they are a non-issue.

I'd like to make cobalt multi-threaded in the future.  If we view this all as stages in a pipeline, we can make each stage threaded but we won't make things threaded across stages.

I also recoil a little at the amount of memory allocated.  I considered us passing along only specific parts of `document.attributes` but we can't predict too well what the user might want.

So should we make this optional? The problem is most people probably want this.

So in the end, oh well.

Issue #160, opened 2016-11-19
Allow users to specify posts sort order


johannhof:
Something like a config flag "sort_reverse"

bddenhartog:
I'd suggest something like the following, simply for added clarity:

```
[posts]
  sort_order = "{asc/desc}"
```

epage:
I've been toying with both ideas without much personal preference as I've been working on my collections refactor.

epage:
Fixed by #220

PR #220, opened 2017-05-10
Allow posts to be reverse sorted

whostolemyhat:
Fix for #160 - post order is controlled by setting `post_order` to `asc` or `desc` in `.cobalt.yml` (default desc).

### DIFF
--- a/src/cobalt.rs
+++ b/src/cobalt.rs
@@ -125,6 +125,10 @@ pub fn build(config: &Config) -> Result<()> {
     // fall back to the default date
     posts.sort_by(|a, b| b.date.unwrap_or(default_date).cmp(&a.date.unwrap_or(default_date)));
 
+    if &config.post_order == "asc" {
+        posts.reverse();
+    }
+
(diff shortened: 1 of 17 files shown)


src/cobalt.rs
```diff
@@ -125,6 +125,10 @@ pub fn build(config: &Config) -> Result<()> {
     // fall back to the default date
     posts.sort_by(|a, b| b.date.unwrap_or(default_date).cmp(&a.date.unwrap_or(default_date)));
 
+    if &config.post_order == "asc" {
```
epage: Random notes for the future
- This is a `post_order` in the root  of the yaml file.  Eventually I'll get back to cobalt (from liquid) and will generalize posts so it will instead be something like "collections.posts.order = asc" (in pseudo-config)
- This should switch to being an enum and we'll want to switch to serde *but* by switching to serde, the case of the fields might change due to Rust's style guide.  Not fully sure what to do about it.
  - We can hold off on input validation until that point

Issue #197, opened 2017-03-02
Automatic date (published) and modified date (updated)


krruzic:
I know this is possible and I'm trying to see where it could fit in. But is this something that would even be wanted as a feature? Adding a custom tag as described in liquid rust for these things I think would be awesome. 

I'm going to try implementing this on my own but wanted an idea of if this is planned anyway or explicitly not wanted.

Thanks!

uwearzt:
Hi,

I would really like such a feature. I would guess two possible source to get the date off: 

- Directory name YYYYMMDD
- Multiple Directories YYYY/MM/DD

see my posts_in_subfolders test:

https://github.com/cobalt-org/cobalt.rs/tree/master/tests/fixtures/posts_in_subfolder/posts

Maybe getting it from a filename like YYYYMMDD post.md would be a third possible solution.

Greetings

Uwe

krruzic:
Thanks, I'll try to get a dirty implementation of this this weekend. There's [filetime](https://github.com/alexcrichton/filetime) that would allow us to get it from the file properties in a cross platform way. Problem with that is passing a bunch of file data around instead of just reading the contents.

epage:
For Jekyll compatibility, we'd want to support YYYY-MM-DD-name.extension

More specifically, their filename regexes are
'''ruby
    DATELESS_FILENAME_MATCHER = %r!^(?:.+/)*(.*)(\.[^.]+)$!
    DATE_FILENAME_MATCHER = %r!^(?:.+/)*(\d{4}-\d{2}-\d{2})-(.*)(\.[^.]+)$!
'''

epage:
Would something like filetime work?  I didn't think git preserved mtime so it'd work on your local machine but if you rely on something like travis to generate your site, then the dates will all be off.

krruzic:
I didn't know that regarding git. However, at least having automatic creation time with filetime would be cool. Plus, Creation time is already one of the properties so it would be easier to add anyway.

epage:
I've not seen any pages confirming, but I doubt git also preserves ctime and atime as well.

To be more automatic than extracting from the file name, I think you'll need to dig into the git history with all that entails (gracefully handling non-git code with some config flags to control behavior at minimum; SCM abstraction ideally).

krruzic:
True. Starting a `git log` with args and then parsing that per file could be good. Or maybe just don't support travis builds ;)

bddenhartog:
[...]

PR #223, opened 2017-05-19
New post command

whostolemyhat:
First pass at #208 - this adds a `cobalt new` command which creates a new post. (The old `cobalt new` command to create a new project [has been renamed to `cobalt init`](https://github.com/cobalt-org/cobalt.rs/pull/222))

- `cobalt new` creates a post under `posts/new_post.md` (or folder specified in `.cobalt.yml`)
- `cobalt new post hello.md` creates `posts/hello.md`
- `cobalt new layout layout.liquid` creates `_layouts/layout.liquid`
- `cobalt new page about.md` creates `about.md`

All templates are the ones currently used in `new.rs`, and folder names are read from `.cobalt.yml`.

There's a few areas which could be improved:
- Doesn't check for default extensions
- Doesn't check if file already exists (will just overwrite)
- Fails if folder doesn't already exist (eg no `/posts` folder)
- No way to choose templates, and timestamp in posts aren't updated


epage:
@Svenito, looks like @whostolemyhat has an implementation up for review already.  Do you mind checking it out to see where there might have been different UXs or implementations?  Let's see if we can get the best of both worlds.

src/main.rs
```diff
@@ -258,6 +268,56 @@ fn main() {
...
 
+        "new" => {
+            let filetype = matches.value_of("FILETYPE").unwrap();
+            let mut filename = matches.value_of("FILENAME").unwrap();
+
+            match filetype {
+                "layout" => {
+                    if filename == "new_post.md" {
+                        filename = "new_layout.liquid";
+                    }
```
epage: Looks like clap has conditional defaults that we should take advantage of
https://docs.rs/clap/2.24.2/clap/struct.Arg.html#method.default_value_if

```rust
.arg(Arg::with_name("FILENAME")
    .help("Filter to create")
    .default_value_if("FILETYPE", Some("page"), "new_page.md"))
    .default_value_if("FILETYPE", Some("post"), "new_post.md"))
    .default_value_if("FILETYPE", Some("layout"), "new_layout.md"))
    .default_value("new.md"))
```

src/new.rs
```diff
@@ -88,6 +89,30 @@ pub fn create_new_project<P: AsRef<Path>>(dest: P) -> Result<()> {
...
+    try!(create_file(full_path, post_liquid));
+
+    Ok(())
+}
+
+pub fn create_new_page(name: &String) -> Result<()> {
+    try!(create_file(name, index_liquid));
+
+    Ok(())
+}
```
epage: I wonder about the value of having just a single function.

```rust
pub fn create_new_document(type: &str, name: &str, config: &Config)) -> Result<()> {
   // ...
}
```

epage:
Is there enough of a use case for a `cobalt new layout <NAME>`?

epage:
> Doesn't check for default extensions

What do you mean by this?

> Doesn't check if file already exists (will just overwrite)

We should probably protect against this

> Fails if folder doesn't already exist (eg no /posts folder)

Does this have a good error?

> No way to choose templates, 

The feature can evolve

> timestamp in posts aren't updated

I'm thinking its better to have that handled by a `publish` command.

Svenito:
Late to the party here. But agree with above points. Command shouldn't overwrite existing files. 
The layout should have a default value (possibly via config) that maybe can be overridden on the command. 
Is a new layout command needed? Won't they just be created from scratch?

Also agree that last edited should be done via the filesystem modified timestamp at build time.

I'll check more when I get time. Next week is looking a lot clearer for me

epage:
> The layout should have a default value (possibly via config) that maybe can be overridden on the command.

See my comment on the [208](https://github.com/cobalt-org/cobalt.rs/issues/208#issuecomment-302810197)

> Is a new layout command needed? Won't they just be created from scratch?

I'm wondering the same thing.  In addition, I think it makes sense that this is a tool for creating documents.  Seems odd to also include layouts.

> Also agree that last edited should be done via the filesystem modified timestamp at build time.

As I mention in #197, I doubt this is what people really want for posts. That is why I suggest a `publish` command for setting it along with marking it as not a draft.

src/new.rs
Svenito: Also would be worth stubbing out the new file. I ended up adding something like this

```
const new_post_tmpl: &'static [u8] = b"extends: default.liquid

title: %title%
date: %date%
---
";

const new_page_tmpl: &'static [u8] = b"extends: default.liquid

title: %title%
route: %title%
---
";

pub fn create_post<P: AsRef<Path>>(dest: P, title: &str) -> Result<()> {
  let dest = dest.as_ref();
  // 7 may 2017 21:03:34 +0000
  let local = Local::now().format("%d %b %Y %H:%M:%S %z").to_string();
  let vector: Vec<u8> = Vec::from(new_post_tmpl);
  let mut out = String::from_utf8(vector).unwrap();
  out = out.replace("%title%", title);
  out = out.replace("%date%", &local);
  let new_post = &dest.join(format!("posts/{}.md", title));
  try!(create_file(new_post, out.as_bytes()));
  Ok(())
}
```

which basically adds the title and time stamp to the new post or page.

whostolemyhat:
Layout command: yeah, I'm not sure how much use this would be - I was basing this on [this comment](https://github.com/cobalt-org/cobalt.rs/issues/208#issuecomment-292647490) and used layouts instead of templates.

By default extensions, I was referring to the template extensions set in `config.rs`, defaulting to `md` and `liquid`. I think the only thing this would affect is the new command with no filename, which at the mo creates a file called `new_post.md` (and so ignores the template extension config).

epage:
> By default extensions, I was referring to the template extensions set in config.rs, defaulting to md and liquid. I think the only thing this would affect is the new command with no filename, which at the mo creates a file called new_post.md (and so ignores the template extension config).

Ok, so it doesn't validate that the file specified confirms to one of the liquid filenames, so someone could accidentally create a static file instead.  It also assumes the extension `md` even if someone overrode the extensions to not include that.

whostolemyhat:
I've updated the error handling and refactored the create command to handle posts and pages, and removed the new layouts function.

src/new.rs
```diff
@@ -95,21 +95,22 @@ pub fn create_new_document(doc_type: &str, name: &str, config: &Config) -> Resul
...
-        //         error!("here");
-        //     }
+            match Path::new(full_path).exists() {
+                false => {
+                    info!("Created new {} {}", doc_type, name);
+                    try!(create_file(full_path, post_1_md))
+                },
+                true => error!("{} already exists", name)
+            };
         },
```
epage: Seems like you could consolidate the error check to be before your `doc_type` `match`.  You could also remove the nested match if you report it as an error rather than "silently" fail (just reporting to logging).

Something like
```rust
if Path::new(name).exists() {
    bail!("{} already exists", full_path);
}

let content = match doc_type {
    "page" => index_liquid,
    "post" => post_1_md,
}

create_file(full_path, content)?;
```

(note: the `page` branch calls `create_file` with `name` and not `full_path`)

whostolemyhat: I've just been pointed to [create_new](https://doc.rust-lang.org/std/fs/struct.OpenOptions.html#method.create_new) which allows us to push the error back up to `main.rs`, which simplifies things a lot

src/main.rs
```diff
@@ -270,11 +270,12 @@ fn main() {
         }
 
         "new" => {
-            let filetype = matches.value_of("FILETYPE").unwrap();
-            let filename = matches.value_of("FILENAME").unwrap();
+            // TODO: 'cobalt new nonsense' + unwrap = crash so provide default again
+            let filetype = matches.value_of("FILETYPE").unwrap_or("post");
+            let filename = matches.value_of("FILENAME").unwrap_or("new_post.md");
```
epage: Why `unwrap_or`? I thought clap provided the default.

I could see at least changing it to `.expect("default value passed to clap")`

src/new.rs
```diff
@@ -133,10 +117,10 @@ fn create_file<P: AsRef<Path>>(name: P, content: &[u8]) -> Result<()> {
 
     let mut file = try!(OpenOptions::new()
         .write(true)
-        .create(true)
+        .create_new(true)
```
epage: I take it this will report an error?  Whats the error look like? Is it good enough for our users?

whostolemyhat: Yep, this reports an error if the file already exists, otherwise works fine. Apparently this is the preferred way to check if a file exists since it prevents race conditions.

I think the error is clear enough - here's the output:

```
$ cobalt new

[info]   Using config file .cobalt.yml
[error]  File exists (os error 17)
[error]  Could not create post
```

src/new.rs
```diff
@@ -94,24 +94,8 @@ pub fn create_new_document(doc_type: &str, name: &str, config: &Config) -> Resul
...
-                false => {
-                    info!("Created new {} {}", doc_type, name);
-                    try!(create_file(full_path, post_1_md))
-                },
-                true => error!("{} already exists", name)
-            };
-        },
+        "page" => create_file(name, index_liquid)?,
+        "post" => create_file(full_path, post_1_md)?,
         _ => println!("Can only create post or page")
```
epage: should this be a `bail!("Unsupported document type: {}", doc_type);`?

src/main.rs
epage: Oh, because you only have conditional defaults.  Couldn't you mix conditional with unconditional defaults?

src/main.rs
```diff
@@ -270,9 +270,8 @@ fn main() {
         }
 
         "new" => {
-            // TODO: 'cobalt new nonsense' + unwrap = crash so provide default again
-            let filetype = matches.value_of("FILETYPE").unwrap_or("post");
-            let filename = matches.value_of("FILENAME").unwrap_or("new_post.md");
+            let filetype = matches.value_of("FILETYPE").unwrap();
+            let filename = matches.value_of("FILENAME").unwrap();
```
epage: As I mentioned before, you could make these "expects".  That way we document in the message why we think its safe to unwrap.

whostolemyhat:
Hiya, do you need me to make any more changes to this PR?

epage:
Sorry for not making the steps forward clear before I went on vacation.

Overall, it looks good.  Would you be able to resolve the conflicts so we can see what the CI says?

epage:
Drat.  I need to change our CI to not `cargo install rustfmt` because of how long the compile can sometimes take (or find a way to cause it to output as it goes).

Issue #221, opened 2017-05-12
Propose not skipping hidden folders/files


Svenito:
The walkdir dir filter skips files prefixed with `_` and `.`. However hidden folders are required for example to prove keybase identity and setup Let's Encrypt verifications in a `.well-known` folder. 

Would it make sense to not skip hidden files and folders?

Although wrangling files like `.cobalt.yml`, `.gitlab-ci.yml`, etc would require them to be added to the ignore list in the config. This could be done by providing a default set of the most common config and directories.

Alternatively add an option to force include certain folders and files, somewhat like Pelican does

Svenito:
Ref #211

epage:
Let me see if I understand the use case.  There are source files (assets of templates) that are hidden or in hidden folders because in the output they are expected to be there and hidden?  You gave the example of `.well-known`.  `.htaccess` is an example from Jeyll's documentation.

As you pointed out, the challenge with this is there are a lot of dot files that someone would expect to not be copied over.  `.git` is another one. 

So we have two choices
- Copy dot files, requiring a user to blacklist files
- Don't copy dot files unless they are in a whitelist

I feel the latter would follow the principle of least surprise to users and is the direction I'm leaning.  Also, there is precedence in at least [Jekyll](https://jekyllrb.com/docs/configuration/#global-configuration).  I've actually started work on this in my dev branch but have been stalled on it because I've been spending time on polishing Liquid first.

Svenito:
I agree with you that the 

> Don't copy dot files unless they are in a whitelist

is the preferred solution to this. But yes, you are right, there is a need to have certain hidden files and folders in your site structure for various reasons, `.well-known` being one example.

I am happy to carry on with this from your dev branch if you like. Currently I am looking at addressing #208 as it's something I also want

epage:
My dev branch is a lot bigger change than just this.  If you'd like this sooner than later, feel free to add this feature and send a PR.

I was actually thinking of suggesting #208 to you.  I have a couple of ideas for it I'll post over there.

Svenito:
No problem. I'll make a start soon. I've started in the new post issue, but not with subcommand plugins. I'll probably put that down the list and push this up then.

PR #226, opened 2017-05-22
Add include option to config

Svenito:
Override usually ignored folders/files to be included. Such as those prefixed with `.` or `_`

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -22,6 +22,7 @@ pub struct Config {
     pub description: Option<String>,
     pub link: Option<String>,
     pub ignore: Vec<Pattern>,
+    pub include: Vec<Pattern>,
     pub excerpt_separator: String,
 }
 
@@ -42,6 +43,7 @@ impl Default for Config {
             description: None,
             link: None,
             ignore: vec![],
+            include: vec![],
             excerpt_separator: "\n\n".to_owned(),
         }
     }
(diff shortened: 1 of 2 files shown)


src/config.rs
```diff
@@ -22,6 +22,7 @@ pub struct Config {
     pub description: Option<String>,
     pub link: Option<String>,
     pub ignore: Vec<Pattern>,
+    pub include: Vec<Pattern>,
```
epage: Please give more detail on this change in your commit message.  They are an important place to communicate.

For example, some useful things to include
- Link to #221 to help give context since sometimes designs are worked out in issues
- Especially if the user-facing design isn't fleshed out in the issue, describe it (the What?).  For example "To allow overriding cobalt's default ignore behavior, this adds a new `include` field to cobalt.yml that ...".
- On related note, be sure to include how this feature interacts with other features.  How do `ignore` and `include` interact? Should it be defined or undefined behavior?  What should it be defined to?  Why?
- "Why?" on your implementation, if any is needed, is also important.  This is a simple enough change, I don't expect much of a "Why?" regarding the implementation.

Notice that I separate out user-facing design / workflow separate from implementation.  "How?" with implementation isn't as important; the code covers that.

Svenito: Fleshed out the commit message.

epage:
So while I said I was focusing on liquid and wasn't working on this, that changed over the weekend and I found my very first step towards a generalized collection/section feature was refactoring how we find files.  In working on it, I had an idea that would solve several implementation problems on my side and address this issue.

While my change is still in progress, I'm curious what you think of my idea on how to solve this

See https://github.com/epage/cobalt.rs/commit/f307346ca110228a7535163fbfe1b0a01bc08b12

epage:
Note: I planned to force push to the branch :)

Svenito:
I'd be for this. Having a single file where you can black and whitelist items is ideal, and gitignore is a known format. +1 from me

epage:
FYI I've written the iterator and have some basic tests in place.  Now it just needs to be integrated into the rest of cobalt.  I realized I should probably stop rewriting commits until I have this ready to submit as a PR in case you want to add anything to it.
https://github.com/epage/cobalt.rs/commit/9918d1ec045259974a33735a67d34841016de1a2

epage:
Closing since we agreed to go with #227

Issue #81, opened 2016-04-08
Incremental build with Watch command


LucioFranco:
Right now every time you change a file it rebuilds the whole site. We may want to do a few things:

- [ ] Refactor `cobalt.rs` so that the builds can be done incrementally.
- [ ] Add delay for watcher so that we do not need to do as many builds

If there is anything else let me know! :smile:

epage:
>  Add delay for watcher so that we do not need to do as many builds

`notify` has built in support for this now.  We'd just have to switch from the raw to debounced API which are a bit different.

epage:
At minimum, we rebuild the file that was directly modified.  This doesn't help us if a layout file, config file, index file, etc change.  My crazy, far out there, idea is to build up the dependency graph on the fly and use something like [depraph](https://docs.rs/depgraph/0.1.0/depgraph/) (maybe with some added features) to take care of the fancy stuff.

skierpage:
Could `cobalt serve` only generate files in its temporary output directory when someone requests them? (I realize this doesn't avoid the dependency graph problem.) I have a large-ish site with a lot of static assets and it would be neat if the temporary directory was only populated as needed.

epage:
So instead of building based off of file watches, build when a web request comes in?  That could make that initial web request quite slow.  Its also not quite clear how much that would save.  Do you do a lot of file saves before looking at the pages?

PR #227, opened 2017-05-24
feat: Increase control over files to process 

### LINKED ISSUE #221: Propose not skipping hidden folders/files

epage:
The `ignore` field in `.cobalt.yml` is changing from a list of glob patterns to
a list `gitignore` patterns, allowing whitelisting and more advanced
whitelisting/blacklisting with well defined semantics.

Previuously, Cobalt processed all files except `.dot` and `_hidden`,
*dest* files/directories.

Considerations when designing this:
- User workflow: users need `.dot` asset files, like `.htaccess`, copied over.
- Implementation: Finding files for pages and posts is coupled together
  which will be a problem as we refactor them into more general
  collections or sections.
- Performance: Eventually we'll want to implement incremental rebuild
  (see issue #81) and to do so we'll need to know what collection a file
  notification is relevant to.

`gitignore` patterns were selected for the first item to ensure we had
well defined semantics for how blacklist and whitelist features would
interact, directory vs file patterns, etc.

The last two are the motivation for factoring out file detection.

A couple downsides, for now
- `cobalt watch` will now rebuild when hidden files are touched.  This is
  unfortunate but seems like an acceptable short term sacrifice until we
  have a better to expose what files cobal reads.
- Handling of overlapping source/dest is a bit ugly. In the future, we
  should just make these cases error
- Handling of hidden post folders is ugly.  In the future, we should
  generalize the concept of pages and posts into collections/sections at
  which point we'll walk each collection type independently.  This will
  make the non-hidden posts folders the ugly case.  We should consider
  requiring them to be hidden.
- There is a known bug where we enable all hidden folders that match the
  posts folders name rather than only some.

Fixes #221 by allowing the user to whitelist `.dot` and `_hidden` files.
See the added unit tests for examples.

BREAKING CHANGE: The format of `.cobalt.yml`'s `ignore` field has changed
from a list of glob patterns to gitignore patterns.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,8 +10,8 @@ dependencies = [
  "env_logger 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "error-chain 0.10.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "ghp 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "glob 0.2.11 (registry+https://github.com/rust-lang/crates.io-index)",
  "hyper 0.10.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "ignore 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "itertools 0.5.9 (registry+https://github.com/rust-lang/crates.io-index)",
  "lazy_static 0.2.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "liquid 0.9.1 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -212,6 +212,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "crossbeam"
+version = "0.2.10"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "dbghelp-sys"
 version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -320,9 +325,16 @@ dependencies = [
 ]
 
 [[package]]
-name = "glob"
-version = "0.2.11"
+name = "globset"
+version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "aho-corasick 0.6.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "fnv 1.0.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "log 0.3.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "memchr 1.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "regex 0.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
+]
 
 [[package]]
 name = "httparse"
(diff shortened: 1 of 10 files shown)


epage:
@Svenito, this should now be done.  Could you take a look to let me know your thoughts?

Svenito:
I can't see an example of the new syntax. It also seems that a new blog will still have the same ignore syntax? Or is this now handled only in `FilesBuilder` making entries in `new.rs` obsolete?

epage:
The problem with having documentation in a separate repo :)

We still read `ignore` from the config file.  Each string in the `ignore` list is [added](https://github.com/cobalt-org/cobalt.rs/pull/227/files#diff-1adad72e4e1dbe7019b775a6acb9455aR62).

The following two tests cover both blacklisting files and whitelisting files.  You can take the strings from their `add_ignore` call and just plop it right into your `.cobalt.yml` file.
https://github.com/cobalt-org/cobalt.rs/pull/227/files#diff-acdaa19225c80ac237c32d3ebb0e1214R256

Svenito:
Well, the ignore is working, but the negation doesn't seem to:

```
ignore:                                                                         
  - !**/.well-known/*
  - .git/*                                                                      
  - .gitignore/*                                                                
  - build/*                                                                     
  - output/*                                                                    
  - .cobalt.yml                                                                 
  - .gitlab-ci.yml                                                              
```

epage:
I suspect there is some behavior differences in `Files` API

- `includes_file` / `includes_dir` check the file name as-is
- `FilesIterator` effectively checks each level of the file heirarchy against the ignores.

I need to change `includes_file` to check each parent directory (starting at the root most) against the ignore pattern.

epage:
> Well, the ignore is working, but the negation doesn't seem to:

Could you post the output with running with debug logging enabled?

Svenito:
Long list, so cut to the last bits.

```[debug]  "./.cobalt.yml": ignored ".cobalt.yml"
[debug]  "./output/images": ignored "output/*"
[debug]  "./output/2014": ignored "output/*"
[debug]  "./output/2017": ignored "output/*"
[debug]  "./output/2015": ignored "output/*"
[debug]  "./output/2013": ignored "output/*"
[debug]  "./output/2012": ignored "output/*"
[debug]  "./output/2009": ignored "output/*"
[debug]  "./output/2011": ignored "output/*"
[debug]  "./output/archive": ignored "output/*"
[debug]  "./output/rss.xml": ignored "output/*"
[debug]  "./output/2016": ignored "output/*"
[debug]  "./output/assets": ignored "output/*"
[debug]  "./output/index.html": ignored "output/*"
[debug]  "./.git": ignored "**/.*"
[debug]  Created new directory "output/assets"
[debug]  Copied "./assets/main.css" to "output/assets/main.css"
[debug]  "./.well-known": ignored "**/.*"
[debug]  "./.gitignore": ignored "**/.*"
[info]   Build successful
```

Could this be a matter of precendence? It would seem from the gitignore docs that the whitelist works *after* the file has been black listed. https://git-scm.com/docs/gitignore

epage:
btw the log from when the ignore lines are added is also useful but might not be needed in this case

> Could this be a matter of precendence? It would seem from the gitignore docs that the whitelist works after the file has been black listed. https://git-scm.com/docs/gitignore

"**/.*" is added before any of your ignores are added, so its not a precedence issue.

My suspicion is that the `**` in your `!**/.well-known/*` is requiring there to be directories below your well knowsn directory but instead your well knowns are in the root (`"./.well-known": ignored "**/.*"`).

I should make a pattern testing tool to help people in debugging this stuff.  I iterated *a lot* with using tests to make sure I got my logic right.

Svenito:
Ok, so that's not it. Tried `!.well_known/*` and `!.well_known`, neither include it

Svenito:
The ignores added in the code don't end up in the config file and might confuse people why their files aren't being included. Might be worth adding those patterns to the default config line?

epage:
Previous to my change, the ignores weren't in the config file.  The only difference is I'm now using the same mechanism for all ignores rather than two different mechanisms.

Its reasonable to question if they should have been in the ignore file from the beginning.  I still think not though because these ignores are part of the invariants and not just a handy default for the user.  One option is to log them to draw attention to them.

Svenito:
Seems fair enough. I'm still having issues with whitelisting however

```
name: cobalt blog
source: "."
dest: "build"
ignore:
  - .git/*
  - build/*
  - !.test
```

where `.test` is a file. Is my syntax and understanding of the config's ignore format correct?

PR #228, opened 2017-05-24
fix(CI): Protect against timeout on Travis

epage:
Building rustfmt was slow enough that it sometimes timed out our CI, so
switching back to downloading it but leaving it upgraded.


epage:
Appveyor seems to be a random failure.

Travis failed because
- installing clippy failed
- format differences between 0.8.3 and 0.8.3-20170509

epage:
Downloading rustfmt has its own problems

Using Travis' `travis_wait` doesn't seem to work with Cobalt and yet it works with Liquid.

*sigh*

PR #230, opened 2017-05-27
Continuation of #227

epage:
- Improved the behavior of the immediate calls compared to the walking behavior
- Added tests for whitelisting

### DIFF
--- a/src/files.rs
+++ b/src/files.rs
@@ -14,10 +14,8 @@ pub struct FilesBuilder {
 impl FilesBuilder {
     pub fn new(root_dir: &Path) -> Result<FilesBuilder> {
         let mut ignore = GitignoreBuilder::new(root_dir);
-        ignore.add_line(None, "**/.*")?;
-        ignore.add_line(None, "**/.*/**")?;
-        ignore.add_line(None, "**/_*")?;
-        ignore.add_line(None, "**/_*/**")?;
+        ignore.add_line(None, ".*")?;
+        ignore.add_line(None, "_*")?;
         let builder = FilesBuilder {
             root_dir: root_dir.to_path_buf(),
             ignore: ignore,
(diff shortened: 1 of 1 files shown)


epage:
Huh, I forgot I submitted #227.

@Svenito this includes tests for how to whitelist.  It seems odd that directories need extra work.  

So I guess the question is
- Is this just how gitignore works?
- Am I misusing burntsushi's crate?
- Is there a bug in burntsushi's crate?

epage:
Also, I'll be on vacation tomorrow.  I'll have limited email but not access to my development machine.

Svenito:
Thanks. I have it working now. The issue was that you have to quite the patterns because the `!` returns a blank line in the yaml parser.

epage:
Good to know. We'll want to document that somewhere as a common pitfall.

epage:
Printing the calls to `includes_path_leaf`

Windows:
```
Testing "/" true
Testing "/usr" false
Testing "/usr/cobalt" false
Testing "/usr/cobalt/site" false
Testing "/usr/cobalt/site/child" false
Testing "/usr/cobalt/site/child/child.txt" false
Testing "" false
Testing "." true
Testing "./child" false
Testing "./child/child.txt" false
test files::tests::files_includes_child_dir_file ... ok
```

Linux
```
---- files::tests::files_includes_child_dir_file stdout ----
        Testing "/" true
Testing "/usr" true
Testing "/usr/cobalt" false
Testing "/usr/cobalt/site" false
Testing "/usr/cobalt/site/child" false
Testing "/usr/cobalt/site/child/child.txt" false
Testing "" false
Testing "." true
thread 'files::tests::files_includes_child_dir_file' panicked at 'assertion failed: `(left == right)` (left: `false`, right: `true`)', src/files.rs:235
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

epage:
I've made a change to prevent us from recursing up past `root_dir`.  So it seems that my code is acting the same between Windows/Linux and that Gitignore is acting different.  Preparing a workaround.

PR #231, opened 2017-06-06
Create new post page with title and date

Svenito:
This expands on the existing `new post` command and creates a post from a template with a title set by the name of the post and its date set to the current date and time.

It also adds the `.md` extension to the filename
    
Referencing #208 and #223

### DIFF
--- a/src/new.rs
+++ b/src/new.rs
@@ -65,11 +65,27 @@ const index_liquid: &'static [u8] = b"extends: default.liquid
 </div>
 ";
 
+const new_post_tmpl: &'static [u8] = b"extends: default.liquid
+
+title: %title%
+date: %date%
+---
+";
+
+
+//const new_page_tmpl: &'static [u8] = b"extends: default.liquid
+//title: %title%
+//route: %title%
+//  ---
+//";
+
+
 use std::path::Path;
 use std::fs::{DirBuilder, OpenOptions};
 use std::io::Write;
 use error::Result;
 use config::Config;
+use chrono::Local;
 
 pub fn create_new_project<P: AsRef<Path>>(dest: P) -> Result<()> {
     let dest = dest.as_ref();
@@ -89,13 +105,24 @@ pub fn create_new_project<P: AsRef<Path>>(dest: P) -> Result<()> {
     Ok(())
 }
 
-pub fn create_new_document(doc_type: &str, name: &str, config: &Config) -> Result<()> {
+pub fn create_post(name: &str, config: &Config) -> Result<()> {
     let path = Path::new(&config.source);
-    let full_path = &path.join(&config.posts).join(name);
+
+    let local = Local::now().format("%d %b %Y %H:%M:%S %z").to_string();
+    let vector: Vec<u8> = Vec::from(new_post_tmpl);
+    let mut out = String::from_utf8(vector).unwrap();
+    out = out.replace("%title%", name);
+    out = out.replace("%date%", &local);
+    let new_post = &path.join(format!("{}/{}.md", &config.posts, name));
+    try!(create_file(new_post, out.as_bytes()));
+    Ok(())
+}
+
+pub fn create_new_document(doc_type: &str, name: &str, config: &Config) -> Result<()> {
 
     match doc_type {
         "page" => create_file(name, index_liquid)?,
-        "post" => create_file(full_path, post_1_md)?,
+        "post" => create_post(name, config)?,
         _ => bail!("Unsupported document type {}", doc_type),
     }
 
--- a/src/new.rs
+++ b/src/new.rs
@@ -73,13 +73,6 @@ date: %date%
 ";
 
 
-//const new_page_tmpl: &'static [u8] = b"extends: default.liquid
-//title: %title%
-//route: %title%
-//  ---
-//";
-
-
 use std::path::Path;
 use std::fs::{DirBuilder, OpenOptions};
 use std::io::Write;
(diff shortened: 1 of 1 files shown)


src/new.rs
```diff
@@ -89,13 +98,24 @@ pub fn create_new_project<P: AsRef<Path>>(dest: P) -> Result<()> {
...
-pub fn create_new_document(doc_type: &str, name: &str, config: &Config) -> Result<()> {
+pub fn create_post(name: &str, config: &Config) -> Result<()> {
     let path = Path::new(&config.source);
-    let full_path = &path.join(&config.posts).join(name);
+
+    let local = Local::now().format("%d %b %Y %H:%M:%S %z").to_string();
+    let vector: Vec<u8> = Vec::from(new_post_tmpl);
+    let mut out = String::from_utf8(vector).unwrap();
+    out = out.replace("%title%", name);
+    out = out.replace("%date%", &local);
```
epage: Why not use `format!`?

src/new.rs
```diff
@@ -89,13 +98,24 @@ pub fn create_new_project<P: AsRef<Path>>(dest: P) -> Result<()> {
...
+    try!(create_file(new_post, out.as_bytes()));
+    Ok(())
+}
+
+pub fn create_new_document(doc_type: &str, name: &str, config: &Config) -> Result<()> {
 
     match doc_type {
         "page" => create_file(name, index_liquid)?,
-        "post" => create_file(full_path, post_1_md)?,
+        "post" => create_post(name, config)?,
```
epage: I feel like pages should also get a title set on them

src/new.rs
```diff
@@ -89,13 +98,24 @@ pub fn create_new_project<P: AsRef<Path>>(dest: P) -> Result<()> {
...
+pub fn create_post(name: &str, config: &Config) -> Result<()> {
     let path = Path::new(&config.source);
-    let full_path = &path.join(&config.posts).join(name);
+
+    let local = Local::now().format("%d %b %Y %H:%M:%S %z").to_string();
+    let vector: Vec<u8> = Vec::from(new_post_tmpl);
+    let mut out = String::from_utf8(vector).unwrap();
+    out = out.replace("%title%", name);
+    out = out.replace("%date%", &local);
+    let new_post = &path.join(format!("{}/{}.md", &config.posts, name));
```
epage: Question: are we assuming `name` is a file name or a post name? 

If file name, then we should conditionally add `.md`.  Are we assuming the name is a slug?  Should we slugify it first and then titlize_slug for the `title` field?

If a post name, we should slugify the `name` when creating a path.

See https://github.com/cobalt-org/cobalt.rs/blob/master/src/slug.rs

Note: this question also applies to `create_new_document` and the `cobalt new` defaults.

src/new.rs
```diff
@@ -89,13 +98,24 @@ pub fn create_new_project<P: AsRef<Path>>(dest: P) -> Result<()> {
     Ok(())
 }
 
-pub fn create_new_document(doc_type: &str, name: &str, config: &Config) -> Result<()> {
+pub fn create_post(name: &str, config: &Config) -> Result<()> {
```
epage: This function only uses `config` for `source`.  Unless we expect to use more of config, we should probably have this function take a `path` instead.

epage:
@whostolemyhat any thoughts on this?

whostolemyhat:
It would be nice if the `init`/`create_new_project` command used the new template, then we can remove `post_1_md` from `new.rs`.

epage:
> It would be nice if the init/create_new_project command used the new template, then we can remove post_1_md from new.rs.

The challenge is what would you put in for the title?

epage:
btw @Svenito, it turns out that `clap` has at least some basic support for external commands (hadn't noticed this before)

https://users.rust-lang.org/t/using-external-subcommands-with-clap/10753/4

epage:
btw @Svenito, it turns out that `clap` has at least some basic support for external commands (hadn't noticed this before)

https://users.rust-lang.org/t/using-external-subcommands-with-clap/10753/4

whostolemyhat:
I was imagining something like `create_post("First Post", config)` in `create_new_project`, so the title's just hardcoded in that function.

However, it'd mean that there's no content in the post generated by starting a new project, which would be confusing if a user has never used Cobalt or a static site generator before, so it could be better to keep the `post_1.md` template.

src/new.rs
Svenito: Yes, that makes sense

src/new.rs
Svenito: Figured I'd give more control as to positioning and you don't need to put the name and local in the right order. Also allows you to have items in more than one place. But seeing as control of format and inputs is in our control, format will be a better option

src/new.rs
Svenito: In my mind it was the title of the post, which then also becomes the filename. This is how Pelican does it, but that doesn't mean it's correct necessarily, so whatever you prefer here.

src/new.rs
Svenito: yes, I haven't got round to making the pages yet

Svenito:
> However, it'd mean that there's no content in the post generated by starting a new project, which would be confusing if a user has never used Cobalt or a static site generator before, so it could be better to keep the post_1.md template.

Or we just add some content to the new template by default for all cases?

src/new.rs
epage: fmt seems to support [named arguments](https://doc.rust-lang.org/std/fmt/#named-parameters).  Eventually we want to allow the skeleton files to be user-provided.

Svenito: Aha, I was looking for this documentation, as it seemed like it should support named args, but at the time couldn't find it. Will update PR soon

src/new.rs
epage: I can see either direction.  We just need to decide and document the behavior and why we made the choices we did in the PR.  We can later transfer the behavior documentation over to the user docs.

So if its the title
- Should we change [`FILENAME`](https://github.com/cobalt-org/cobalt.rs/blob/master/src/main.rs#L111) to `TITLE`?
- Should we slugify the name for creating the file name?  I'd lean towards "yes"
- Should the user be expected to quote the parameter or should we capture all following text as part of the name (not sure if multiple occurences or multiple values is the right approach in clap)?

epage:
#325 took care of this I think.

PR #232, opened 2017-06-06
sub spaces in urls

Svenito:
Some URLs might have spaces in them if the filenames have spaces. This patch will replace the spaces with a `-` character.

### DIFF
--- a/src/document.rs
+++ b/src/document.rs
@@ -89,6 +89,9 @@ fn format_path(p: &str,
         }
     }
 
+    p = p.replace(" ", "-");
+    p = p.to_lowercase();
+
     let mut path = Path::new(&p);
 
     // remove the root prefix (leading slash on unix systems)


epage:
Why not allow spaces in them?

Svenito:
It was causing issues for me with spaces. If this isn't an issue for anyone else however, then we can close this

IIRC it didn't parse out the URLs correctly in index links, but I'd have to double check.

epage:
> It was causing issues for me with spaces. If this isn't an issue for anyone else however, then we can close this

> IIRC it didn't parse out the URLs correctly in index links, but I'd have to double check.

Could you post an example site and point out what is going wrong because of spaces?

epage:
@Svenito, if this is still a problem for you, could you create an issue about the problem you are running into with spaces so we can look into a fix?  Modifying the name behind the user's back will probably not be the solution we go with.

Closing this out for now as root cause is determined.

PR #238, opened 2017-06-16
feat(debug): Dump intermediate state

epage:
This is useful for
- Seeing what variables are available
- Isolate bugs to be either in cobalt or liquid


epage:
For some reason, syntect is being weird.

PR #254, opened 2017-07-03
Add timestamp to layouts for cachebusting

whostolemyhat:
This allows layouts to access a millisecond timestamp which can be used for cachebusting (adding a unique string to static assets to get around long caching times in the browser).

## Notes
- I used `Value::Str` since `Value::Num` only accepts `f32` - the timestamp is `i64` so casting to `f32` trims the number to the point where it's not unique.
- There's a bit of a dance in the test since the timestamp is generated when the `build` command is run - this means the target to compare with needs to be built when `cargo test` runs.
- It might be better to add `date` instead of `timestamp` and then format in the layout file - this would give users more flexibility since they can access the Date object, but would mean each asset would have to format the date individually.

### DIFF
--- a/src/cobalt.rs
+++ b/src/cobalt.rs
@@ -176,12 +176,14 @@ pub fn build(config: &Config) -> Result<()> {
         .collect();
 
     trace!("Generating other documents");
+    let timestamp = Value::Str(UTC::now().timestamp().to_string());
     for mut doc in documents {
         trace!("Generating {}", doc.path);
 
         if config.dump.contains(&Dump::Liquid) {
             create_liquid_dump(dest, &doc.path, &doc.content, &doc.attributes)?;
         }
+        doc.attributes.insert("timestamp".to_owned(), timestamp.clone());
 
         let mut context = doc.get_render_context(&posts_data);
         let doc_html = try!(doc.render(&mut context, source, &layouts, &mut layouts_cache));
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1,6 +1,6 @@
 [root]
 name = "cobalt-bin"
-version = "0.6.1"
+version = "0.7.1"
 dependencies = [
  "assert_cli 0.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "chrono 0.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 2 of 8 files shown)


src/cobalt.rs
```diff
@@ -176,12 +176,15 @@ pub fn build(config: &Config) -> Result<()> {
         .collect();
 
     trace!("Generating other documents");
+    let timestamp = Value::Str(UTC::now().timestamp().to_string());
```
epage: > I used Value::Str since Value::Num only accepts f32 - the timestamp is i64 so casting to f32 trims the number to the point where it's not unique.

liquid-rust/#88 will require us to have integer values.  So in the future we can change this.

epage:
Note: overall, I think it might be useful to have some kind of "generated-time" value but I'd expect it to parallel the eventual "published-time" and "last-modified-time" attributes which would all be in date formats.  If it doesn't exist, we could add a liquid filter to turn a datetime into an integer value.

So if we decide a checksum solution is best but you want a solution now, reframing this as "generated-time" timestamp might work out.

whostolemyhat:
You're right, this would end up breaking the cache each time a build is run, so if you're creating a post a day it'd have a pretty detrimental affect on caching (the opposite of what we want!)

I think it makes sense to turn this into the `generated-time` attribute then parse it in the layouts for a quick solution; the better option for cachebusting would be to hash/checksum each asset so only updated ones are affected. This would be more complex to do since we'd need to map each asset and insert the hashed filename in each document but would be better to avoid breaking caching for unchanged assets.

epage:
>  the better option for cachebusting would be to hash/checksum each asset so only updated ones are affected. This would be more complex to do since we'd need to map each asset and insert the hashed filename in each document but would be better to avoid breaking caching for unchanged assets.

cobalt already copies over every asset, so it can easily get a checksum as it goes.  Building up a `liquid::Object` to store it all in isn't too bad.  The main annoyance is updating the pipeline to handle assets before everything else.  Hopefully my planned refactors will help with it.  I'm about done with the second step of my refactor (first was file walking, second is document parsing).  Next up is config file handling.  I think after that is fixing up the pipeline.  This is not a quick process.  See #257 for some of the care I'm taking in tracking everything.

Issue #257, opened 2017-07-04
Make frontmatter future compatible


epage:
Currently, the frontmatter contains
- Cobalt customization
  - With path it also then gets turned into an attribute
- Overriding of cobalt variables
- custom variables

The problem is that we could break custom variables as we add new cobalt variables or customization.

## Proposal

Frontmatter may contain:
- path
  - will be renamed to `permalink` to match Jenkins when we break compatibility on the format (from Jekyll's custom format to liquid)
  - why: for determining dest path
  - post default will change to `/:categories/:year/:month/:day/:slug.html` rather than having the same default as a page to align with Jekyll
- slug
  - why: to contribute to default dest path
- title:
  - why: rss
- description
  - why: rss
- categories
  - why: to contribute to default dest path, RSS, etc
- excerpt_separator
  - why: to determine the excerpt
- **excerpt** should we still support overriding the auto-created excerpt?
- published_date
  - renamed from `date` to make the role more clear when we added `modified_date`
  - why: rss, sorting
- format (markdown, raw, etc):
  - reinvented form of `ext`
  - why: control formatting
- layout
  - renamed from `extends` to match Jekyll
- is_draft
  - renamed from `draft` for consistency
  - why: filtering posts
- ~~is_post~~
  - Removed: Unlikely used; need to actually define semantics if we want to allow customizing this
- custom
  - why: user defined attributes

Page attributes will be
- url: absolute path
  - Renamed from `path` to clarify intent; to be like jekyll
  - See #216 for a disussion on whether it should be relative or absolute
- path: path on disk
  - Renamed from `source` to be like jekyll
- ~~slug~~
  - Removed: it was only added for Path attributes
- title
- description
- categories
- ~~excerpt_separator~~
  - Removed: it was only meant for the frontmatter to customize cobalt
- published_date
  - renamed from `date` to make the role more clear when we added `modified_date`
- ~~ext~~
  - Removed: it was meant for bookkeeping within cobalt
- ~~extends~~
  - Removed: it was only meant for the frontmatter to customize cobalt
- is_draft
  - renamed from `draft` for consistency
- is_post
- custom (eventually)
- content
- excerpt
- next
- previous

Path attributes (will be re-examined in #198)
- path: relative path of the file, without output_ext.  Normally limited to collections but I think we should make it generally available
- slug: slugified `slug` attribute, if available, otherwise file name (missing date prefix, missing extension)
[...]

Issue #211, opened 2017-04-11
Restructure project file tree and build sources


bddenhartog:
Conversation pulled from https://github.com/cobalt-org/cobalt.rs/pull/209#pullrequestreview-31691911.

bddenhartog:
_Comment left by @epage on Apr 8 2017 02:44 AM UTC_

> Reason you are removing the leading "_"? I believe those tell cobalt to ignore these folders so they don't get processed or copied over as assets?

bddenhartog:
_Comment left by @bddenhartog on Apr 10 2017 19:56 PM UTC_

> Ah, I see that you are correct - [`/src/cobalt.rs L23-L25`](https://github.com/cobalt-org/cobalt.rs/blob/master/src/cobalt.rs#L23-L25). I'll add an underscore to `templates`, but -- off-topic -- I wonder if this is really the best way to handle this? Using a pattern for ignoring could cause problems for the user (what if I decide to add a folder called `_images`?); usually things like this are handled via an explicit list or from a configuration value, like the `ignored` property in `.cobalt.yml`. What are your thoughts? This conversation should probably be in its own issue.
> 
> `_drafts` actually shouldn't be changed here, as this PR is strictly for `_layouts -> templates`. I'll remove this change.

bddenhartog:
_Comment left by @epage on Apr 10 2017 21:40 PM UTC_

> I think thats a good idea to split this into another issue.
> 
> > Personally, I think moving all build/cli-related files into a single .cobalt subdirectory allows us to add new tooling and features - like skeleton files - without having to worry about polluting the project space. It makes it clear what the user is responsible for maintaining within their project, and where they can go to modify how cobalt works in regards to their project.
> 
> Just to check, is your comment about a ".cobalt/" meant to be related to avoiding hiding files with a "_" prefix?

bddenhartog:
> Just to check, is your comment about a ".cobalt/" meant to be related to avoiding hiding files with a "_" prefix?

Yes, my previous comment is suggesting using a `.cobalt/` directory in the project root to store things used _for_ building the project but not actually built to html themselves, like `skeleton files` and other things the user may want to modify (`templates`). In general, I think the file tree for cobalt projects could be vastly improved upon for both clarity and sanity. Something like:

```text
my-awesome-site
|- .cobalt/
|  |- skel/
|  |  |- template.liquid
|  |  |- page.liquid
|  |  |- post.liquid
|  |
|  |- templates/
[...]

Issue #200, opened 2017-03-09
Syntax highlight vimscript


NickeZ:
Hi, Really nice project! How would I implement syntax highlighting for vimscript? Could I add a sublime syntax highlight definition file to my project somehow or do I need to modify cobalt or syntect?

epage:
Hey, sorry for not getting back to you on this.

We use `syntect` for syntax highlighting.  We load the existing dump of [syntax files](https://github.com/trishume/syntect/tree/master/assets) but I don't see any documentation for what languages are included or how to add more.

I think it'd be a good idea for us to add a "syntax folder" so people can add syntax files on top of what is already supported.

For someone wanting the details, we'd need to refactor cobalt.rs/src/syntax_highlight.rs so that we can call
```rust
if load_syntax_files {
    syntax_set.load_syntaxes(syntax_path, false);
    syntax_set.link_syntaxes();
}
```

`load_syntax_files` is a placeholder for whatever the logic should be.  We should avoid doing this unless we have to because the default approach comes pre-linked and syntect's docs say " This operation is idempotent, but takes time even on already linked syntax sets."

I'm leaning towards checking if the directory exists.  I'm for minimizing configuration unless we have a good use case so I'd like to avoid a `load_syntax_files` or `syntax_path` in the config file.

epage:
Looks like `[load_from_folder](http://thume.ca/rustdoc/syntect/syntect/highlighting/struct.ThemeSet.html#method.load_from_folder)` is how to load themes but not syntaxes.  For those we'd want `[load_syntaxes](http://thume.ca/rustdoc/syntect/syntect/parsing/struct.SyntaxSet.html#method.load_syntaxes)`.  We'd then need to `link_syntaxes`

boxofrox:
@epage, `load_syntaxes` requires a mutable `SyntaxSet` which would require some modification to `SETUP` in `syntax_highlight.rs` [\[1](https://github.com/cobalt-org/cobalt.rs/blob/master/src/syntax_highlight.rs#L34)].

\[1]: https://github.com/cobalt-org/cobalt.rs/blob/master/src/syntax_highlight.rs#L34

There are a few ways to handle the mutability requirement:

1. `std::cell::RefCell`
    ```rust
    struct Setup {
        syntax_set: RefCell<SyntaxSet>,
        theme_set: ThemeSet,
    }
    lazy_static!{
        static ref SETUP: Setup = Setup {
[...]

PR #259, opened 2017-07-07
Allow configuring the theme for syntax highlighting

### LINKED ISSUE #202: Allow configuring the theme for syntax highlighting
Looking into a way to make the syntax highlighting theme adjustable from the config file.
For example it would be great if I could write something like this in the `.cobalt.yml`:

```yaml
theme: base16-ocean.dark
```

I have a rough prototype here: https://github.com/cobalt-org/cobalt.rs/compare/master...mre:config-theme

Unfortunately I need some more guidance for the next steps.

The part around `CodeBlock`s is not working yet.
What I wanted to do is inject the config as a parameter into the codeblock.
But the codeblock needs to implement `Renderable` from Liquid. 
(https://github.com/cobalt-org/cobalt.rs/compare/master...mre:config-theme#diff-b9237993dc7f8fb4f95923e1c2a64499R46)

The error message I get is:
```Rust
# cargo build --features="syntax-highlight"
   Compiling cobalt-bin v0.6.1 (file:///.../cobalt.rs)
error[E0281]: type mismatch: the type `fn(&str, &[liquid::Token], std::vec::Vec<liquid::lexer::Element>, &liquid::LiquidOptions, &config::Config) -> std::result::Result<Box<liquid::Renderable>, liquid::Error> {syntax_highlight::initialize_codeblock}` implements the trait `
for<'r, 'r, 'r, 'r> std::ops::Fn<(&'r str, &'r [liquid::Token], std::vec::Vec<liquid::lexer::Element>, &'r liquid::LiquidOptions, &'r config::Config)>`, but the trait `for<'r, 'r, 'r> std::ops::Fn<(&'r str, &'r [liquid::Token], std::vec::Vec<liquid::lexer::Element>, &'r li
quid::LiquidOptions)>` is required (expected a tuple with 4 elements, found one with 5 elements)
   --> src/document.rs:272:56
    |
272 |         options.blocks.insert("highlight".to_string(), Box::new(initialize_codeblock));
    |                                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: required for the cast to the object type `for<'r, 'r, 'r> std::ops::Fn(&'r str, &'r [liquid::Token], std::vec::Vec<liquid::lexer::Element>, &'r liquid::LiquidOptions) -> std::result::Result<Box<liquid::Renderable + 'static>, liquid::Error> + 'static`
```

That should be the last roadblock. Maybe you can give me a hint on what I could try.

boxofrox:
These patches build on @mre's initial work.  Only the default themes included by syntect are allowed.

~~I reverted @mre's changes to `Config`, `Document`, and `CodeBlock` since we only need to set the theme name once, and the `SETUP` variable already distributes themes to the `initialize_codeblock` and `decorate_markdown` functions.~~

~~I might be able to avoid the reversion with closures, but this PR is my third attempt, and I'd rather get feedback before I spend more time on this.~~

I tested the PR via the following process on Linux:

```
cd /tmp
git clone git@github.com:boxofrox/cobalt.rs.git
cd cobalt.rs
git checkout config-theme
cargo install --force cobalt-bin --path ./ --features=syntax-highlight
cd <cobalt-blog-directory>
cat "syntax-highlight:\n  theme: base16-ocean.light" >> .cobalt.yml
cobalt build
cobalt serve
# open localhost:3000 in browser
```

Use `cobalt list-syntax-themes` to view a list of available themes.

If acceptable to @mre's needs, fixes #202.  Otherwise, let's discuss what's missing.

*Updates:*

- Switch to patch set using closures instead of global `RefCell` variable.
- Syntax theme configured in `.cobalt.yml` with
    ```yaml
    syntax-highlight:
      theme: base16-eighties.dark
    ```
- Rename subcommand used to list syntax themes from `syntax-themes` to `list-syntax-themes`.
- Fix clippy lint errors.

### DIFF
--- a/src/cobalt.rs
+++ b/src/cobalt.rs
@@ -183,8 +183,12 @@ pub fn build(config: &Config) -> Result<()> {
 
         let mut context = post.get_render_context(&simple_posts_data);
 
-        post.render_excerpt(&mut context, source, config)?;
-        let post_html = post.render(&mut context, source, &layouts, &mut layouts_cache, config)?;
+        post.render_excerpt(&mut context, source, &config.syntax_theme)?;
+        let post_html = post.render(&mut context,
+                                    source,
+                                    &layouts,
+                                    &mut layouts_cache,
+                                    &config.syntax_theme)?;
         create_document_file(post_html, &post.file_path, dest)?;
     }
 
@@ -224,7 +228,11 @@ pub fn build(config: &Config) -> Result<()> {
         }
 
         let mut context = doc.get_render_context(&posts_data);
-        let doc_html = doc.render(&mut context, source, &layouts, &mut layouts_cache, config)?;
+        let doc_html = doc.render(&mut context,
+                                  source,
+                                  &layouts,
+                                  &mut layouts_cache,
+                                  &config.syntax_theme)?;
         create_document_file(doc_html, doc.file_path, dest)?;
     }
 
--- a/src/cobalt.rs
+++ b/src/cobalt.rs
@@ -183,8 +183,8 @@ pub fn build(config: &Config) -> Result<()> {
 
         let mut context = post.get_render_context(&simple_posts_data);
 
-        post.render_excerpt(&mut context, source)?;
-        let post_html = post.render(&mut context, source, &layouts, &mut layouts_cache)?;
+        post.render_excerpt(&mut context, source, config)?;
+        let post_html = post.render(&mut context, source, &layouts, &mut layouts_cache, config)?;
         create_document_file(post_html, &post.file_path, dest)?;
     }
 
@@ -224,7 +224,7 @@ pub fn build(config: &Config) -> Result<()> {
         }
 
         let mut context = doc.get_render_context(&posts_data);
-        let doc_html = doc.render(&mut context, source, &layouts, &mut layouts_cache)?;
+        let doc_html = doc.render(&mut context, source, &layouts, &mut layouts_cache, config)?;
         create_document_file(doc_html, doc.file_path, dest)?;
     }
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -39,6 +39,7 @@ pub struct Config {
     pub ignore: Vec<String>,
     pub excerpt_separator: String,
     pub dump: Vec<Dump>,
+    pub theme: String,
 }
 
 impl Default for Config {
@@ -61,6 +62,7 @@ impl Default for Config {
             ignore: vec![],
             excerpt_separator: "\n\n".to_owned(),
             dump: vec![],
+            theme: "base16-ocean.dark".to_owned(),
         }
     }
 }
@@ -139,6 +141,10 @@ impl Config {
 
         if let Some(excerpt_separator) = yaml["excerpt_separator"].as_str() {
             config.excerpt_separator = excerpt_separator.to_owned();
+        };
+
+        if let Some(theme) = yaml["theme"].as_str() {
+            config.theme = theme.to_owned();
         };
 
         Ok(config)
--- a/src/config.rs
+++ b/src/config.rs
@@ -5,6 +5,9 @@ use std::io::Read;
 use error::Result;
 use yaml_rust::YamlLoader;
 
+#[cfg(all(feature="syntax-highlight", not(windows)))]
+use syntax_highlight::has_syntax_theme;
+
 arg_enum! {
     #[derive(Debug, PartialEq, Copy, Clone)]
     pub enum Dump {
@@ -143,9 +146,17 @@ impl Config {
             config.excerpt_separator = excerpt_separator.to_owned();
         };
 
-        if let Some(theme) = yaml["syntax-highlight"]["theme"].as_str() {
-            config.syntax_theme = theme.to_owned();
-        };
+        #[cfg(all(feature="syntax-highlight", not(windows)))]
+        {
+            if let Some(theme) = yaml["syntax-highlight"]["theme"].as_str() {
+                if has_syntax_theme(theme) {
+                    config.syntax_theme = theme.to_owned();
+                } else {
+                    warn!("Theme named '{}' is not found.  Using default 'base16-ocean.dark'.",
+                          theme);
+                }
+            };
+        }
 
         Ok(config)
     }
--- a/src/config.rs
+++ b/src/config.rs
@@ -39,7 +39,7 @@ pub struct Config {
     pub ignore: Vec<String>,
     pub excerpt_separator: String,
     pub dump: Vec<Dump>,
-    pub theme: String,
+    pub syntax_theme: String,
 }
 
 impl Default for Config {
@@ -62,7 +62,7 @@ impl Default for Config {
             ignore: vec![],
             excerpt_separator: "\n\n".to_owned(),
             dump: vec![],
-            theme: "base16-ocean.dark".to_owned(),
+            syntax_theme: "base16-ocean.dark".to_owned(),
         }
     }
 }
@@ -143,8 +143,8 @@ impl Config {
             config.excerpt_separator = excerpt_separator.to_owned();
         };
 
-        if let Some(theme) = yaml["theme"].as_str() {
-            config.theme = theme.to_owned();
+        if let Some(theme) = yaml["syntax-highlight"]["theme"].as_str() {
+            config.syntax_theme = theme.to_owned();
         };
 
         Ok(config)
--- a/src/config.rs
+++ b/src/config.rs
@@ -152,7 +152,7 @@ impl Config {
                 if has_syntax_theme(theme) {
                     config.syntax_theme = theme.to_owned();
                 } else {
-                    warn!("Theme named '{}' is not found.  Using default 'base16-ocean.dark'.",
+                    warn!("Theme named '{}' is not found.  Using default 'base16-ocean.dark'.  Run `cobalt list-syntax-themes` to see available themes.",
                           theme);
                 }
             };
--- a/src/config.rs
+++ b/src/config.rs
@@ -5,7 +5,7 @@ use std::io::Read;
 use error::Result;
 use yaml_rust::YamlLoader;
 
-#[cfg(all(feature="syntax-highlight", not(windows)))]
+#[cfg(all(feature="syntax-highlight"))]
 use syntax_highlight::has_syntax_theme;
 
 arg_enum! {
@@ -146,14 +146,26 @@ impl Config {
             config.excerpt_separator = excerpt_separator.to_owned();
         };
 
-        #[cfg(all(feature="syntax-highlight", not(windows)))]
+        #[cfg(all(feature="syntax-highlight"))]
         {
+            use error::{Error, ErrorKind};
+
             if let Some(theme) = yaml["syntax-highlight"]["theme"].as_str() {
-                if has_syntax_theme(theme) {
-                    config.syntax_theme = theme.to_owned();
-                } else {
-                    warn!("Theme named '{}' is not found.  Using default 'base16-ocean.dark'.  Run `cobalt list-syntax-themes` to see available themes.",
-                          theme);
+                match has_syntax_theme(theme) {
+                    Ok(true) => config.syntax_theme = theme.to_owned(),
+                    Ok(false) => {
+                        info!(concat!("Theme named '{}' is not found.",
+                                      "  Using default 'base16-ocean.dark'.",
+                                      "  Run `cobalt list-syntax-themes` to see available themes."),
+                              theme)
+                    }
+                    Err(Error(ErrorKind::UnsupportedPlatform(..), _)) => {
+                        info!(concat!("Theme named '{}' ignored.",
+                                      "  This build of cobalt does not include support",
+                                      " for syntax highlighting."),
+                              theme)
+                    }
+                    Err(_) => unreachable!(),
                 }
             };
         }
--- a/src/main.rs
+++ b/src/main.rs
@@ -207,12 +207,8 @@ fn run() -> Result<()> {
                                  .value_name("COMMIT-MESSAGE")
                                  .help("Commit message that will be used on import")
                                  .default_value("cobalt site import")
-                                 .takes_value(true)));
-
-    #[cfg(all(feature="syntax-highlight"))]
-    let app_cli =
-        app_cli
-            .subcommand(SubCommand::with_name("list-syntax-themes").about("list available themes"));
+                                 .takes_value(true)))
+        .subcommand(SubCommand::with_name("list-syntax-themes").about("list available themes"));
 
     let global_matches = app_cli.get_matches();
 
@@ -418,7 +414,9 @@ fn run() -> Result<()> {
             }
 
             #[cfg(not(feature="syntax-highlight"))]
-            bail!(global_matches.usage());
+            bail!(concat!("No themes available.",
+                          "  This build of cobalt does not include",
+                          " support for syntax highlighting."));
         }
 
         _ => {
--- a/src/main.rs
+++ b/src/main.rs
@@ -209,7 +209,7 @@ fn run() -> Result<()> {
                                  .default_value("cobalt site import")
                                  .takes_value(true)));
 
-    #[cfg(all(feature="syntax-highlight", not(windows)))]
+    #[cfg(all(feature="syntax-highlight"))]
     let app_cli =
         app_cli
             .subcommand(SubCommand::with_name("list-syntax-themes").about("list available themes"));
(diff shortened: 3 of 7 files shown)


mre:
Highly appreciated! Thanks @boxofrox. 😃 

>  reverted @mre's changes to Config, Document, and CodeBlock

Whoops. Sorry for the inconvenience.

boxofrox:
No inconvenience.  I wouldn't know where to start on this without your initial attempt.  I tried to keep that code as I prefer local/scoped variables over global vars, but my first two attempts ran into issues with `liquid::Block` being a `Fn` type instead of a trait.

Cheers

boxofrox:
Awww... AppVeyor doesn't like my mutable variable behind a feature gate.  :slightly_frowning_face:
Rewrite!

src/main.rs
```diff
@@ -204,8 +207,13 @@ fn run() -> Result<()> {
                                  .value_name("COMMIT-MESSAGE")
                                  .help("Commit message that will be used on import")
                                  .default_value("cobalt site import")
-                                 .takes_value(true)))
-        .get_matches();
+                                 .takes_value(true)));
+
+    #[cfg(all(feature="syntax-highlight", not(windows)))]
+    let app_cli = app_cli.subcommand(SubCommand::with_name("syntax-themes")
+                        .about("list available themes"));
```
epage: Would it simplify things if we went the "null object" route and made `list_themes` return nothing on Windows?

It let's us isolate the platform checks to just a few areas

src/syntax_highlight.rs
```diff
@@ -34,7 +36,8 @@ unsafe impl Sync for Setup {}
 lazy_static!{
     static ref SETUP: Setup = Setup {
         syntax_set: SyntaxSet::load_defaults_newlines(),
-        theme_set: ThemeSet::load_defaults()
+        theme_set: ThemeSet::load_defaults(),
+        theme_name: RefCell::new(DEFAULT_THEME_NAME.to_owned()),
```
epage: This feels a little dirty to me.

I'll need to look at this more to see if there is another reasonable feasible way.

src/main.rs
boxofrox: Agreed.

src/syntax_highlight.rs
boxofrox: It doesn't give me a warm fuzzy feeling either.

It's just simple and works for this situation.  Should we want to add to it later to, I dunno, support varying syntax themes or anything that needs to change `theme_name` more than once, then it's worth overhauling the code to do away with this.

I might also be able to rewrite this with closures, but I fear implementation details in `syntax_highlight` will leak into the `document` module since Rust closures cannot be returned from a function, afaik.  I'll see what I come up with.

See my [other comment](https://github.com/cobalt-org/cobalt.rs/issues/202#issuecomment-313758943) regarding changing `liquid::Block` into a trait as one option.  I'd be happy to submit a PR for this.

Thanks for the review.

boxofrox:
Rebased all patches and applied `cargo fmt` to each using `rustfmt` v0.8.3.

src/syntax_highlight.rs
boxofrox: I found a way with closures.  You can [view the alternate patch set](https://github.com/boxofrox/cobalt.rs/commits/config-theme-closure-example) or [compare against this PR](https://github.com/boxofrox/cobalt.rs/compare/config-theme...boxofrox:config-theme-closure-example).  With closures I was able to drop the reversion path and `RefCell`.

Let me know if the alternate is preferred and I'll swap it into this PR.

boxofrox:
Minor edit on [18c7e97](https://github.com/cobalt-org/cobalt.rs/pull/259/commits/18c7397f5292c6d21e0762d2ba6ddb7373dc6360#diff-639fbc4ef05b315af92b4d836c31b023R411).  `cobalt syntax-themes` will use `println!` instead of `info!` for use in shell scripts, should anyone desire such.

boxofrox:
One point of concern.  I named the CLI subcommand `syntax-themes` thinking it may be possible to have themes for some other aspect of the blog.  But the yaml setting was left as `theme`.  If some other component of a cobalt blog becomes "theme-able", then renaming the setting to `syntax_theme`, or some such, may cause problems for users.

Maybe this can wait until we cross that bridge, maybe it doesn't matter.  Blogging is unfamiliar territory for me.

What say you?

epage:
> One point of concern. I named the CLI subcommand syntax-themes thinking it may be possible to have themes for some other aspect of the blog. But the yaml setting was left as theme. If some other component of a cobalt blog becomes "theme-able", then renaming the setting to syntax-theme, or some such, may cause problems for users.

> Maybe this can wait until we cross that bridge, maybe it doesn't matter. Blogging is unfamiliar territory for me.

I think it'd be good to have its name more clearly describe what it means.

I'm "almost done" with my change to using serde for documents which includes some breaking changes (#257).  Cobalt config will go through a similar clean up afterwards.  So the main value of settling on a term now is less likely for people to break (I'm tempted to write migration tools though to minimize the pain though, it'll just slow down my PR even further :( ).

uwearzt:
Hi,

maybe we should prefix with the highlighting engine? Something like ```syntax-syntect-theme```, there is already a issue for adding different theming engines. Then we would need a config like
```
syntax-highlighting=syntect
syntax-syntect-theme=darcula
```

boxofrox:
It is yaml.  How about:

```yaml
syntax-highlight:
  engine: syntect
  theme: darcula
```

uwearzt:
Yep, that looks nice :),

@boxofrox Sorry I was confused by https://github.com/cobalt-org/cobalt.rs/issues/200, but this one refers to highlight vim syntax, not with vim.

boxofrox:
@uwearzt ah, ok.  In that case, it appears we're sticking with only syntect for now.

For now, let's leave out the `engine` field.  Should cobalt later support other highlight engines, it'll be trivial to add it, I think.

```yaml
syntax-highlight:
  theme: base16-eighties.dark
```

As for custom themes, I haven't figured out how syntect handles that.  If I figure it out, I'll submit a PR to support them.  My plan there is if `theme` is a path to a file, then load it into the `SyntaxSet`, otherwise, try to use `theme` as name for default themes, and finally, fall back to 'base16-ocean.dark'.  For example:

```yaml
syntax-highlight:
  theme: syntax/darcula.tmTheme
```

epage:
> As for custom themes, I haven't figured out how syntect handles that. If I figure it out, I'll submit a PR to support them. My plan there is if theme is a path to a file, then load it into the SyntaxSet, otherwise, try to use theme as name for default themes, and finally, fall back to 'base16-ocean.dark'. For example:

Don't wait on perfection; it can always be added in a later PR>

As for how it should behave, I was thinking that we could have a `syntax` directory under the `_cobalt_ directory we've been talking in #211.  Under that could be a `syntax` directory.  Those syntases could act as if they are built-in to syntect.  Maybe if a theme name isn't in syntect, we append an extension and load it from this directory.

src/syntax_highlight.rs
epage: Heh, no wonder you haven't replied to me, I forgot to hit submit...

boxofrox:
> Don't wait on perfection; it can always be added in a later PR

Certainly.  I haven't made any changes to this patch set regarding the yaml discussion because I wasn't sure whether I was churning patches too much as it is for review purposes.  I wanted to get your thoughts, @epage, on my [previous comment](https://github.com/cobalt-org/cobalt.rs/pull/259#discussion_r126252855) before I attached the yaml patch.  Then again, I probably misunderstood what you meant by that statement.

> Maybe if a theme name isn't in syntect, we append an extension and load it from this directory.

Will do.

src/syntax_highlight.rs
boxofrox: :grin:

My pros (:+1:) and cons (:-1:) are pretty subjective.

The alternate patch set...

1.  :+1: Avoids using a global variable `SETUP.theme_name` in favor of scoped variable.
2.  :+1: Simplifies the PR by dropping two patches.
4.  :+1: Puts `theme` in `Config` (I'm assuming this continues to keep all config file options within one struct).
5.  :neutral_face: Wraps `initialize_codeblock()` in a closure.
6.  :neutral_face: No `RefCell`.  I'm not sure why this would matter.

The original patch set...

1.  :+1: Keeps `theme_name` organized more closely with relevant code in `syntax_highlight.rs`.
2.  :-1: Uses the global variable mentioned above.
3.  :-1: Reverts quite a bit of @mre's initial work (which was heading in the direction I favored).
4.  :-1: Stores a config file option separate from `Config` struct.
5.  :neutral_face: Works now, but global vars tend to cause problems later.

My OCD prefers the alternate patch set.  I would have submitted it instead had I figured it out sooner.

src/cobalt.rs
```diff
@@ -190,7 +190,7 @@ pub fn build(config: &Config) -> Result<()> {
         }
 
         let mut context = doc.get_render_context(&posts_data);
-        let doc_html = try!(doc.render(&mut context, source, &layouts, &mut layouts_cache));
+        let doc_html = try!(doc.render(&mut context, &source, &layouts, &mut layouts_cache));
```
boxofrox: This change is squashed across two commits.

Bad revert on my part.  @mre had appended a `&config` parameter and switched `source` to a reference.  `&config` wasn't necessary if using a global variable to store the theme name, and I failed to restore `&source` to `source`.

If we go with the alternate patch set, then the reversion doesn't happen.  If you prefer to keep `source` over `&source`, I'll need to update that patch set.

src/syntax_highlight.rs
epage: Sounds like we should go with the alternative then :)

Thanks for all your work on this!

src/cobalt.rs
epage: iirc clippy will complain about a useless `&`

src/syntax_highlight.rs
boxofrox: My pleasure.  Swapping out the patches now.

boxofrox:
Alternate patch set is ready to go! (pending CI)

- Use closures instead of global RefCell variable.
- Syntax theme configured in .cobalt.yml with
    ```yaml
    syntax-highlight:
      theme: base16-eighties.dark
    ```
- Rename subcommand used to list syntax themes from `syntax-themes` to `list-syntax-themes`.
- Fix clippy lint errors.

src/config.rs
```diff
@@ -133,6 +138,18 @@ impl Config {
...
+            if let Some(theme) = yaml["syntax-highlight"]["theme"].as_str() {
+                if has_theme(theme) {
+                    config.theme = theme.to_owned();
+                } else {
+                    warn!("Theme named '{}' is not loaded. Using default 'base16-ocean.dark'.",
+                          theme);
+                }
+            };
+        }
+
```
epage: I think i'd rather isolate the feature knowledge by having this run unconditionally.

In thinking of the case where someone is developing a blog on windows and deploying with TravisCI, it'd be nice if `has_theme` unconditionally returned true.  This is a bit of a lie but then again, no validation is performed otherwise because of the feature toggle.

src/main.rs
```diff
@@ -204,8 +207,14 @@ fn run() -> Result<()> {
...
                                  .help("Commit message that will be used on import")
                                  .default_value("cobalt site import")
-                                 .takes_value(true)))
-        .get_matches();
+                                 .takes_value(true)));
+
+    #[cfg(all(feature="syntax-highlight"))]
+    let app_cli =
+        app_cli
+            .subcommand(SubCommand::with_name("list-syntax-themes").about("list available themes"));
```
epage: You made list_syntax_themes run regardless of whether the feature is supported or not, so why not take advantage of it and remove the feature toggles here?

src/syntax_highlight.rs
```diff
@@ -41,33 +39,33 @@ lazy_static!{
...
 
 pub struct DecoratedParser<'a> {
     h: Option<HighlightLines<'a>>,
     parser: Parser<'a>,
+    config: &'a Config,
 }
 
 impl<'a> DecoratedParser<'a> {
-    pub fn new(parser: Parser<'a>) -> Self {
+    pub fn new(parser: Parser<'a>, config: &'a Config) -> Self {
```
epage: I'm a little iffy on `Config` being passed everywhere.

It starts to turn it into a god object which will make unit testing more difficult.  I know it saved a little on parameter lists with `excerpt_separator` but in that case I'm actually removing that in change I'm preparing to post.

src/main.rs
boxofrox: I think my reasoning at the time was without `features=syntax-highlight`, then:

1.  `list-syntax-themes` is an invalid subcommand,
2.  don't not show `list-syntax-themes` when `cobalt help` runs,
3.  bail in the subcommand match to mimic invoking the default match arm when the user invoked `list-syntax-themes` as an invalid subcommand.

Though (3) doesn't bail at the moment since *clap* generates an error for invalid subcommands.

Sounds like you don't mind `list-syntax-themes` being an available subcommand regardless whether the build includes *syntect*, so I can drop the feature toggle at [212](https://github.com/cobalt-org/cobalt.rs/pull/259#discussion-diff-126850433R212), and instead of mimicking the default subcommand behavior, bail with a message on stderr like "No themes available.  This build of cobalt does not include support for syntax highlighting".

src/config.rs
boxofrox: TravisCI will catch the `warn!` and fail?  If not, then I'd rather `has_theme` returns false and generates the non-fatal warning.

I'm opposed to sending mixed signals to users.  If a Windows user sets the `theme` field in `.cobalt.yml`, I don't want to give them an indication that cobalt used their syntax theme and then they see the code blocks on their site aren't highlighted.

Another option is to change the `warn!` to `info!` if TravisCI complains about the former and ignores the latter.  It reduces the visibility of the message, but it's a trade-off worth making if the user is informed with cobalt's actual behavior.

If neither option is suitable, I'll defer and update `has_theme` to return false.

src/syntax_highlight.rs
boxofrox: Fixed in commit 'Remove use of Config as god object'.

src/main.rs
boxofrox: Fixed in commit 'Remove feature-gate on `list-syntax-themes` subcommand'.

src/config.rs
boxofrox: Possible fix in commit 'Isolate platform checks for `has_syntax_theme`'.

There's a bit more code to it, but I like the explicitness.  If the platform is supported, return `Ok(bool)`, otherwise return an `Err(reason)`.  An additional benefit is the user is notified why an unsupported feature they're trying to use doesn't work.  Using an `Option` reduces the amount of code in the patch, but seemed like a [possibly] appropriate use of `Option`.  `Option` was definitely less explicit in it's intent.

Overall, the commit is not as succinct as returning a `vec![]` was with `list_syntax_themes`, but I wanted to put this implementation on the table in case you find it preferable.  If so, I'm open to refactoring `list_syntax_themes` to use `Result` for consistency.

boxofrox:
> This is looking great! Thanks for pushing a little harder on finding a way to make this work!

My pleasure.  Thanks for taking the time to review and point out areas of improvement.

I went through the patches and renamed `has_theme` to `has_syntax_theme`.  Seemed inconsistent to have `list_syntax_themes` and `has_theme`.  Also tried to wrap long strings by sentence using `concat!`.

With the last three commits of this PR, I've addressed your second and third concerns directly.  The last patch pertains to your first concern with `list_syntax_themes`.  It demonstrates a solution using `Result<bool>` to handle unsupported platforms, as I mentioned in my other reply.  If you still prefer returning false instead, I'll replace the last patch and make it so.

*clippy* is clear, and *rustfmt* has two minor complaints for lines it can't wrap.

boxofrox:
Those two minor complains from `rustfmt` bit me.  More aggressive wrapping with `concat!`.

epage:
Sorry for my delayed response. I've been away from my keyboard the last three days. There is a small chance that I'll get to look at it tonight, otherwise it'll be Monday at best (but I'll push hard to make Monday work; you've put a lot of work into this and its dragged on long enough).

src/main.rs
epage: I think its a matter of perspective. Is syntax highlighting optional because people want to turn it off or because it is platform specific?

I don't know the implementer's intent and so I lean towards platform specific. The benefit to it being a dedicated feature are platforms or environments where the external dep isn't as easily available and to make the conditional nature of this feature more explicit.

src/config.rs
epage: Agreed about not giving windows users false expectations. There is a tension between conveying accurate information and allowing a linux-deployed blog be written on Windows when the supported features are different.

While I've not yet looked at the diff, I wonder if I was unclear about my suggestion. If syntax highlighting is unsupported, that is when you ignore the value. That allows someone to write their blog on windows with successful builds (I consider WSL a cop out). They then get their feedback on themes when the build fails or succeeds on Travis. Logging a message when a theme is set on a platform with no syntax highlighting seems like a good compromise if a windows developer's expectations aren't met to help them realize what is going on.

boxofrox:
Thanks for heads up, and no worries.  I'd rather do this well than rush it.  We spend time on this as we find such time not otherwise needed by other priorities in our lives.

In the meantime, all the instructions are here for anyone to try out the feature before the PR is merged.  Don't feel you have to rush on my account.

Cheers.

epage:
Everything looks fantastic with the one open question of how to handle windows.  I'm fine merging this as-is and having that resolved in another PR.

PR #261, opened 2017-07-11
Add site uwe-arzt.de to README.md

uwearzt:
Hi,

my site is now also switched to Cobalt :)


uwearzt:
That was fast, thanks 🥇

PR #263, opened 2017-07-12
RFC: Stablize frontmatter 

### LINKED ISSUE #257: Make frontmatter future compatible

epage:
Steps before submission
- [x] Submit #262 and rebase this
- [ ] Write migration tool
- [ ] Update tests
- [ ] excerpt overriding (if needed?)

The frontmatter is now a strictly defined format with a `custom` field
for any user-provided values.  Access to this in path formatting or
liquid templates mirrors the documents structure.

Fixes #257

BREAKING CHANGES
- fronmatter
  - excerpt is unsupported
  - is_post is unsupported
  - date is renamed to published_date and the format is validated.
  - extends is renamed to layout
  - draft is renamed to is_draft
  - ext is now exposed as format
- page (liquid)
  - path renamed to url
  - source renamed to path
  - date renamed to published_date
  - draft is renamed to is_draft
  - custom frontmatter attributes parented under `custom`
- path variables
  - filename renamed to name
  - custom frontmatter attributes parented under `custom`
  - permalink default is changed to `/:categories/:year/:month/:day/:slug.html` for posts

### DIFF
--- a/src/cobalt.rs
+++ b/src/cobalt.rs
@@ -79,16 +79,15 @@ pub fn build(config: &Config) -> Result<()> {
         let src_path = source.join(file_path.as_path());
         let is_post = src_path.starts_with(posts_path.as_path());
 
-        let default_front = frontmatter::FrontmatterBuilder::new()
+        let mut default_front = frontmatter::FrontmatterBuilder::new()
             .set_post(is_post)
             .set_draft(false)
             .set_excerpt_separator(config.excerpt_separator.clone());
+        if is_post {
+            default_front = default_front.set_permalink(config.post_path.clone());
+        }
 
-        let doc = Document::parse(source,
-                                  &file_path,
-                                  &file_path,
-                                  default_front,
-                                  &config.post_path)?;
+        let doc = Document::parse(source, &file_path, &file_path, default_front)?;
         if !doc.front.is_draft || config.include_drafts {
             documents.push(doc);
         }
@@ -114,15 +113,15 @@ pub fn build(config: &Config) -> Result<()> {
 
             let is_post = true;
 
-            let default_front = frontmatter::FrontmatterBuilder::new()
+            let mut default_front = frontmatter::FrontmatterBuilder::new()
                 .set_post(is_post)
                 .set_draft(true)
                 .set_excerpt_separator(config.excerpt_separator.clone());
-            let doc = Document::parse(&drafts_root,
-                                      &file_path,
-                                      new_path,
-                                      default_front,
-                                      &config.post_path)?;
+            if is_post {
+                default_front = default_front.set_permalink(config.post_path.clone());
+            }
+
+            let doc = Document::parse(&drafts_root, &file_path, new_path, default_front)?;
             documents.push(doc);
         }
     }
--- a/src/document.rs
+++ b/src/document.rs
@@ -20,7 +20,6 @@ use syntax_highlight::{initialize_codeblock, decorate_markdown};
 use liquid::{Renderable, LiquidOptions, Context, Value, LocalTemplateRepository};
 
 use frontmatter;
-use datetime;
 use pulldown_cmark as cmark;
 use liquid;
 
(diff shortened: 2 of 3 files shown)


epage:
See #346

epage:
#346 supersedes this

PR #266, opened 2017-07-19
Add list-syntaxes subcommand to list file types supported by syntect

boxofrox:
I wrote this patch when I was messing around with #259.  Figured I'd toss it on here in case it could be of use to someone.  I think I wrote this when I was curious what keyword was needed on a code block I wanted to highlight, so it does that well, at least.

Would have provided the list of what languages are included in #200 (spoiler: not vimscript).

Format of output for `cobalt --list-syntaxes` is:

```
SyntaxName [FileExtension, ...]
...
```

From what I gleaned from the *syntect* docs/code, one can use the `SyntaxName` or `FileExtension` as the language token on a code block.  e.g.

    ```rust
    fn main() {}
    ```
    
    ```rs
    fn main() {}
    ```

However, I doubt that works for the `SyntaxName` or `FileExtension` that contains spaces.  I am baffled how one highlights regular expressions for javascript and python.

```sh
$ cobalt list-syntaxes | grep 'Regular Expressions'
Regular Expression [re]
Regular Expressions (Javascript) []
Regular Expressions (Python) []
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -61,4 +61,4 @@ mod frontmatter;
 mod syntax_highlight;
 
 #[cfg(all(feature="syntax-highlight"))]
-pub use syntax_highlight::list_syntax_themes;
+pub use syntax_highlight::{list_syntax_themes, list_syntaxes};
--- a/src/main.rs
+++ b/src/main.rs
@@ -43,7 +43,7 @@ use std::io::Result as IoResult;
 use std::fs::File;
 
 #[cfg(all(feature="syntax-highlight", not(windows)))]
-use cobalt::list_syntax_themes;
+use cobalt::{list_syntaxes, list_syntax_themes};
 
 error_chain! {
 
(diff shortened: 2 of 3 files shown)


epage:
Great idea!  I'm all for making the cobalt self-describing so people don't have to guess or we have to write a ton of documentation that quickly goes out of date (see [`--dump`](https://github.com/cobalt-org/cobalt.rs/blob/master/src/config.rs#L12).

epage:
Appveyor: 2 jobs failed due to appveyor infrastructure problems, ignoring.

boxofrox:
Cool. I like the utility of `--dump` for diagnostics.  Was that just an example, or did you mean to imply I should integrate `list-syntaxes` into it?

epage:
> Cool. I like the utility of --dump for diagnostics. Was that just an example, or did you mean to imply I should integrate list-syntaxes into it?

Example.  At least at the moment, dump is for saving state to disk though maybe we should generalize yours to `--debug-print themes` and `--debug-print syntaxes` (I've been meaning to rename `--dump` to `--debug-dump`).

PR #267, opened 2017-07-21
feat(syntax-highlight): Succeed on Windows

epage:
Create a dummy implementation of syntax highlighting so users can successfully run cobalt locally on a Windows system before deploying to TravisCI where the real syntax highlighting will be applied.

In addition, this leverages null objects to isolate the knowledge of the syntax-highlight feature with tests for the various cases.


epage:
@boxofrox: thoughts?

epage:
For Appveyor
http://help.appveyor.com/discussions/problems/7115-curl-cant-download-files-through-ssl#comment_43044339

Workaround: https://github.com/rust-lang-nursery/rand/commit/bb78689

src/config.rs
```diff
@@ -147,29 +146,20 @@ impl Config {
...
+                }
+                Ok(false) => Err(format!("Theme '{}' is unsupported", theme).into()),
+                Err(err) => {
+                    warn!("Theme named '{}' ignored. Reason: {:?}", theme, err);
+                    Ok(())
                 }
             };
-        }
+            result?;
+        };
```
boxofrox: With `feature=syntax-highlight`, an invalid `syntax-highlight.theme` entry in `.cobalt.yml` is a fatal error.  This is intentional?

I had this fall back to a sane default.  Thought it might make for a better user experience, but I can see the advantage to the fatal error, especially for automated deployments.

src/syntax_highlight/mod.rs
```diff
@@ -0,0 +1,9 @@
+#[cfg(all(feature="syntax-highlight", not(windows)))]
+mod syntect;
+#[cfg(all(feature="syntax-highlight", not(windows)))]
+pub use self::syntect::*;
+
+#[cfg(any(not(feature="syntax-highlight"), windows))]
+mod null;
+#[cfg(any(not(feature="syntax-highlight"), windows))]
+pub use self::null::*;
```
boxofrox: Heh.  I started doing this with a patch I'm tinkering with for the dependency injection idea we discussed over on #200.  Glad to see I was on the same track with this idea.

src/syntax_highlight/null.rs
```diff
@@ -0,0 +1,170 @@
+use liquid;
+use liquid::Token::{self, Identifier};
+use liquid::lexer::Element::{self, Expression, Tag, Raw};
+
+use pulldown_cmark as cmark;
+
+use error;
+
+pub fn has_syntax_theme(_name: &str) -> error::Result<bool> {
+    bail!("Themes are unsupported in this build.");
```
boxofrox: Nitpick: "Syntax themes ares unsupported in this build."

src/config.rs
epage: My preference would be to let the user know if we aren't doing as they say unless we come up with a reason otherwise.

src/syntax_highlight/null.rs
```diff
@@ -0,0 +1,170 @@
...
+}
+
+pub fn list_syntax_themes<'a>() -> Vec<&'a String> {
+    vec![]
+}
+
+pub fn list_syntaxes<'a>() -> Vec<String> {
+    vec![]
+}
+
```
boxofrox: I think we should bail on `list_syntax_themes` and `list_syntaxes`, like with `has_syntax_theme`.  These two functions are intended to only drive the CLI, and shouldn't be used anywhere else in cobalt.  But that can be the focus of another patch.

src/syntax_highlight/null.rs
epage: I decided to apply this idea to `config.rs` since it'll be more universal.

src/syntax_highlight/null.rs
epage: For a null "object", these are acting the most correct without them bailing.  No syntaxes and no themes are supported and they are accurately representing that.  `has_syntax_theme` is the one that is not acting as nicely but that is for the idea UX.

epage:
For Appveyor
http://help.appveyor.com/discussions/problems/7115-curl-cant-download-files-through-ssl#comment_43044339

Workaround: https://github.com/rust-lang-nursery/rand/commit/bb78689

or even better
```
init:
  - ps: Disable-NetFirewallRule -DisplayName 'Core Networking - Group Policy (LSASS-Out)'
```
https://users.rust-lang.org/t/appveyor-builds-failing-today/11949/5

src/syntax_highlight/null.rs
boxofrox: I make a distinction between a) no syntaxes/themes due to unsupported feature vs b) no syntaxes/themes due to misconfiguration, though cobalt isn't capable of (b), yet, unless syntect changes/breaks something with the defaults it loads.  This distinction may not be necessary, so I'm interested to see how the null "object" works out.

Issue #199, opened 2017-03-09
Make site-wide configuration forward compatible


epage:
See https://jekyllrb.com/docs/configuration/

Prior art:
- [Jekyll](https://jekyllrb.com/docs/configuration/)
- [Hugo](https://gohugo.io/getting-started/configuration/)

# Proposal

- src -> source
- dest -> destination
  - default: ./ -> ./_site
- template_extensions
- ignore
- layouts -> layouts_dir
- include_drafts (just as descriptive as alternatives)
- posts (new top-level)
  - dir (formerly `posts`)
    - default: posts -> _posts (since default doesn't include the original path)
  - sort (formerly `post_order`)
    - Add a `None` option
  - rss (moved from root)
  - default (frontmatter format)
    - root excerpt_separator moves here
    - post_path moves here as permalink
      - default ? -> date alias
- page (new top-level)
  - default (frontmatter format)
    - root excerpt_separator moves here
- site
  - name (moved from root)
    - Used by RSS
    - Added to `site.*`
  - description (moved from root)
    - Used by RSS
    - Added to `site.*`
  - base_url (was root `link`)
    - Used by RSS
    - Added to `site.*`

Removed
- `drafts` (use `is_draft` instead)
  - Note: this will break at least two of the example sites

Issue #170, opened 2016-12-12
Watch command loops endlessly


nokaa:
I ran `cobalt watch` on my project and then ran `cd /my/project/dir` in another terminal. Cobalt starts to endlessly rebuild the project.

Work around: add an `ignore` for your `dest` folder, like 
```
ignore: ["build/*"]
```

nott:
@nokaa what version of cobalt are you using? a similar issue was fixed in 0.4

nokaa:
I am using 0.4.0. I installed cobalt from crates.io a few days ago.

LucioFranco:
@nokaa what is in your `.cobalt.yml`?

nokaa:
```
rss: rss.xml
name: name
description: description
link: https://example.com
```

I'm pretty sure this was happening without a `.cobalt.yml` though.

LucioFranco:
@nokaa So it might be picking up a change in the folder you are running it from. Maybe try and add an ignore to the config to ignore the files that are changing and causing it to rebuild. If that is not the case. What OS are you on?

nokaa:
It looks as though this only happens if I have a `.git` folder. My shell displays the current status of a git repository; so my guess is that it causes some change in one of the `.git` files.

Adding `ignore: [.git/]` to my config doesn't seem to do anything.

LucioFranco:
@nokaa try `ignore: ".git/*"`

nokaa:
That didn't work. I can try building from source and debugging it in a couple days if you'd like.

nokaa:
I happened to find out today that the correct ignore syntax would be `ignore: [".git/*"]`, which solves this issue.

mrkgnao:
I just created a Cobalt site locally and have the same problem without having changed anything in `.cobalt.yml` whenever I run `cobalt watch` from inside the Cobalt site directory.

Running it from a different location triggers a full rebuild (I understand incremental builds aren't there yet), but is there any way to get `watch` to work without `cd`ing out elsewhere (or writing a wrapper script, etc.)?

epage:
Let me see if I understand.

When you run `cobalt watch` from inside the same directory as your `.cobalt.yml`, your project endlessly rebuilds?

And if your run `cobalt watch -c <PATH.YML>` from a different location, it doesn't endlessly loop but rebuilds when it is supposed to?

What version of cobalt?  What does your `.cobalt.yml` look like?

mrkgnao:
[...]

PR #278, opened 2017-09-27
Add SCSS support

### LINKED ISSUE #99: Include sass
It would be cool to support SASS. https://github.com/compass-rs/sass-rs

gnunicorn:
This is an attempt to implement supporting of compiling SASS/SCSS files, too.

It's using the nice sass-rs (which in turn uses sass-sys, to bind to libsass)
in order to compile statics files with the extension 'scss' into css files.

Progress / todo:

 - [x] add test for scss with
   - [x] imports
   - [x] nested imports
   - [x] variables
 - [x] implement rudimentary compiling scss->css
 - [x] enable imports form default `${sourceDir}/_sass`-folder
 - [x] make sass-import folders a configuration option
 - [x] expose other libsass-rs configuration options (like compress), too
 - [x] feature-gate sass
 - [x] fix CI
 - [ ] ~~include libsass in CI bundle~~
 
Comments, Review, Feedback and help welcome!

fixes #99

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,9 +39,12 @@ error-chain = "0.11"
 lazy_static = "0.2"
 itertools = "0.6"
 ignore = "0.2"
-sass-rs = "0.2.0"
 serde = "1.0"
 serde_yaml = "0.7"
+
+[dependencies.sass-rs]
+version = "0.2.0"
+optional = true
 
 [dependencies.hyper]
 version = "0.10"
@@ -64,4 +67,4 @@ unstable = []
 dev = []
 
 syntax-highlight = ["syntect"]
-
+sass = ['sass-rs']
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,7 @@ error-chain = "0.11"
 lazy_static = "0.2"
 itertools = "0.6"
 ignore = "0.2"
+sass-rs = "0.2.0"
 serde = "1.0"
 serde_yaml = "0.7"
 
--- a/src/cobalt.rs
+++ b/src/cobalt.rs
@@ -7,7 +7,7 @@ use liquid::Value;
 use rss;
 use jsonfeed::Feed;
 use jsonfeed;
-use sass_rs::{compile_file as sass_compile_file, Options as SassOptions};
+use sass_rs;
 
 use datetime;
 use document::Document;
@@ -239,7 +239,7 @@ pub fn build(config: &Config) -> Result<()> {
     // compile SASS along the way
     {
         info!("Copying remaining assets");
-        let mut sass_opts = SassOptions::default();
+        let mut sass_opts = sass_rs::Options::default();
         //  FIXME: make this a config option
         sass_opts.include_paths =
             vec![source.join("_sass").into_os_string().into_string().unwrap()];
@@ -267,8 +267,8 @@ pub fn build(config: &Config) -> Result<()> {
             }
             let src_file = source.join(file_path.as_path());
 
-            if file_path.extension().unwrap_or_else(|| OsStr::new("")) == "scss" {
-                let content = sass_compile_file(src_file.as_path(), sass_opts.clone())?;
+            if file_path.extension() == Some(OsStr::new("scss")) {
+                let content = sass_rs::compile_file(src_file.as_path(), sass_opts.clone())?;
                 let mut dest_file = dest.join(file_path.clone());
                 dest_file.set_extension("css");
 
--- a/src/cobalt.rs
+++ b/src/cobalt.rs
@@ -7,6 +7,8 @@ use liquid::Value;
 use rss;
 use jsonfeed::Feed;
 use jsonfeed;
+
+#[cfg(feature = "sass")]
 use sass_rs;
 
 use datetime;
@@ -239,10 +241,7 @@ pub fn build(config: &Config) -> Result<()> {
     // compile SASS along the way
     {
         info!("Copying remaining assets");
-        let mut sass_opts = sass_rs::Options::default();
-        //  FIXME: make this a config option
-        sass_opts.include_paths =
-            vec![source.join("_sass").into_os_string().into_string().unwrap()];
+
         let mut asset_files = FilesBuilder::new(source)?;
         for line in &config.ignore {
             asset_files.add_ignore(line.as_str())?;
@@ -265,27 +264,47 @@ pub fn build(config: &Config) -> Result<()> {
                     fs::create_dir_all(parent_dir)?;
                 }
             }
-            let src_file = source.join(file_path.as_path());
 
-            if file_path.extension() == Some(OsStr::new("scss")) {
-                let content = sass_rs::compile_file(src_file.as_path(), sass_opts.clone())?;
-                let mut dest_file = dest.join(file_path.clone());
-                dest_file.set_extension("css");
+            #[cfg(feature = "sass")]
+            {
+                let mut sass_opts = sass_rs::Options::default();
+                //  FIXME: make this a config option
+                sass_opts.include_paths =
+                    vec![source.join("_sass").into_os_string().into_string().unwrap()];
+                    
+                let src_file = source.join(file_path.as_path());
+                if file_path.extension() == Some(OsStr::new("scss")) {
+                    let content = sass_rs::compile_file(src_file.as_path(), sass_opts.clone())?;
+                    let mut dest_file = dest.join(file_path.clone());
+                    dest_file.set_extension("css");
 
-                let mut file =
-                    File::create(&dest_file)
-                        .map_err(|e| format!("Could not create {:?}: {}", file_path, e))?;
+                    let mut file =
+                        File::create(&dest_file)
+                            .map_err(|e| format!("Could not create {:?}: {}", file_path, e))?;
 
-                file.write_all(content.as_bytes())?;
+                    file.write_all(content.as_bytes())?;
 
-            } else {
-                let dest_file = dest.join(file_path);
-                debug!("Copying {:?} to {:?}", src_file, dest_file);
-                fs::copy(src_file.as_path(), dest_file.as_path())
-                    .map_err(|e| format!("Could not copy {:?} into {:?}: {}", src_file, dest_file, e))?;
+                } else {
+                    copy_file(src_file.as_path(), dest.join(file_path).as_path())?;
+                }
+            }
+
+            #[cfg(not(feature = "sass"))]
+            {
+
+                let src_file = source.join(file_path.as_path());
+                copy_file(src_file.as_path(), dest.join(file_path).as_path())?;
             }
         }
     }
+
+    Ok(())
+}
+
+fn copy_file(src_file: &Path, dest_file: &Path) -> Result<()> {
+    debug!("Copying {:?} to {:?}", src_file, dest_file);
+    fs::copy(src_file.as_path(), dest_file.as_path())
+        .map_err(|e| format!("Could not copy {:?} into {:?}: {}", src_file, dest_file, e))?;
 
     Ok(())
 }
--- a/src/cobalt.rs
+++ b/src/cobalt.rs
@@ -15,6 +15,8 @@ use datetime;
 use document::Document;
 use error::*;
 use config::{Config, SortOrder};
+#[cfg(feature = "sass")]
+use config::SassOutputStyle;
 use files::FilesBuilder;
 use frontmatter;
 
@@ -268,10 +270,18 @@ pub fn build(config: &Config) -> Result<()> {
             #[cfg(feature = "sass")]
             {
                 let mut sass_opts = sass_rs::Options::default();
-                //  FIXME: make this a config option
-                sass_opts.include_paths =
-                    vec![source.join("_sass").into_os_string().into_string().unwrap()];
-                    
+                sass_opts.include_paths = vec![source
+                                                   .join(&config.sass.import_dir)
+                                                   .into_os_string()
+                                                   .into_string()
+                                                   .unwrap()];
+                sass_opts.output_style = match config.sass.style {
+                    SassOutputStyle::Nested => sass_rs::OutputStyle::Nested,
+                    SassOutputStyle::Expanded => sass_rs::OutputStyle::Expanded,
+                    SassOutputStyle::Compact => sass_rs::OutputStyle::Compact,
+                    SassOutputStyle::Compressed => sass_rs::OutputStyle::Compressed,
+                };
+
                 let src_file = source.join(file_path.as_path());
                 if file_path.extension() == Some(OsStr::new("scss")) {
                     let content = sass_rs::compile_file(src_file.as_path(), sass_opts.clone())?;
@@ -303,9 +313,8 @@ pub fn build(config: &Config) -> Result<()> {
 
 fn copy_file(src_file: &Path, dest_file: &Path) -> Result<()> {
     debug!("Copying {:?} to {:?}", src_file, dest_file);
-    fs::copy(src_file.as_path(), dest_file.as_path())
+    fs::copy(src_file, dest_file)
         .map_err(|e| format!("Could not copy {:?} into {:?}: {}", src_file, dest_file, e))?;
-
     Ok(())
 }
 
--- a/src/cobalt.rs
+++ b/src/cobalt.rs
@@ -7,6 +7,7 @@ use liquid::Value;
 use rss;
 use jsonfeed::Feed;
 use jsonfeed;
+use sass_rs::{compile_file as sass_compile_file, Options as SassOptions};
 
 use datetime;
 use document::Document;
@@ -235,8 +236,13 @@ pub fn build(config: &Config) -> Result<()> {
     }
 
     // copy all remaining files in the source to the destination
+    // compile SASS along the way
     {
         info!("Copying remaining assets");
+        let mut sass_opts = SassOptions::default();
+        //  FIXME: make this a config option
+        sass_opts.include_paths =
+            vec![source.join("_sass").into_os_string().into_string().unwrap()];
         let mut asset_files = FilesBuilder::new(source)?;
         for line in &config.ignore {
             asset_files.add_ignore(line.as_str())?;
@@ -260,11 +266,24 @@ pub fn build(config: &Config) -> Result<()> {
                 }
             }
             let src_file = source.join(file_path.as_path());
-            let dest_file = dest.join(file_path);
 
-            debug!("Copying {:?} to {:?}", src_file, dest_file);
-            fs::copy(src_file.as_path(), dest_file.as_path())
-                .map_err(|e| format!("Could not copy {:?} into {:?}: {}", src_file, dest_file, e))?;
+            if file_path.extension().unwrap_or_else(|| OsStr::new("")) == "scss" {
+                let content = sass_compile_file(src_file.as_path(), sass_opts.clone())?;
+                let mut dest_file = dest.join(file_path.clone());
+                dest_file.set_extension("css");
+
+                let mut file =
+                    File::create(&dest_file)
+                        .map_err(|e| format!("Could not create {:?}: {}", file_path, e))?;
+
+                file.write_all(content.as_bytes())?;
+
+            } else {
+                let dest_file = dest.join(file_path);
+                debug!("Copying {:?} to {:?}", src_file, dest_file);
+                fs::copy(src_file.as_path(), dest_file.as_path())
+                    .map_err(|e| format!("Could not copy {:?} into {:?}: {}", src_file, dest_file, e))?;
+            }
         }
     }
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -38,6 +38,33 @@ impl Default for SortOrder {
     }
 }
 
+#[derive(Debug, Eq, PartialEq, Hash, Copy, Clone)]
+#[derive(Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub enum SassOutputStyle {
+    Nested,
+    Expanded,
+    Compact,
+    Compressed,
+}
+
+#[derive(Debug, PartialEq)]
+#[derive(Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct SassOptions {
+    pub import_dir: String,
+    pub style: SassOutputStyle,
+}
+
+impl Default for SassOptions {
+    fn default() -> SassOptions {
+        SassOptions {
+            import_dir: "_sass".to_owned(),
+            style: SassOutputStyle::Nested,
+        }
+    }
+}
+
 #[derive(Debug, PartialEq)]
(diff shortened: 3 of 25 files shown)


epage:
RE CI

- appveyor: I have no clue whats going on with that build.  I also had problems with it with liquid
- travis: looks like its just a rustfmt issue

src/cobalt.rs
```diff
@@ -7,6 +7,7 @@ use liquid::Value;
 use rss;
 use jsonfeed::Feed;
 use jsonfeed;
+use sass_rs::{compile_file as sass_compile_file, Options as SassOptions};
```
epage: If you're just going to rename them to effectively include the crate name in the names, why bother doing anything more than `use sass_rs;`

(Granted I'm biased by my general dislike of this style of `user` despite it being effectively a standard practice)

src/cobalt.rs
```diff
@@ -262,11 +268,29 @@ pub fn build(config: &Config) -> Result<()> {
                 }
             }
             let src_file = source.join(file_path.as_path());
-            let dest_file = dest.join(file_path);
 
-            fs::copy(src_file.as_path(), dest_file.as_path())
-                .map_err(|e| format!("Could not copy {:?} into {:?}: {}", src_file, dest_file, e))?;
-            debug!("Copied {:?} to {:?}", src_file, dest_file);
+            if file_path.extension().unwrap_or_else(|| OsStr::new("")) == "scss" {
```
epage: Could you do `if file_path.extension() == Some("scss")`?

Cargo.toml
```diff
@@ -39,6 +39,7 @@ error-chain = "0.11"
 lazy_static = "0.2"
 itertools = "0.6"
 ignore = "0.2"
+sass-rs = "0.2.0"
```
epage: [sass-rs](https://github.com/compass-rs/sass-rs) wraps libsass.  Is there anything about that additional dependency that we should consider making this an optional feature?

src/cobalt.rs
```diff
@@ -262,11 +268,29 @@ pub fn build(config: &Config) -> Result<()> {
...
+                let content = sass_compile_file(src_file.as_path(), sass_opts.clone())?;
+                let mut dest_file = dest.join(file_path.clone());
+                dest_file.set_extension("css");
```
epage: I'm confused.  How is the path `tests/fixtures/sass/_sass/_nested.scss` being converted into the path `tests/target/sass/style/nest.css`?

I'm not seeing you change the folder nor remove the leading `_`.

src/cobalt.rs
```diff
@@ -237,8 +238,13 @@ pub fn build(config: &Config) -> Result<()> {
     }
 
     // copy all remaining files in the source to the destination
+    // compile SASS along the way
     {
         info!("Copying remaining assets");
+        let mut sass_opts = SassOptions::default();
+        //  FIXME: make this a config option
+        sass_opts.include_paths =
+            vec![source.join("_sass").into_os_string().into_string().unwrap()];
```
epage: Adding an option for sass_dir would imply to me that that is where all the sass files should live but the implementation allows them to live anywhere.  Thoughts on which direction this should go?

epage:
> make sass-import folders a configuration option

> expose other libsass-rs configuration options (like compress), too

Not saying this is the only way but Jelyll seems to expose this as
```
sass:
    sass_dir: _sass
    style: compressed
```

https://jekyllrb.com/docs/assets/

Cargo.toml
LucioFranco: This was by far the biggest issue I ran into when I tried to initially implement this feature.

gnunicorn:
> travis: looks like its just a rustfmt issue

rustfmt here complained about a range of unrelated lines to be changed, but not the one that it complains about on travis. Seems to be cause by me running `0.8.4` here, [while travis runs on `0.8.3`](https://travis-ci.org/cobalt-org/cobalt.rs/jobs/280641369#L1460). @epage Any thoughts on how you want to handle that? Should I make it work for `0.8.3`? And/or should we consider that unrelated and upgrade to 0.8.6 as the latest version, which requires touching a bunch more lines when this is upgraded? 

> appveyor: I have no clue whats going on with that build. I also had problems with it with liquid

From a first glance at it, this appears to be caused by [libsass-sys not support win- _GNU_ environments at the moment](https://github.com/compass-rs/sass-rs/blob/master/sass-sys/build.rs#L51). Don't think this is related to any liquid-issues you might have seen. How important is the GNU-win-support? I could get in touch with the sass-maintainer and try to figure it out with them. _Disclaimer_ just a Sass-User, not involved in the development of that lib just yet ;) .

> Not saying this is the only way but Jelyll seems to expose this as

Yeah, I was thinking something along the ways of that. Hadn't yet looked at how jekyll handles it, but looks quite straight forward to me. Just would also add a `sass: false`-option in order to disable compiling alltogether.

But while we are on it, how important is jekyll-compatability with this feature or in general? Jekyll right now requires you to have a header on top of the scss-file rather than detection through the extension as this version does it. I _think_ this also allows for template-pre-processing of scss files in jekyll, though I haven't seen this being used in production yet. Would that be of interest/necessity?

Replying to the others inline.

src/cobalt.rs
gnunicorn: Good point. Will do.

Cargo.toml
gnunicorn: Well, apparently, it doesn't compile on win-GNU currently :D (see my other comment). Which we should (and I believe we can) fix.

Aside from that, I don't know. As said, I am not involved in the sass thing, just want to replace my jekyll with cobalt and need sass support ;) . As such I have no knowledge about the implications this might have. But I think it might be troublesome for some embedded archs -- does cobalt support those?

How about the following: we make this a feature, but it is on by default. I think it'd be confusing to peeps, if cobalt wouldn't ship sass per default just because some archs have trouble with it. I'd rather have people explicitly disable it, if compile fails so they know it won't support sass. What do you think?

src/cobalt.rs
gnunicorn: from my understand `_sass` is in the default ignore pattern (because of it's underscore), thus being excluded from copy and converting. `_sass/_nested` (similar to `_layout/main`) shall never be converted directly.

The "conversion" happens within the scss-compiling through the `@import`-directive. In `style/nest.scss` it is importing `_nested`, which in turn has an import for the other import file - both can be and should be found in the `_sass`-folder (the entry point for all imports).

So the only converting that happens here is `style/nest.scss` into `style/nest.css` (setting the extension), which internally imports nested and others from `_sass`.

epage: Oh, I missed the existence of ` tests/fixtures/sass/style/nest.scss`.  That makes sense now.

epage:
> rustfmt here complained about a range of unrelated lines to be changed, but not the one that it complains about on travis. Seems to be cause by me running 0.8.4 here, while travis runs on 0.8.3. @epage Any thoughts on how you want to handle that? Should I make it work for 0.8.3? And/or should we consider that unrelated and upgrade to 0.8.6 as the latest version, which requires touching a bunch more lines when this is upgraded?

I've not looked at the current state of rustfmt (been waiting for it to no longer require nightly) but the reason for 0.8.3 was I was seeing various crashes with 0.8.4.  I can give 0.8.6 a try.

LucioFranco:
When I originally tried to tackle this problem, the best solution I came up with was maybe to call some extra libs, so maybe have the end user install node-sass and then call the node-sass bin from cobalt. That way we didn't not have to depend on libsass being installed externally.

epage:
> But while we are on it, how important is jekyll-compatability with this feature or in general? Jekyll right now requires you to have a header on top of the scss-file rather than detection through the extension as this version does it. I think this also allows for template-pre-processing of scss files in jekyll, though I haven't seen this being used in production yet. Would that be of interest/necessity?

We're not aiming for Jekyll compatibility, though it would be nice to make import support not terrible.  I generally look at Jekyll, Hugo, and a couple other static site generators when implementing new features to try to learn from them rather than reinventing the problems they ran into.

iirc if there isn't a frontmatter, [cobalt will still work fine](https://github.com/cobalt-org/cobalt.rs/blob/master/src/document.rs#L47).  We could start without frontmatter/liquid support for sass and add it later if needed.

Cargo.toml
epage: I think being a feature is fine.  Unsure whether it should be a default one though.

epage: @LucioFranco 

> When I originally tried to tackle this problem, the best solution I came up with was maybe to call some extra libs, so maybe have the end user install node-sass and then call the node-sass bin from cobalt. That way we didn't not have to depend on libsass being installed externally.

epage: Is it dynamically or statically linking against libsass?  Our CIs pass.  I could see it maybe being installed on the Linux one, but Mac and Windows-msvc as well?

LucioFranco: @epage https://www.npmjs.com/package/sass here is an interesting cross-platform version of sass that might be easier for users to install (only need 3 files).

gnunicorn: If I read this right [libsass-sys is statically linking libsass](https://github.com/compass-rs/sass-sys/blob/master/build.rs#L75), otherwise I'd also not understand why it'd need sass-rs and sass-sys, why would you do two separate wrappers around the same dynamic lib? I understand sass-sys to be a static-linked-wrapper around libsass and sass-rs a nice rust-api on top. I might be wrong though.

>  I could see it maybe being installed on the Linux one, but Mac and Windows-msvc as well?

I think that is what sass-sys tries to provide. Just not yet for windows-gnu.

Either way, I once was told by @johannhof , that one aim of cobalt was (for him at least) to a have a one-file-all-batteries-included-executable (which is why we picked the syntax-highlighting that completely inlines all its lang-files, too) and I think that is a good goal to have. So I'd rather not want this to depend on any other build-environment or packager (like npm or ruby) to be present at execution-time (compile time, I think we can discuss). 

I'll investigate further how sass-rs and sass-sys actually do this and if not statically linked per default, I'd try to see if that is an option.

gnunicorn:
Sorry, partly commented on this in the inline part, @LucioFranco . I agree with not wanting to expect to have libsass install on the target-system. But rather than calling an external cli (which then needs to be present), I'd prefer the approach of statically linking libsass to ensure single-all-batteries-included-executable-drop-in-exeprience.

Cargo.toml
epage: > If I read this right sass-sys is statically linking libsass, otherwise I'd also not understand why it'd need sass-rs and sass-sys, why would you do two separate wrappers around the same dynamic lib? I understand sass-sys to be a static-linking build-environment around libsass and sass-rs a nice rust-api on top. I might be wrong though.

So it seems like it explicitly statically links on Windows but does the default build on Linux/Mac.

> I think that is what sass-sys tries to provide. Just not yet for windows-gnu.

Yeah, I didn't realize `sass-sys` built it, so I thought I could infer static linking based on what platforms might have it installed.

>  So I'd rather not want this to depend on any other build-environment or packager (like npm or ruby) to be present at execution-time (compile time, I think we can discuss).

I think the extra binary idea is meant as a work around *if* `libsass` is dynamically linked.  It effectively acts as a delay load for platforms that don't support it.  If an extra binary is needed either way, a delay load one would be nice. 

Granted, I'm more comfortable distributing for Linux with a requirement for libsass to be installed than Windows just because of the differences between the platforms (package managers).  If libsass isn't statically linked, I wonder if we could modify our CI process to get it added to our zip file.

src/cobalt.rs
epage: In that case, maybe we should call the config option

```
sass:
    import_dir: _sass
    style: compressed
```

It'll make the semantics more clear I feel.

src/config.rs
```diff
@@ -37,6 +37,35 @@ impl Default for SortOrder {
...
+    Expanded,
+    Compact,
+    Compressed,
+}
+
+#[derive(Debug, PartialEq)]
+#[derive(Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct SassOptions {
+    pub sass_dir: String,
```
epage: What were your thoughts on my suggesting we call this `import_dir`.  I feel like it'd make the semantics more clear.  This isn't a directly to hold your sass files.  This is a directory to hold the imported sass files.

src/config.rs
```diff
@@ -37,6 +37,35 @@ impl Default for SortOrder {
...
+    pub extensions: Vec<String>,
```
epage: How come this needs to be configurable?

src/config.rs
```diff
@@ -37,6 +37,35 @@ impl Default for SortOrder {
...
+    pub style: SassOutputStyle,
+}
+
+impl Default for SassOptions {
+    fn default() -> SassOptions {
+        SassOptions {
+            sass_dir: "_sass".to_owned(),
+            extensions: vec!["scss".to_owned(), "sass".to_owned()],
```
epage: Since I'm new to sass, could you explain the difference in the extensions?

src/config.rs
```diff
@@ -74,6 +103,7 @@ pub struct Config {
     #[serde(skip)]
     pub dump: Vec<Dump>,
     pub syntax_highlight: SyntaxHighlight,
+    pub sass: SassOptions,
```
epage: Have you rebased against master?  I don't think this is going to work quite as you expect.

Like with frontmatter, I've decoupled the external and internal implementations of `Config`.  You've added support for sass to the internal implementation but not the external (what we actually use to read the file).

Why did I add this extra layer of more work?

This is to open  the door for us to experiment with evolving the config format while deferring breaking compatibility.  Particularly, I want to create a conversion utility when we do break compatibility to lessen the pain.  I believe this work is being tracked in #199.

tests/mod.rs
```diff
@@ -26,6 +26,8 @@ fn run_test(name: &str) -> Result<(), cobalt::Error> {
         .unwrap_or_default();
     let destdir = TempDir::new(name).expect("Tempdir not created");
 
+    println!("Test Config {}: {:?}", name, config);
```
epage: There a reason to print what the config looks like?  Its on disk for someone to look at in case of failure.

Cargo.toml
```diff
@@ -63,4 +67,4 @@ unstable = []
 dev = []
 
 syntax-highlight = ["syntect"]
-
+sass = ['sass-rs']
```
epage: Please update .travis.yml
- `cargo check` for various feature combinations
- `cargo test` and clippy should run with all features enabled

If we feel we should have sass be included by default in our pre-built binaries, then we should also mention this features in [`before_deploy`](https://github.com/cobalt-org/cobalt.rs/blob/master/ci/before_deploy.sh#L16).

I'd recommend appveyor also being updated but I'm unsure if there is a good way for us to conditionally enable for this (because of it being unsupported for gnu).

src/config.rs
gnunicorn: I only saw your comment after I pushed it all. Used the same name for compat-reasons, but don't have any feelings on any either name. Will change it.

src/config.rs
gnunicorn: I am not super sure in this either, but from my understand sass is the old (pre version 3.0) version of the syntax and scss (standing for "Sassy CSS", which is also a Superset of CSS) now is the standard. I have seen older or long-term-maintained projects still use the old extension, though I am not even sure they are compatible or the current libsass-version does indeed support (or even detect) the old style.

Though without certainty on this, there is the argument to be made that this shouldn't include both, I suppose 🤔 .

src/config.rs
gnunicorn: Well, I come from the world of multimedia where rule number one is: if you know the file name extension, you know _nothing_ (what is in an `avi` or `mov` again? Can you play an `mp4`?). There, unless you read its first bytes until you detect something, you have no idea what you are dealing with. With that background I always feel super insecure with coding something that purely relies on file path extensions for feature or file type detection (as it is bullocks), and therefore I have the urge to _at least_ allow the user to reconfigure it, if they also chose to rename the extensions (even if that is not common place).

But especially with the argument of `scss` and `sass` I make later, there is a fair point in arguing that no one really needs or wants to configure this ever ... I most certainly never did and don't see a good reason to do so in the future. I am fine with removing it and setting it to `scss` only.

src/config.rs
gnunicorn: > Have you rebased against master?
Not since I started no.

>  I don't think this is going to work quite as you expect.

okay..... Well it did work for the test case I had added. But I suppose I should rebase against master and see for myself, as I don't quite understand what else you referring to. Have you seen the test case and would you expect it to fail? If not, could we design a test case that will fail now 
without the "frontmatter" implementation?

tests/mod.rs
gnunicorn: Ehem. That shouldn't be there...

src/config.rs
epage: Nice thing is it can always be added later if needed.

src/config.rs
epage: I defer to you to decide (or anyone else who knows better about these things).  Like above, we can always add support for it later if needed.

src/config.rs
epage: I expect the test with custom settings to fail.  Either it will report an error about unsupported fields (ideally) or it will act as if those fields weren't set.

The relevant [lines of code](https://github.com/cobalt-org/cobalt.rs/blob/master/src/config.rs#L118):
```rust
let config: wildwest::GlobalConfig = serde_yaml::from_str(&content)?;
let mut config: Config = config.into();
```

[`wildwest::GlobalConfig`](https://github.com/cobalt-org/cobalt.rs/blob/master/src/legacy/wildwest.rs#L83) represents the current `.cobalt.yml` file format.

[`config::Config`](https://github.com/cobalt-org/cobalt.rs/blob/master/src/config.rs#L54) represents what will eventually become the `.cobalt.yml` file format.

src/config.rs
gnunicorn: Agreed. Will simplify it.

src/config.rs
gnunicorn: I rebased and Added the wildwest config, but I am still unclear wether I understood you right. Do you mind taking a look?

Cargo.toml
gnunicorn: Done and done. And travis is passing, [but clippy complains on nightly, though I think that it is a false positive](https://travis-ci.org/cobalt-org/cobalt.rs/jobs/282053850#L2178-L2223).

gnunicorn:
Two final things, @epage : 

1. I did the rustfmt changes the travis version required even though at least some of them weren't even related to this code base (and using the same here gives me very different complaints) 
2. I've started to investigate why the sass-sys compile fails with windows-gnu even though [it works on the lib itself](https://github.com/compass-rs/sass-sys/pull/8). But if you prefer, we can also put this off and merge the PR first.

src/config.rs
epage: Looks great!

src/legacy/wildwest.rs
```diff
@@ -67,6 +67,33 @@ impl From<FrontmatterBuilder> for frontmatter::FrontmatterBuilder {
...
+    Nested,
+    Expanded,
+    Compact,
+    Compressed,
+}
+
+#[derive(Debug, PartialEq)]
+#[derive(Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct SassOptions {
```
epage: Someone might ask why duplicate `SassOutputStyle` and `SassOptions` rather than load it from `config`.

The benefit is that we can freely evolve the `config` version after this is submitted and released while we won't be able to change this version after release unless we are breaking compatibility and don't feel the upgrade burden will be too much.

epage:
> And travis is passing, but clippy complains on nightly, though I think that it is a false positive.

I'll get to this.

> I've started to investigate why the sass-sys compile fails with windows-gnu even though it works on the lib itself. But if you prefer, we can also put this off and merge the PR first.

My preference is a passing CI.  If for no other reason, so I can get a fix out for #170 once I get it implemented.

Option 0: Get it working as part of this PR

Option 1. Remove `sass` from the built-binaries and remove sass completely from appveyor.  Removing sass from appveyor might be the simplest way of getting it passing.  We are removing it also from the pre-built binary on Linux just to ease communication.

Option 2: Find a way for appveyor to conditionally compile sass based on if the target is GNU or not.

gnunicorn:
> Option 1. Remove sass from the built-binaries and remove sass completely from appveyor. Removing sass from appveyor might be the simplest way of getting it passing. We are removing it also from the pre-built binary on Linux just to ease communication.

I decided for this for now. We can still investigate and fix this later. I hope all checks pass now.

gnunicorn:
@epage looks like the CI is finally happy :)

epage:
Sorry for the delay!

This is looking great!  I agree about submitting now rather than wait.  It'll avoid this PR becoming stale as cobalt continues to evolve.

Now that everything is reviewed and ready to go in, could you clean up your commit history and let me know when its done?  github doesn't usually send notifications when those are done.

gnunicorn:
> Now that everything is reviewed and ready to go in, could you clean up your commit history and let me know when its done? github doesn't usually send notifications when those are done.

Sure thing. Do you have some Format-Guidelines or Styles you usually use, or you just want to me to stash it into one?

epage:
> Sure thing. Do you have some format guidelines or styles you usually use, or you just want to me to stash it into one?

If you follow [these guidelines](https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-angular/convention.md) then it'll make my life easier for creating a changelog.

As for what should be separate commits, that I leave for you to decide based on what makes sense to you, as long as each commit is standalone.

gnunicorn:
I've made it one main feature-commit and two separate "fixes" of things missing from the original feature. I couldn't find a key-word for CI-stuff, so I assumed this to be a `chore` ;) .

Hope this is good to be merged now :) .

epage:
> I've made it one main feature-commit and two separate "fixes" of things missing from the original feature. I couldn't find a key-word for CI-stuff, so I assumed this to be a chore ;) .

Yeah, `chore` is my grab bag of things that don't matter for release notes.  `fix` and `test` might also be appropriate but its all subjective.

> Hope this is good to be merged now :) .

Going in!

epage:
Would you like me to prioritize getting a release out with this feature or is it lower priority considering its not enabled by default?

src/config.rs
```diff
@@ -38,6 +38,33 @@ impl Default for SortOrder {
...
+    Expanded,
+    Compact,
+    Compressed,
+}
+
+#[derive(Debug, PartialEq)]
+#[derive(Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct SassOptions {
+    pub import_dir: String,
```
epage: Is there a use case for import_dir to be configurable?  The problem with extra configuration is that it means there is extra places corner cases to handle.  For example, we should probably enforce that this directory starts with a `_` to ensure it doesn't get picked up as assets.

I'm sorry, we talked about what to name this but I lost track of my questioning what should be hard coded or configurable.

Issue #276, opened 2017-09-26
Importing Jekyll post


Geobert:
Sub-command to import jekyll posts

Step 1: #289 

Remaining work
- [ ] Recursively walk directory
- [ ] Convert config
- [ ] Copy assets
- [ ] permalink conversion

epage:
RE frontmatter

We're in transition between frontmatter formats.  The public-facing format is closer to Jekyll and is very sloppy (a liquid Object).  The new internal-facing format is more strictly defined.  We need to finish settling on the format (#257) and write an upgrade tool (similar to what you're import will be) before we can make the swap.

What we currently do is load the original format and effectively import it into the new format, see [the code](https://github.com/cobalt-org/cobalt.rs/blob/master/src/legacy/wildwest.rs).

My suggested course of action
1. Create subcommand
2. Write code to [read document](https://github.com/cobalt-org/cobalt.rs/blob/master/src/document.rs#L237) with Jekyll frontmatter format
3. Convert jekyll frontmatter to the modern [frontmatter](https://github.com/cobalt-org/cobalt.rs/blob/master/src/frontmatter.rs)
4. Convert modern frontmatter into legacy frontmatter (effectively the opposite of what we do on load)
5. [Dump the document](https://github.com/cobalt-org/cobalt.rs/blob/master/src/document.rs#L460).

This will require more round trips but it means that when we move away from the legacy format, we're just deleting code rather than rewriting it.

epage:
RE subcommand

I've been meaning to refactor the subcommands into external programs (like cargo).  Feel free to do that as a separate PR if you want.  Eventually I'll get to it though.

epage:
RE Behavior

What do you see being the arguments this needs?  Will this work in place (modify existing files) or will it write it out to a separate directory?

A great source of inspiration is to see how other static site generators handle import.  Long term, I'd like to see multiple import types, including Evernote's xml.

Geobert:
Thanks for your inputs. First, what "RE" means? ^^' I'm French so sometime I'm missing some vocabulary ^^

I'll check on cargo's source to see how to do external program for subcommand then :)

I was thinking of something like:
`cobalt import jekyll -s <directory with jekyll's posts> -d <output dir>`

epage:
RE effectively means "regarding".

That seems reasonable.

epage:
Oh, the problem with `import` as a subcommand is that is used right now for the git integration.  We might need to rename one of them.

[...]

Issue #198, opened 2017-03-09
Make permalink substitutions future-compatible


epage:
cobalt.rs currently supports setting a `path` attribute on a document.  This `path` attribute can contain some Jekyll-compatible fields, some other custom fields, plus whatever other attributes are associated with the document.

There are a couple of challenges
1. Some unsupported Jekyll :substitutes conflict with attributes
2. Adding new :substitutes could break compatibility with attributes set by users
3. It doesn't support Jekyll's short-hand
4. The default path is not Jekyll compatible
5. There is some syntax ambiguity

# Jekyll compatible `:substitutes `

- path: relative path of the file, without output_ext.  Normally limited to collections but I think we should make it generally available
- output_ext: the post-templating file extension. Normally limited to collections but I think we should make it generally available
- collection: Name of the collection.  Only relevant to collections
- name: "slugified" file name (missing date prefix, missing extension)
- title: pretty case-preserved slugified `slug` attribute, if available, otherwise file name (missing date prefix, missing extension)
- slug: slugified `slug` attribute, if available, otherwise file name (missing date prefix, missing extension)
- categories: `downcased/categeory/names`
- year:
- month
- day
- hour
- minute
- second
- i_day
- i_month
- short_month
- short_year
- y_day

See
- https://jekyllrb.com/docs/permalinks/#template-variables
- https://jekyllrb.com/docs/collections/#permalinks

## Cobalt-extensions

As mentioned, Cobalt supports attributes in `:substitutes` which is a useful feature.  To avoid breaking people as new native substitutes are added, I propose the following syntax

`:custom.*attribute name*`

# URL Templates

- date: `/:categories/:year/:month/:day/:title.html`
- pretty: `/:categories/:year/:month/:day/:title/`
- ordinal: `/:categories/:year/:y_day/:title.html`
- none: `/:categories/:title.html`

See
- https://jekyllrb.com/docs/permalinks/#builtinpermalinkstyles

# Auto-adjusting to document type

The permalink attribute applies to posts, pages, and collections.  Pages and collections do not support dates and categories within permalinks.    Therefore, those fields will evaluate to blanks

Default style
- Post: `date`
- Page: source structure (almost `/<path>/:slug.:output_ext`).  Ideally, this could be expressed as variables so it can be tweaked.
- Collection: source structure, prefixed with `:collection` (almost `/:collection/<path>/:slug.:output_ext`).  Ideally, this could be expressed as variables so it can be tweaked.

See
[...]

PR #289, opened 2017-10-04
convert jekyll frontmatter to cobalt's one for #276

Geobert:
First step for importing jekyll posts to cobalt. I'm planning to add at least another feature: specifying the excerpt marker (for WP->jekyll documents)

I have run rustfmt but there is an issue with it: https://github.com/rust-lang-nursery/rustfmt/issues/878
and I have a trailing space in a raw string used in a test.

Step towards #276

### DIFF
--- a/src/jekyll/jk_document.rs
+++ b/src/jekyll/jk_document.rs
@@ -123,7 +123,7 @@ impl JkDocument {
         if !dest_dir.exists() {
             create_dir_all(&dest_dir)?;
         }
-        let dest_file = dest_dir.join(source_file.file_name().unwrap());
+        let dest_file = dest_dir.join(source_file.with_extension("md").file_name().unwrap());
         let mut dest = File::create(dest_file)?;
         let converted = format!("{}\n---\n{}", &front, &doc.content.unwrap());
         dest.write_all(converted.as_bytes())?;


Geobert:
forgot to run clippy

epage:
FYI you don't need to close/reopen or create a new one.  Just let me know.  I'll start reviewing it and you can post an update with clippy having run.

Geobert:
Ok my clippy errors are cleaned up. But there are some others from lines I didn't touch :-/

src/bin/cobalt/main.rs
```diff
@@ -227,9 +308,9 @@ fn run() -> Result<()> {
     let mut builder = LogBuilder::new();
     builder.format(format);
 
-    match matches
-              .value_of("log-level")
-              .or_else(|| global_matches.value_of("log-level")) {
+    match matches.value_of("log-level").or_else(|| {
```
epage: You seem to have a different version of rustfmt.  I plan to fix this all up ... when there aren't PRs outstanding to make life difficult for them.

Please undo the format changes.  It'll break the CI and make reviewing the code more difficult.

src/document.rs
```diff
@@ -44,7 +44,7 @@ fn read_document<PB: Into<PathBuf>, P: AsRef<Path>>(root: PB, relpath: P) -> Res
     Ok(text)
 }
 
-fn split_document(content: &str) -> Result<(Option<&str>, &str)> {
+pub fn split_document(content: &str) -> Result<(Option<&str>, &str)> {
```
epage: You've made this and other code above `pub`.

Something to keep in mind when reusing code is if the reused code is inherent (same requirements) or incidental (same implementation but might diverge as requirements diverge).

I'd recommend creating a `split` designed for jekyll files.

(as for `read_file`, if its going to be reused, it should probably be moved to another `.rs` file).

src/error.rs
```diff
@@ -31,5 +31,25 @@ error_chain! {
...
             display("{} is not implemented for the {} platform", functionality, platform)
         }
+
+        MissingFrontStart {
+            description("Malformed jekyll document, missing frontmatter start")
+            display("Malformed jekyll document, missing frontmatter start")
+        }
+
+        MissingFrontmatter {
+            description("Malformed jekyll document, missing frontmatter")
```
epage: If these errors are specific to Jekyll, they should probably have that in their name.

src/jekyll/jk_document.rs
```diff
@@ -0,0 +1,219 @@
...
+
+impl From<JkFrontmatterBuilder> for frontmatter::FrontmatterBuilder {
+    fn from(jk_front: JkFrontmatterBuilder) -> Self {
+        // Convert jekyll frontmatter into frontmatter (with `custom`)
+        let mut custom_attributes = jk_front.0;
+        frontmatter::FrontmatterBuilder::new()
+            .merge_title(custom_attributes.remove("title").and_then(|v| {
+                v.as_str().map(|s| s.to_owned())
+            }))
+            .merge_slug(custom_attributes.remove("permalink").and_then(|v| {
```
epage: I don't think these fields map

src/jekyll/jk_document.rs
```diff
@@ -0,0 +1,219 @@
...
+    }
+}
+
+pub fn convert_from_jk(source: &Path, dest: &Path) -> Result<()> {
+    if dest.is_file() {
+        Err(ErrorKind::CantOutputInFile.into())
+    } else if source.is_file() {
+        JkDocument::convert(source, dest)
+    } else if source.is_dir() {
+        for file in source.read_dir()? {
```
epage: Should this be recursive?

src/jekyll/jk_document.rs
```diff
@@ -0,0 +1,219 @@
...
+        dest.write_all(converted.as_bytes())?;
+        Ok(())
+    }
+}
+
+pub fn convert_from_jk(source: &Path, dest: &Path) -> Result<()> {
+    if dest.is_file() {
+        Err(ErrorKind::CantOutputInFile.into())
+    } else if source.is_file() {
+        JkDocument::convert(source, dest)
```
epage: Should this check the file type and convert according to it?
- template: Use `JkDocument`
- asset, copy
- jekyll site configuration, convert that

src/jekyll/jk_document.rs
```diff
@@ -0,0 +1,219 @@
...
+        // Convert jekyll frontmatter into frontmatter (with `custom`)
+        let mut custom_attributes = jk_front.0;
+        frontmatter::FrontmatterBuilder::new()
+            .merge_title(custom_attributes.remove("title").and_then(|v| {
+                v.as_str().map(|s| s.to_owned())
+            }))
+            .merge_slug(custom_attributes.remove("permalink").and_then(|v| {
+                v.as_str().map(|s| s.to_owned())
+            }))
+            .merge_description(custom_attributes.remove("excerpt").and_then(|v| {
```
epage: Jekyll doesn't have a description?  I thought I came across it when digging in the code

src/jekyll/jk_document.rs
```diff
@@ -0,0 +1,219 @@
...
+                v.as_str().map(|s| s.to_owned())
+            }))
+            .merge_categories(custom_attributes.remove("categories").and_then(|v| {
+                v.as_array().map(
+                    |v| v.iter().map(|v| v.to_string()).collect(),
+                )
+            }))
+            .merge_permalink(custom_attributes.remove("permalink").and_then(|v| {
```
epage: Does the permalink format need converting?

Even if it doesn't need it now, we should ensure there are tests for the variable substitution because it might change in the future and we need to make sure we know to update this code in that case.

src/jekyll/jk_document.rs
```diff
@@ -0,0 +1,219 @@
...
+        let doc = res.unwrap();
+        assert_eq!(doc.content.unwrap(), correct_content);
+        assert_eq!(doc.front.unwrap(), correct_front);
+    }
+
+    #[test]
+    fn parse_string_no_front() {
+        let res = JkDocument::parse_string(get_correct_content().to_owned());
+        assert!(res.is_err());
+        //assert_eq!(res.unwrap_err(), ErrorKind::MissingFrontmatter);
```
epage: Please resolve

src/jekyll/jk_document.rs
```diff
@@ -0,0 +1,219 @@
...
+
+    #[test]
+    fn parse_string_no_front_starter() {
+        let correct_front = get_correct_jk_front();
+        let correct_content = get_correct_content();
+        let correct_doc = format!("{}---\n{}", correct_front, correct_content);
+        let res = JkDocument::parse_string(correct_doc);
+
+        assert!(res.is_err());
+        //assert_eq!(res.unwrap_err(), ErrorKind::MissingFrontStart);
```
epage: Please resolve

src/jekyll/jk_document.rs
```diff
@@ -0,0 +1,219 @@
...
+    }
+}
+
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    // can't have custom: the order of fields is not stable
+    fn get_correct_cb_front() -> &'static str {
```
epage: Why make these functions rather than just static strings?

src/jekyll/jk_document.rs
```diff
@@ -0,0 +1,219 @@
...
+        Err(ErrorKind::InternalError.into())
+    }
+}
+
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    // can't have custom: the order of fields is not stable
```
epage: Alternatively, you could load the cobalt frontmatter and then do an equality check on that object.

The annoying part is it won't tell you which field failed the equality check.  String diffing is better in that regard.

Hmmm, mixed.  For the majority case, its probably ok.  We probably do want a custom test though.  Maybe you can parse the jk frontmatter and just check inside the custom to see if the field exists and has the right value.  We have other tests verifying the whole front matter conversion, we can get away with a single tests just verifying the one field under test.

src/document.rs
Geobert: Ok I'll move `read_file` and see if a version of `split` for jekyll is a lot different or not :)

src/error.rs
Geobert: The description is clumsy, I think this error can happen for cobalt's document too. May I change the description to make it more generic?

src/jekyll/jk_document.rs
Geobert: `title`? I agree for `slug` I need to dig more the link you gave me

src/jekyll/jk_document.rs
Geobert: I couldn't find it, maybe I oversight it, will triple check.

src/jekyll/jk_document.rs
Geobert: Sorry, I didn't understand what you're meaning here :-/

src/jekyll/jk_document.rs
Geobert: I don't know if under `_posts` folder, jekyll can have subfolder, if yes, then yeah, it should :)

src/jekyll/jk_document.rs
Geobert: uh… I don't know, I think I got carried away in a "functional programming" mood ^^'

Geobert:
The goal with this module is to allow me to export my WP blog into jekyll format using a plugin or exitwp, and generate my blog with Cobalt instead (including theming) so yeah, big job but I'll do it piece by piece, switching contribution field to not get burn by this (I want embedded videos, comments, tags…)

src/jekyll/jk_document.rs
Geobert: Didn't cross any `description` field in Jekyll's documentation.  On the other hand, I don't see an `excerpt` field in cobalt, so I made this mapping

src/error.rs
epage: In #203, we decided to support missing front matters, so the error only applies to Jekyll.

src/jekyll/jk_document.rs
epage: Was referring to slug.

src/jekyll/jk_document.rs
epage: The change unconditionally runs files through this conversion process and erroring if they don't match up.

Some blogs will have assets, like image files.  We shouldn't attempt this process on them.

My guess is Jekyll uses a whitelist of file extensions with some being added via the [`markdown_ext`](https://jekyllrb.com/docs/configuration/#default-configuration) field.

In addition, some people will want their [configuration](https://jekyllrb.com/docs/configuration/) converted.

src/jekyll/jk_document.rs
epage: I think it can.

epage:
>> One interesting trade off we'll need to make is between a solution that works for your use case and a full Jekyll converter. If you're up for the latter great, but I don't necessarily want to hold you to it; thats too much of a barrier to contributing. So if there are items you want to push back on, let's talk about a way for us to ensure there is some mechanism to draw attention to not-implemented features so someone can catch that when they are converting and report it (or fix it themselves).

> The goal with this module is to allow me to export my WP blog into jekyll format using a plugin or exitwp, and generate my blog with Cobalt instead (including theming) so yeah, big job but I'll do it piece by piece, switching contribution field to not get burn by this (I want embedded videos, comments, tags…)

Thanks for the effort.  It is a big task, but like I said, it doesn't have to be all tackled at once.  We can get a minimal set in with enough notification to people of missing features and then add the rest over time.  People might even see this feature, find it is missing a case they need, and join in making it great!

src/jekyll/jk_document.rs
epage: Also, there are pages not in the `_posts` folder.

You can find an example of [`WalkDir`](https://github.com/cobalt-org/cobalt.rs/blob/master/src/files.rs#L45) to handle the recursive walking for you.

src/error.rs
Geobert: Ok, I'll move it to jk_document then. By curiosity: what's the title if no front is present?

src/jekyll/jk_document.rs
Geobert: Oh I see, this struct only manage post at the moment, maybe I should rename it accordingly and when we add more file type, create a struct for them too?

Geobert:
Yeah, I'll do the minimum set I need to convert my Blog from WP->Jekyll->cobalt and doing so, entering in the fabulous world of statically generated blog! :D

src/jekyll/jk_document.rs
Geobert: After reading "Custom Path" on http://cobalt-org.github.io/docs/pages.html it seems that the plan was to copy Jekyll behavior, so no conversion needed.

src/error.rs
epage: We turn the slug into a title
https://github.com/cobalt-org/cobalt.rs/blob/master/src/frontmatter.rs#L283

Otherwise we fallback to a default
https://github.com/cobalt-org/cobalt.rs/blob/master/src/frontmatter.rs#L321

src/jekyll/jk_document.rs
epage: Oh well.  Can always be modified later if we find out more.

src/jekyll/jk_document.rs
epage: At first, I thought this project was meant to be a Jekyll clone.  Later I found out its more Jekyll inspired.  I've not yet updated that page.  I have made some minor tweaks to #198 to reflect that.

One example change I want to make: if people think its reasonable, I'd like to move away from `:var` substitutions.  The grammar is ambiguous (when is the end of a `var`?).  Currently, we handle this by enumerating all the supported vars and doing string substitutions.  We don't have the opportunity to detect unsupported vars.

One idea I had was to replace the system with Liquid.

src/jekyll/jk_document.rs
epage: I think the name works.  We could just add `JkConfig` for that.  For assets, we might just do a straight file copy.

At minimum, I'd appreciate it if the code would only apply `JkDocument` to the right type of files and `warn!` when it comes to unsupported files.

src/error.rs
Geobert: Are you sure about the default fallback? `ok_or_else` returns a `Result` and with the `?` if no title, we generate an `Error` with the message "No title", aren't we?

src/jekyll/jk_document.rs
Geobert: ok, I kinda understand the concept here, my knowledge on Jekyll and Cobalt is not advanced enough to fully grasp the issue but I can add a function call in the conversion here so if some conversions are needed, we already have the entry point to do so

src/jekyll/jk_document.rs
Geobert: Yup I understand better what you said, I'm processing only if extension is `md` or `markdown` for now. A JkConfig is a good idea as a second step, totally agree.

Geobert:
the only issue left is the recursive parsing of `_posts` but I'll add this later, I want to have a min set to migrate my blog before adding this, if this is ok with you :)

Geobert:
hm… didn't do the basic test: build a cobalt site with the converted file and it does not work :( 
`Caused by: data did not match any variant of untagged enum Value at line 1 column 5`
on this front:
```
path: "/:path/:filename:output_ext"
slug: "ghost-le-moteur-de-blog-qui-veut-detroner-wordpress"
title: "Ghost, le moteur de blog qui veut détrôner Wordpress"
description: ~
categories: 
  - Non classé
excerpt_separator: "\n\n"
published_date: ~
format: Raw
layout: post
is_draft: false
custom: 
  comments: true
  wordpress_id: 3551
  link: "https://haurchefant.fr/2017/09/20/ghost-le-moteur-de-blog-qui-veut-detroner-wordpress/"
  author: geob
  tags: 
    - geekage
---
```

CHANGELOG.md
```diff
@@ -1,5 +1,27 @@
-## 0.7.2 (2017-07-04)
+<a name="0.7.3"></a>
```
epage: FYI something went wrong when you tried to rebase.  Its counting all of the existing commits as new.

epage:
> hm… didn't do the basic test: build a cobalt site with the converted file and it does not work :(
Caused by: data did not match any variant of untagged enum Value at line 1 column 5
on this front:

Interesting, sounds like its failing to find a `Value` variant that is appropriate

Looking at the variants, I can't imagine what it is:
https://github.com/cobalt-org/liquid-rust/blob/master/src/value.rs#L9

Could you do some elimination to find out which field is causing problems?

epage:
> the only issue left is the recursive parsing of _posts but I'll add this later, I want to have a min set to migrate my blog before adding this, if this is ok with you :)

You might have seen that I've added notes to #276 for unhandled items.

.gitignore
```diff
@@ -7,3 +7,6 @@
 .*.swp
 /.vscode
 
+/.idea
+
+*.iml
```
epage: What are these files that are getting ignored?

src/jekyll/jk_document.rs
```diff
@@ -0,0 +1,285 @@
...
+pub struct JkDocument {
+    pub front: Option<String>,
+    pub content: Option<String>,
+}
+
+impl JkDocument {
+    pub fn parse_string(doc: String) -> Result<JkDocument> {
+        let (front, content) = split_document(&doc)?;
+        if front != None {
+            Err(ErrorKind::MissingFrontStart.into())
```
epage: `ok_or_else` will be a help here

```rust
let (front, content) = split_document(&doc)?;
let front = front.ok_or_else(|| Err(ErrorKind::MissingFrontStart.into())?;
let (front, content) = split_document(content)?;
let front = front.ok_or_else(|| Err(ErrorKind::MissingFrontStart.into())?;
Ok(JkDocument {front: Some(front.to_owned()), content: Some(content.to_owned()),})
```

See also https://rustbyexample.com/error/multiple_error_types/define_error_type.html

src/jekyll/jk_document.rs
```diff
@@ -0,0 +1,285 @@
...
+                                                    d.as_str().and_then(datetime::DateTime::parse)
+                                                }))
+            .merge_custom(custom_attributes)
+    }
+}
+
+#[derive(Debug, Default, Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub struct JkDocument {
+    pub front: Option<String>,
+    pub content: Option<String>,
```
epage: If one or both of these are required, why bother putting them in an `Option`?

src/jekyll/jk_document.rs
```diff
@@ -0,0 +1,285 @@
...
+    pub fn convert_front(front: Option<String>) -> Result<String> {
+        let front_value: JkFrontmatterBuilder = front
+            .map(|s| serde_yaml::from_str(&s))
+            .map_or(Ok(None), |r| r.map(Some))?
+            .unwrap_or_else(JkFrontmatterBuilder::new);
+
+        let front_builder: FrontmatterBuilder = front_value.into();
+        let front = front_builder.build()?;
+        let mut converted = serde_yaml::to_string(&front)?;
+        converted.drain(..4);
```
epage: Whats the `drain` doing?

src/jekyll/jk_document.rs
```diff
@@ -0,0 +1,285 @@
...
+    } else if source.is_file() {
+        JkDocument::convert(source, dest)
+    } else if source.is_dir() {
+        for file in source.read_dir()? {
+            if let Ok(file) = file {
+                let file_path = file.path();
+                let ext = file_path.extension().unwrap_or(OsStr::new(""));
+                if ext == "md" || ext == "markdown" {
+                    JkDocument::convert(&file.path(), dest)?
+                }
```
epage: Will `read_dir` return sub-directories?  Should we also watch for those and warn on them?

src/jekyll/jk_document.rs
```diff
@@ -0,0 +1,285 @@
...
+}
+
+fn read_file<P: AsRef<Path>>(path: P) -> Result<String> {
+    let mut file = File::open(path.as_ref())?;
+    let mut text = String::new();
+    file.read_to_string(&mut text)?;
+    Ok(text)
+}
+
+fn split_document(content: &str) -> Result<(Option<&str>, &str)> {
```
epage: Ideally this function would implement all the error handling policy and double-split you need.

src/jekyll/jk_errors.rs
```diff
@@ -0,0 +1,43 @@
+#![cfg_attr(feature = "cargo-clippy", allow(redundant_closure))]
```
epage: How come this is needed?

src/jekyll/jk_document.rs
```diff
@@ -0,0 +1,285 @@
...
+        Ok(())
+    }
+}
+
+pub fn convert_from_jk(source: &Path, dest: &Path) -> Result<()> {
+    if dest.is_file() {
+        Err(ErrorKind::CantOutputInFile.into())
+    } else if source.is_file() {
+        JkDocument::convert(source, dest)
+    } else if source.is_dir() {
```
epage: Should this create a `.cobalt.yml` (from `new.rs`) if we don't end up converting one?

epage:
So besides the things I just pointed out, I think the only thing left is warning on unsupported content (unsupported file types, subdirectories, etc) so people know what they need to hand-convert.

CHANGELOG.md
Geobert: that's why I deleted and recreated my fork. I'll create a new PR when I'm ready, this one will be cancelled I think

.gitignore
Geobert: These are file created by the IDE I'm using: IntelliJ

src/jekyll/jk_document.rs
Geobert: remove from self the range and return it: https://doc.rust-lang.org/std/string/struct.String.html#method.drain

src/jekyll/jk_document.rs
Geobert: At this point, I don't know, this command was really just to convert jekyll's post into cobalt's

src/jekyll/jk_errors.rs
Geobert: Just blindly copy pasted from `error.rs`

.gitignore
epage: Ok, seems reasonable.   Maybe add a comment?

src/jekyll/jk_document.rs
epage: Let me clarify, what is it you're trying to remove?

src/jekyll/jk_document.rs
epage: Thats fine, we can worry about it later.  

btw if its not clear, some of my questions are not saying "this needs to happen" but to make sure some implicit assumptions become explicit.

CHANGELOG.md
epage: Ok.  Sorry you've run into these problems.  Too bad our time zones are so far off (technically at work right now) otherwise we could get on gitter at the same time to try to work through them to make things easier (I presume).

epage:
This has been a learning process for you and I don't want to take away from that, I also don't want you burned out.  If there are parts of getting this PR in that you'd like me to wrap up, let me know.

Geobert:
Very nice of you :) I was on the verge of burn out, I admit, cumulated with bad/hard days at work atm, but don't worry, I'm not giving up :) I'm working on my next PR for this issue, applying all your good comments at a slower pace but it's progressing.

I want to migrate my blog to a static web site generator, and I want to learn Rust, so Cobalt is perfect for that :)

src/jekyll/jk_document.rs
Geobert: I'm removing the first "---" from Jekyll's front as it seems that Cobalt's one does not start with "---"

epage: Ah, makes sense.  Thanks for clarifying that.

Might be good to have a comment here to explain intent.  Maybe even put it in a function?  

```rust
fn dump_frontmatter(var: FrontMatter) -> Result<String> {
    let mut converted = serde_yaml::to_string(&front)?;
    // Remove yaml's leading `---`
    converted.drain(..4);
    Ok(converted)
}
```

epage:
See #294

PR #292, opened 2017-10-09
[Adding support for data files

### LINKED ISSUE #256: Support data files
Data files are ways to define custom attributes for use in pages in a centralized place.

## Use cases

- Defining menus (see #89 and #91)

## Data files in Hugo

- Lives in `./data`
- Supports `.yml`, `.yaml`, `.json` or `.toml`
- Accessed by `.Site.Data.path.relative.to.data.filename`

See https://gohugo.io/extras/datafiles/

## Data files in Jekyll

- Lives in `./_data`
- Supports `.yml`, `.yaml`, `.json` or `.csv` (with header row)
- Accessed by `site.data.path.relative.to.data.filename`

gnunicorn:
Sharing with you my progress on supporting data files

 - [x] Add tests
  - [x] nested sub directories
  - [x] top level lists
 - [x] Add support for `yml`/`yaml` files
 - [x] Add support for `json` files
 - [x] Add support for `toml` files
 - [x] Make `_data`-dir configurable
 - [x] replace usage of `unwrap` with `?`
 - [x] error on duplicate date names
 - [x] create top-level `site`-variable

 - ~~add support for CSV (I have a in-progress-stash about that)~~

fixes #256


src/cobalt.rs
```diff
@@ -120,6 +124,57 @@ pub fn build(config: &Config) -> Result<()> {
...
+                let file_name = String::from(file_stem.to_str().unwrap());
+                let full_path = data_root.join(df.clone());
+                let data : Value;
+
+                if ext == OsStr::new("yml") || ext == OsStr::new("yaml")  {
+                    let reader = File::open(full_path)?;
+                    data = serde_yaml::from_reader(reader)?;
+                } else if ext == OsStr::new("json") {
+                    let reader = File::open(full_path)?;
+                    data = serde_json::from_reader(reader).unwrap();
```
gnunicorn: Though the serde_json examples show they are using `?`, if I replace the unwrap there, I get:
```
error[E0277]: the trait bound `error::Error: std::convert::From<serde_json::Error>` is not satisfied
   --> src/cobalt.rs:146:28
    |
146 |                     data = serde_json::from_reader(reader)?;
    |                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::convert::From<serde_json::Error>` is not implemented for `error::Error`

```

Any ideas or hints why that might be?

Same for the toml version

src/cobalt.rs
```diff
@@ -120,6 +124,57 @@ pub fn build(config: &Config) -> Result<()> {
...
+                } else if ext == OsStr::new("toml") {
+                    let mut reader = File::open(full_path)?;
+                    let mut text = String::new();
+                    reader.read_to_string(&mut text)?;
+                    data = toml::from_str(&text).unwrap();
+                } else {
+                    continue
```
epage: Probably should log or error on unsupported file types

src/cobalt.rs
```diff
@@ -120,6 +124,57 @@ pub fn build(config: &Config) -> Result<()> {
...
+                }
+
+                // now find the nested map it is supposed to be in
+                let mut target_map = if let Some(path) = df.parent() {
+                    let mut map = &mut data_map;
+                    for part in path.to_str().unwrap().split("/") {
```
epage: How come you aren't using `Path.components` or `Path.iter`? for splitting the path and iterating on it?

src/cobalt.rs
```diff
@@ -177,6 +232,7 @@ pub fn build(config: &Config) -> Result<()> {
         }
 
         let mut context = post.get_render_context(&simple_posts_data);
+        context.set_val("data", Value::Object(data_map.clone()));
```
epage: So this puts `data` in the root object.  What about following what hugo and jekyll do and call it `site.data`?

src/cobalt.rs
```diff
@@ -120,6 +124,57 @@ pub fn build(config: &Config) -> Result<()> {
...
+
+    for df in data_files.files() {
+
+        if let Some(ext) = df.extension() {
+            if let Some(file_stem) = df.file_stem() {
+                let file_name = String::from(file_stem.to_str().unwrap());
+                let full_path = data_root.join(df.clone());
+                let data : Value;
+
+                if ext == OsStr::new("yml") || ext == OsStr::new("yaml")  {
```
epage: > add support for CSV (I have a in-progress-stash about that)

If there is nothing contentious in the design (like configuring the type of csv it reads) and trivial, let's do it.

src/cobalt.rs
```diff
@@ -120,6 +124,57 @@ pub fn build(config: &Config) -> Result<()> {
...
+                            .entry(String::from(part))
+                            .or_insert_with(|| Value::Object(HashMap::new()))
+                            .as_object_mut().unwrap();
+                    }
+                    map
+                } else {
+                    &mut data_map
+                };
+
+                target_map.insert(file_name, data);
```
epage: > merging data at the same point rather than overwrite it

I assume the only way we can overwrite is if 
- A file stem matches a directory name which contains more data
- Two files have the same stem but different extensions

Ignoring or overwriting will be surprising to people.  We should merge or error.  I lean towards erroring.

src/cobalt.rs
epage: I believe that means we need to chain our error type to their error types so that error-chain will implement `From` for us.

See https://github.com/cobalt-org/cobalt.rs/blob/master/src/error.rs

Speaking of, I recently added some `chain_err` to our code to call out what file a process failed on.  I'd recommend applying it here.

See https://github.com/cobalt-org/cobalt.rs/commit/1e07708badd73523892e4ab7a7c17c40d090bae2#diff-1adad72e4e1dbe7019b775a6acb9455aR182

src/cobalt.rs
gnunicorn: Fine with me. Was actually surprised there was no `site`-object yet (or at least I couldn't find it), but didn't want to add it, because this comes with a bunch of other things attached to it(on [jekyll at least](https://jekyllrb.com/docs/variables/#site-variables) and at least `site.time` I am using ;) )), unrelated to this feature.

Does this `site` already exist? Otherwise we might want to open another compatibility-issue for it?

src/cobalt.rs
gnunicorn: That is exactly the problem, having a folder with `_data/titles/*yaml` and `_data/titles.yaml` or any other combination of supported files. Overwriting isn't my preference, was just easier to implement 😀. But I agree erroring is better. Will implement that.

src/cobalt.rs
gnunicorn: Because I didn't know about them. Will take a look and learn about them. Thanks for the hint!

src/cobalt.rs
gnunicorn: Will log.

src/cobalt.rs
gnunicorn: After playing a bit more with it, the CSV reader is quite a bit more annoying to use (no direct serde-support like all the others here have) and makes the code base a lot more complex at the moment..

Considering that I haven't ever seen this being used in production, I'm tended to skip it for this PR and add open a ticket about it afterwards...

gnunicorn:
> Sorry, this slipped through the cracks

No worries, we all have a life outside of computering :) .

src/cobalt.rs
epage: Ok

src/cobalt.rs
epage: `site` does not already exist.  I was considering adding it but when you mentioned working on this, I decided to wait to not run into merge conflicts.

src/cobalt.rs
gnunicorn: I was actually wrong. [`HashMap.insert` ignores the new value, if the key already exists](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.insert). But using that I made a version that prints an error message to the user that it was skipped. I opted for not hard-error on that...

epage: > If the map did have this key present, the value is updated, and the old value is returned. The key is not updated,

Sounds like it doesn't ignore the new value, only the new key is ignored.  For example, if your key type is a case insensitive string then you can have two keys with different binary representations that are equal.  If you insert a `("key", "A")` followed by `("kEy", "B")`, then the map should contain `("key", "B")`.

gnunicorn: Oh, true, I misread that. Guess I have to make an explicit check before adding it then... meeeh... not cool.

src/cobalt.rs
epage: Is there a reason someone might want to have an unsupported file in here?

If so, we should leave it as a trace.  If not, we should make it a warning.

src/cobalt.rs
```diff
@@ -155,27 +155,36 @@ pub fn build(config: &Config) -> Result<()> {
...
-                            .as_object_mut().unwrap();
+                        if let Some(key) = part.to_str() {
+                            let mut cur_map = map;
+                            map = cur_map
+                                .entry(String::from(key))
+                                .or_insert_with(|| Value::Object(HashMap::new()))
+                                .as_object_mut().unwrap();
+                        } else {
+                            trace!("Skipping loading of data in {:?}: can't convert path to utf-8", path);
+                            break
```
epage: Won't this cause us to insert the data in a the wrong section because of the break?

Maybe the data loading should be moved into a separate function.  At that point, it can return early from this function, bypassing the insert below.

gnunicorn: True.

I've made it its own function and also decided to move to explict exection-stopping errors in these cases to force the user to clean up their data..

src/cobalt.rs
```diff
@@ -20,6 +24,45 @@ use config::SassOutputStyle;
...
+            if part.len() == 0 {
+                continue
+            }
+            if let Some(key) = part.to_str() {
+                let cur_map = map;
+                map = cur_map
+                    .entry(String::from(key))
+                    .or_insert_with(|| Value::Object(HashMap::new()))
+                    .as_object_mut()
+                    .unwrap();
```
epage: Note: this is the main item of concern.

We should probably error rather than unwrap in case there are overlapping data files with different data types.

`_data/data.json`:
```js
{ "key": "value" }
```
`_data/data/key.json`:
```js
{ "a": "b" }
```

src/cobalt.rs
```diff
@@ -20,6 +24,45 @@ use config::SassOutputStyle;
...
+               data: Value)
+               -> Result<()> {
+    // now find the nested map it is supposed to be in
+    let target_map = if let Some(path) = file_path.parent() {
+        let mut map = data_map;
+        for part in path.iter() {
+            if part.len() == 0 {
+                continue
+            }
+            if let Some(key) = part.to_str() {
```
epage: You could possibly clean this up:
```rust
let key = part.to_str().ok_or_else(format!("The data from {:?} can't be loaded as it contains non utf-8 characters", path))?;
```

src/cobalt.rs
```diff
@@ -125,6 +168,43 @@ pub fn build(config: &Config) -> Result<()> {
...
 
+    // load data files
+    let mut data_map: HashMap<String, Value> = HashMap::new();
+    let data_root = source.join(&config.data);
+    let data_files_builder = FilesBuilder::new(data_root.as_path())?;
+    let data_files = data_files_builder.build()?;
+
+    for df in data_files.files() {
+
+        if let Some(ext) = df.extension() {
```
epage: You could possibly clean this up
```rust
let ext = df.extension().unwrap_or("");
```

src/cobalt.rs
```diff
@@ -125,6 +168,43 @@ pub fn build(config: &Config) -> Result<()> {
...
+            if let Some(file_stem) = df.file_stem() {
```
epage: Is there ever a time we'll get an [empty file_stem](https://doc.rust-lang.org/std/path/struct.Path.html#method.file_stem) from `.files()`?

Could we clean this up by doing
```rust
let file_stem = df.file_stem().expect("Files will always return with a stem");
```

src/cobalt.rs
```diff
@@ -230,6 +313,10 @@ pub fn build(config: &Config) -> Result<()> {
         }
 
         let mut context = doc.get_render_context(&posts_data);
+        let mut site = HashMap::new();
+        site.insert("data".to_owned(), Value::Object(data_map.clone()));
```
epage: Is there a reason to create `site` twice?

We'll be adding more information to `site` over time and it'd be nice to not duplicate the logic.

src/cobalt.rs
```diff
@@ -20,6 +24,45 @@ use config::SassOutputStyle;
...
+fn deep_insert(data_map: &mut HashMap<String, Value>,
+               file_path: &Path,
+               target_key: String,
+               data: Value)
+               -> Result<()> {
+    // now find the nested map it is supposed to be in
+    let target_map = if let Some(path) = file_path.parent() {
+        let mut map = data_map;
+        for part in path.iter() {
+            if part.len() == 0 {
```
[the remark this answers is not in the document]
gnunicorn: This was a problem in an earlier version, with split, with top-level items. But you are right, this isn't a problem since the introduction of `path.iter()` anymore. removed.

src/cobalt.rs
gnunicorn: no, not necessary.

epage:
Looks good!  Once you get the CI happy, could you clean up the commits and post that its ready to merge?

epage:
Just think, I'll be improving the rustfmt situation once everyone's PRs are in...

gnunicorn:
> Just think, I'll be improving the rustfmt situation once everyone's PRs are in...


Yeah, just having it on latest would be useful. Even that same version gives me different complains here locally, so I always have to wait for the CI to fix them afterwards ...

Anwho... next trial..

gnunicorn:
it green @epage .

src/legacy/wildwest.rs
```diff
@@ -138,6 +139,7 @@ impl Default for GlobalConfig {
             dest: "./".to_owned(),
             layouts: "_layouts".to_owned(),
             drafts: "_drafts".to_owned(),
+            data: "_data".to_owned(),
```
epage: Something I didn't think about: is there a use case for this to be configurable?

For example, something I want to do is eventually remove the option for configuring `layouts` because I can't think of a justification for it.

I'm looking at tagging a release and changing this before then means I don't have to break compatibility.

gnunicorn: For data (unlike sass) I could see the point being made, that different tools have different names so this could ease import/testing it with an existing repo. But I am also fine with removing this, if you think this causes confusion.

PR #294, opened 2017-10-13
second take on convert Jekyll's post to Cobalt

Geobert:
Because of conflict mess, I preferred to delete my fork and redo a PR for #276 
#289 can be discard

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,10 @@ ignore = "0.2"
 serde = "1.0"
 serde_yaml = "0.7"
 
+[dependencies.sass-rs]
+version = "0.2.0"
+optional = true
+
 [dependencies.hyper]
 version = "0.10"
 default-features = false
@@ -63,4 +67,4 @@ unstable = []
 dev = []
 
 syntax-highlight = ["syntect"]
-
+sass = ['sass-rs']
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,9 +34,13 @@ script:
   - cargo -V
   - cargo check --verbose
   - cargo check --verbose --features "syntax-highlight"
+  - cargo check --verbose --features "sass"
+  - cargo check --verbose --features "syntax-highlight,sass"
   - cargo test  --verbose --features "syntax-highlight"
+  - cargo test  --verbose --features "sass"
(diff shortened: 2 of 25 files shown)


src/bin/cobalt/main.rs
```diff
@@ -227,7 +227,23 @@ fn run() -> Result<()> {
                                  .default_value("cobalt site import")
                                  .takes_value(true)))
         .subcommand(SubCommand::with_name("list-syntax-themes").about("list available themes"))
-        .subcommand(SubCommand::with_name("list-syntaxes").about("list supported syntaxes"));
+        .subcommand(SubCommand::with_name("list-syntaxes").about("list supported syntaxes"))
+        .subcommand(SubCommand::with_name("convert-jekyll")
+                        .about("convert jekyll website to cobalt")
+                        .arg(Arg::with_name("source")
```
epage: My guess is that the CI is failing because there is a global `source`/`destination` and a local one.

Not sure why you aren't running into it.

Cargo.toml
```diff
@@ -42,6 +42,10 @@ ignore = "0.2"
 serde = "1.0"
 serde_yaml = "0.7"
 
+[dependencies.sass-rs]
```
epage: Looks like you aren't fully rebased against master.  Want me to take care of that when the review is done?

Geobert: I need to learn git more, never done PR before :-/ How do I rebase my fork/master to upstream/master and then sync it to fork/dev to verify everything's fine, and then merge to fork/master in order to make a PR?

src/bin/cobalt/main.rs
Geobert: I've runned clippy but nothing came out :-/ I'll change the name of the option

Geobert: Do I need to change the long/short names as well? Or only the `with_name` one?

Cargo.toml
Geobert: And it seems my git is messing around line feeds, everytime I'm running rustfmt all the files looked liked modified but nothing has changed

Geobert:
Canceled

PR #295, opened 2017-10-14
Third take to convert jekyll's posts to cobalt

Geobert:
Had to delete my fork and start again. I don't know how to work with git and forks

### DIFF
--- a/src/jekyll/jk_document.rs
+++ b/src/jekyll/jk_document.rs
@@ -1,0 +1,290 @@
+use liquid;
+
+use super::super::frontmatter;
+use super::super::datetime;
+use std::fs::{create_dir_all, File};
+use std::io::{Read, Write};
+use std::path::Path;
+use frontmatter::FrontmatterBuilder;
+use super::jk_errors::{ErrorKind, Result};
+use serde_yaml;
+use regex::Regex;
+use slug::slugify;
+use std::ffi::OsStr;
+
+lazy_static! {
+    static ref FRONT_MATTER_DIVIDE: Regex = Regex::new(r"---\s*\r?\n").unwrap();
+}
+
+#[derive(Debug, Eq, PartialEq, Default, Clone, Serialize, Deserialize)]
+pub struct JkFrontmatterBuilder(liquid::Object);
+
+impl JkFrontmatterBuilder {
+    pub fn new() -> JkFrontmatterBuilder {
+        JkFrontmatterBuilder(liquid::Object::new())
+    }
+}
+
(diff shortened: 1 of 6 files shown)


epage:
My workflow
```
git clone my repo
git remote add upstream https://github.com/cobalt-org/cobalt.rs.git
git checkout -b myfeature  # master is reserved for tracking upstream
... work
# Prepare to push to github
git checkout master
git pull upstream master
git checkout myfeature
git rebase master  # usually comes out clean.  Resolving conflicts can be a bit confusing
git push
```

epage:
You renamed the key to access the value by but did not rename the flag.

btw when you do this, do you mind also squashing your commits so we have a cleaner history.

```
git rebase -i HEAD~5
```
will bring up your selected editor (WARNING: default is vim) with the last 5 commits.  You edit what action is to be performed (switch pick to squash or fix), save, quite, and your history is edited.

epage:
Thanks for your work on this!

src/jekyll/jk_document.rs
```diff
@@ -0,0 +1,290 @@
...
+
+fn split_document(content: &str) -> Result<(Option<&str>, &str)> {
+    if FRONT_MATTER_DIVIDE.is_match(content) {
+        let mut splits = FRONT_MATTER_DIVIDE.splitn(content, 3);
+        let first = splits.next().unwrap_or("");
+        let second = splits.next().unwrap_or("");
+        let third = splits.next().unwrap_or("");
+
+        if third.is_empty() {
+            // only one "---"
```
epage: Is there a reason you ended up supporting this use case?  Does jekyll actually support it?

Geobert: I don't know ^^' I'm fairly new to the world of static web site generator. I wanted to be robust here. I'll try to find some info and maybe find out that an error is more appropriate :)

epage: Thanks.

Geobert: If I believe Jekyll's documentation on frontmatter https://jekyllrb.com/docs/frontmatter/

> The front matter is where Jekyll starts to get really cool. Any file that contains a YAML front matter block will be processed by Jekyll as a special file. The front matter must be the first thing in the file and must take the form of valid YAML set between triple-dashed lines.

So it seems that I don't need to manage this but spit an error instead :)

epage: Already made the change in #298 as well as fix some bugs I forgot about during the review.

PR #296, opened 2017-10-15
Add example of using :slug in post_path to README

wezm:


epage:
Eventually I need to switch the README to more of a quick start and move a lot of the docs into the website but this should be fine for now.

PR #302, opened 2017-10-21
fix(data,sass): Remove custom folder support

epage:
This is an attempt to try to simplify things until someone has a use
case for it.

What are the problems with custom folders?
- Less confusing to users.  When giving the user a choicem, we should be
  able to describe why they might want to make that choice.
- Less to test
- No semantic problems to worry about.
  - What should be the behavior for a non-`_`-prefixed folder
  - What should be the behavior for deeply nesting particularly with the
    above question mixed in at random points.

Note: It is easier to remove the feature now and add it back later than to remove it later.

See also #278 and #292


epage:
FYI @gnunicorn

PR #308, opened 2017-10-25
Define path config behaviors

### LINKED ISSUE #284: Define path config behaviors
As of right now
- `.cobalt.yml` is read from CWD, defaulted if not found
- `source` is relative to CWD
- `dest` is relative to CWD
- `layouts` is relative to `source`
- `sass_dir` is relative to `source` (see #278)
- `posts` is relative to `source`

Considerations
- Allow `cobalt` to be run in any child directory of a site
  - like `git` and similar commands
- Separate `source` from data used to render source
  - data files (#256)
  - layouts
  - sass (#278)
  - etc

Proposed
- `.cobalt.yml` is searched for, starting in CWD and up through its parents, defaulted if not found
- `source` is relative to `.cobalt.yml`
- `dest` is relative to `.cobalt.yml`
- `layouts` is relative to `source` and not configurable
- `sass_dir` is relative to `source` (see #278) and not configurable
- `data_dir` is relative to `source` and not configurable
- `posts` is relative to `source`

Note: This change should include tests to cover the usage.  See #184 for test ideas.  The hacks for running single-threaded shouldn't be needed though because we'd no longer be relying on CWD (minus the finding of `.cobalt.yml` which could be tested using `assert_cli` instead).

epage:
- Make subdirs relative to config file, not cwd, to avoid unexpected behavior
  - `--destination` is unique, it will be relative to CWD unlike `dest` in the config file
- Only implement cli flags on relevant subcommands

Plus many more changes.

Fixes #284


epage:
CI failure was for a worker that didn't come up.  We still have sufficient coverage for its case.

PR #314, opened 2017-10-27
Provide Information when serde parsing fails

gnunicorn:
When we encounter an error reading a data-file (through serde not being willing to read it) or when merging it into the existing data set, it now names the file that is causing the error in the error message.

### DIFF
--- a/src/site.rs
+++ b/src/site.rs
@@ -165,11 +165,14 @@ fn insert_data_dir(data: &mut liquid::Object, data_root: &path::Path) -> Result<
             .file_stem()
             .expect("Files will always return with a stem");
         let file_stem = String::from(file_stem.to_str().unwrap());
+        let data_fragment = load_data(&full_path)
+            .chain_err(|| format!("Loading data from {:?} failed", full_path))?;
 
-        let data_fragment = load_data(&full_path)?;
-
-        deep_insert(data, rel_path, file_stem, data_fragment)?;
+        deep_insert(data, rel_path, file_stem, data_fragment)
+            .chain_err(|| format!("Merging data into {:?} failed", rel_path))?;
     }
+
+    trace!("Done loading data directory.");
 
     Ok(())
 }


src/site.rs
```diff
@@ -165,11 +165,13 @@ fn insert_data_dir(data: &mut liquid::Object, data_root: &path::Path) -> Result<
             .file_stem()
             .expect("Files will always return with a stem");
         let file_stem = String::from(file_stem.to_str().unwrap());
-
+        trace!("Loading data from {:?}", &full_path);
         let data_fragment = load_data(&full_path)?;
 
         deep_insert(data, rel_path, file_stem, data_fragment)?;
```
epage: Could you also add `chain_err` t both `load_data` and `deep_insert` calls to mention the file, something like

```rust
let data_fragment = load_data(&full_path).chain_err(|| format!("Failed to load data file {:?}", rel_path");
...
deep_insert(data, rel_path, file_stem, data_fragment).chain_err(|| format!("Failed to merging data from {:?}", rel_path");
```

This seems more important than the traces because
- it makes the extra information relevant to the exact use case
- People get it when the failure happens rather than having to do a second run.

gnunicorn:
Changing the approach: instead of using traces, we'll do `.chain_err` to provide better feedback when error occurs immediately.

src/site.rs
```diff
@@ -165,11 +165,14 @@ fn insert_data_dir(data: &mut liquid::Object, data_root: &path::Path) -> Result<
...
+            .chain_err(|| format!("Loading data from {:?} failed", full_path))?;
 
-        let data_fragment = load_data(&full_path)?;
-
-        deep_insert(data, rel_path, file_stem, data_fragment)?;
+        deep_insert(data, rel_path, file_stem, data_fragment)
+            .chain_err(|| format!("Merging data into {:?} failed", rel_path))?;
     }
 
+    trace!("Done loading data directory.");
```
gnunicorn: I left this trace in as I still find it useful to know that parsing the data succeeded and we are moving on to the next thing (and that is why the build fails).

epage: Interesting, I normally prefer the opposite: traces when each section starts rather than ends :)

Issue #241, opened 2017-06-16
Tests need to be refactored; failures from master with various EOL formats.


bddenhartog:
I'm on Fedora 25 (64-bit), using kernel `v4.11.3-202`.

I am unable to successfully run `cargo test` from `master` of this repository, using `rustc 1.17.0 (56124baa99 2017-04-24)` and `cargo 0.18.0 (fe7b0cdcf 2017-04-24)`. For sanity checking, I have attempted to clone the repository with the git configuration `core.autocrlf` set to its various values:

```
$ git clone https://github.com/cobalt-org/cobalt.rs.git && cd cobalt.rs && cargo test ; cd ../
// fails

$ rm -rf ./cobalt.rs && git -c core.autocrlf=true clone https://github.com/cobalt-org/cobalt.rs.git && cd cobalt.rs && cargo test ; cd ../
//fails

$ rm -rf ./cobalt.rs && git -c core.autocrlf=input clone https://github.com/cobalt-org/cobalt.rs.git && cd cobalt.rs && cargo test ; cd ../
//fails

$ rm -rf ./cobalt.rs && git -c core.autocrlf=false clone https://github.com/cobalt-org/cobalt.rs.git && cd cobalt.rs && cargo test ; cd ../
//fails
```

This issue was originally surfaced in #209 under Windows 10, so the issue does not seem to be limited to my machine.

epage:
Oye.  Thanks for looking into this further.

Geobert:
On my end, `git -c core.autocrlf=false clone https://github.com/cobalt-org/cobalt.rs.git` gives me a clone with all the test passing

epage:
The thing that confuses me about this is that I don't think cobalt deals with line endings, it just preserves whatever is in the source file.   For the tests, this means the fixture and target would both be changed to CRLF and should work.

When we validate a test run, we generate a diff using a specific EOL.  I thought this was just advisory only.  Either way, it should treat the two files the same.

epage:
Either way, I'm fine with adding a `.gitattribute` file and and editorconfig file to better specify how to handle this.

Ideally, the editorconfig file would also say to not strip trailing whitespace.  That is an easy way for people to mess up the tests.

Geobert:
I've normalize EOL in assert_dirs_eq and it solves many tests but some are still broken. While investigating these, I've notice a weirdness in the produced file: We have a mix of EOL:
[...]

PR #332, opened 2017-11-15
Restore ability to allow warnings in "dev" feature + "Native" newline_style for rustfmt

Geobert:


Cargo.lock
```diff
@@ -1,36 +1,3 @@
-[root]
```
epage: Any idea why the lock file changed?  Wondering if we should bother including it in this PR

src/lib.rs
```diff
@@ -39,7 +39,7 @@
         unused_parens,
         unused_unsafe,
         unused_variables,
-        while_true)]
+        while_true))]
 // This list is select `allow` warnings
 #![deny(trivial_casts,
```
epage: does it also need to be checked on this one?

src/lib.rs
```diff
@@ -4,7 +4,7 @@
 // 1. Run `rustc -W help`
 // 2. Grab all `default=warn` warnings
 // 3. Paste them here, deleting `warnings`, and any with `deprecated` in the name
-#![deny(const_err,
+#![cfg_attr(not(feature="dev"), deny(const_err,
```
epage: - If we're fixing it here, should we also fix it in the bin?
- Should the "fix" be to get it working again or remove it completely?  How does this impact your workflow?

.rustfmt.toml
```diff
@@ -8,7 +8,7 @@ fn_call_width = 60
 struct_lit_width = 18
 struct_variant_width = 35
 force_explicit_abi = true
-newline_style = "Unix"
+newline_style = "Native"
```
epage: Whats the motivation for changing this?

My biggest concern is people getting in the habit of working with `Native` and completely botching the tests.  It was a big weird to work through previously why someone's tests started all failing.  The person then opened #241 though I don't know if there is a good solution to it.

Geobert: For people using git with autocrlf set to true, it will avoid rustfmt to change all the file on the project :)

Cargo.lock
Geobert: Oh I though I've excluded it. I don't know why it was edited, basically a section moved, but nothing changed

src/lib.rs
Geobert: Maybe, I don't know why they are separated ^^'

src/lib.rs
Geobert: oh yeah, I missed out the bin.
It wasn't really impacting me, I've noticed this by chance and as it was supposed to be ignored in feature "dev" I made it works again :)

src/lib.rs
Geobert: Should I merge them?

.rustfmt.toml
epage: The hard thing is not everyone sets that flag.  I don't when I'm on Windows.

I suspect there is no right answer here.

src/lib.rs
epage: For me, its a matter of whatever is easiest for when we next need to update this list of warnings.  I did a dump of the warnings and put them in.  I don't remember the details, if the `allow`-defaulted ones were separate or not.

src/lib.rs
epage: RE fixing this

I guess my question is more broad, of if anyone cares about the `dev` feature and if not, if we should just remove it.

epage:
Sorry for my delayed responses!  My book is almost done ... but then I'll be out camping for Thanksgiving.  Soon I'll be returning to be more regular.

.rustfmt.toml
Geobert: Yeah :-/ or force a choice with .gitattributes as we spoke on gitter

Geobert:
no worries :D have fun :D

.rustfmt.toml
Geobert: https://www.git-scm.com/docs/gitattributes

src/lib.rs
Geobert: I'm using it sometimes when moving stuff around and experimenting, this phase can left a lot of unused import or result not used but only temporarily.

epage: Thanks for explaining your case.

I guess I'm fine living with the task master of warnings during my experimentation but I shouldn't force that on others :)

.rustfmt.toml
epage: So we have two options
- Keep it on `Unix` and expect people to override the file locally when using autocrlf
- Use editorconfig / gitattributes on the code as well as the tests to force unix style.

The first approach might be less than ideal because for someone being "sloppy" and doing `git -a`, they'll constantly be having to cut that change out of their commits.

Thoughts?

Geobert: I think gitattributes is safer for us

Geobert: After work on my branch https://github.com/Geobert/cobalt.rs/tree/%23241 this `Native` is not needed anymore, thanks to the `.gitattributes` file

Geobert:
I think I'm gonna close this PR and redo it with only `dev` allowing warnings, the code here is obsolete and merging will be a mess

PR #333, opened 2017-11-16
Travis deploy integration

badboy:
After experiencing problems with the old documented way of manually doing the import and push I simply switched over to Travis' native method and it works without a hassle.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,13 +231,13 @@ before_script:
 script:
   - cobalt build
 
-after_success: |
-  [ $TRAVIS_BRANCH = master ] &&
-  [ $TRAVIS_PULL_REQUEST = false ] &&
-  cobalt import &&
-  git config user.name "Cobalt Site Deployer" &&
-  git config user.email "name@example.com" &&
-  git push -fq https://${GH_TOKEN}@github.com/${TRAVIS_REPO_SLUG}.git gh-pages
+deploy:
+  provider: pages
+  skip_cleanup: true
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -231,16 +231,16 @@ before_script:
...
-  git config user.name "Cobalt Site Deployer" &&
-  git config user.email "name@example.com" &&
-  git push -fq https://${GH_TOKEN}@github.com/${TRAVIS_REPO_SLUG}.git gh-pages
+deploy:
+  provider: pages
+  skip_cleanup: true
+  github_token: $GH_TOKEN
+  local_dir: build
+  on:
+    branch: master
```
epage: So this will automatically push the results of `cobalt build` into the appropriate branch?

Have a link where I can learn  more?

badboy: Oh sorry. I put a link to the documentation into the commit message but forgot to include it in the PR text, here it is: https://docs.travis-ci.com/user/deployment/pages/

It pushed into `gh-pages` by default but this can be configured.

epage:
Thanks!  This is much better for our users.

In general, I need to revamp the documentation.  That is planned for [0.10](https://github.com/cobalt-org/cobalt.rs/milestone/7).

btw in 0.10 I'm looking to break compatibility in major ways but provide a upgrade utility.  My current plan is to test against the example sites.  Feel free to let me know of your site if you'd want me testing it too.

epage:
I've now created #334 to track my test cases.

PR #335, opened 2017-11-19
Normalize EOL before diff

Geobert:
This fixes most of the tests, see https://github.com/cobalt-org/cobalt.rs/issues/241 for the remaining failing tests

### DIFF
--- a/tests/mod.rs
+++ b/tests/mod.rs
@@ -51,6 +51,8 @@ fn assert_dirs_eq(expected: &Path, actual: &Path) {
             .expect("Comparison error")
             .read_to_string(&mut created)
             .expect("Could not read to string");
+        original = original.trim().replace("\r\n", "\n");
+        created = created.trim().replace("\r\n", "\n");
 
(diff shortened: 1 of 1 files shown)


epage:
So you found that nearly all the tests fail with autocrlf enabled.

The only reason I can think of this happening is we are somehow inserting our own newlines into the content.  I think this is a problem in of itself and we should investigate and fix that.

In that case, this change could be masking a problem.  Once that is addressed, should this go in?  I'm mixed.
- This could mask future problems with newlines.  
- How much do we care about people without autocrlf messing up the content of the tests?

I think my preference is still towarsds editorconfig and maybe gitattribute.

Geobert:
You're right, this normalization shouldn't be done on the test, but on the generated files

epage:
I'm sorry that solving this hasn't been as straightforward as this originally seemed :(

Geobert:
Don't worry, it's the correct decision ^^ That's why reviews are for ^^

PR #337, opened 2017-11-30
 feat(liquid): Dedicated _includes dir

### LINKED ISSUE #328: Define semantics for `include` directive
@gnunicorn pointed this out on gitter.

## Prior Art

Jekyll
- https://jekyllrb.com/docs/includes/
- `_includes` folder is flat (`include dir/menu.liquid` grabs `_includes/dir/menu.liquid`)

Hugo
- https://gohugo.io/templates/partials/
- `layouts/partials` is the root
- _layouts beat themes (consistent with hugo's theme override policy, see #301)
- no non-layout partial support

epage:
Previously, liquid `includes` were processed in the `source` dir.

Why change?
- Generally, snippets are kept in a central location, making this more
  cumbersome to access
- Jekyll and Hugo have a dedicated directory
- We want to support themes having includes and users overriding them,
  this means their needs to be a virtual file system of some sort
- Knowing what files are included allows us to cache off the data read
  from disk.  For now, this optimization has mixed results because
  `liquid` requires `LiquidOptions` to be cloned for every parse
  performed.

Ideally, we'd keep user snippets and layout snippets separate.  On disk,
that will happen once we add themes, so punting on it for now for
layouts (ad hoc themeing).  They are still within the same namespace in
the virtual filesystem though.  Punting on this for now.

This was hand tested to ensure deprecation warnings were issued and to
ensure retrofitted tests work.

Fixes #328

BREAKING CHANGE: Though we still load files from the old location (and
issue a deprecation warning), there is a very small chance that a file
already exists in the new location and we'll load that instead.


epage:
FYI @gnunicorn

Issue #305, opened 2017-10-21
CI fails when building a tag


epage:
Note: The binaries seem to still get uploaded.  This is mostly superficial (except for the scare on each release).

When pushing a tag, the CI will build binaries and upload them to github.  It does this for every target that builds `stable`.

We have a job dedicated to building rustfmt on stable.  This then fails.  See [example](https://travis-ci.org/cobalt-org/cobalt.rs/builds/290721243).

Requirements
- Should not start passing because it uploads a binary to github (that target is already handled)
- Should not start passing because we just moved rustfmt-nightly.  New rustfmt will be running on stable again soon.  If we want to move rustfmt on nightly just to avoid this, it should be recognized as such.

Options
- Add to the deploy condition
- Move rustfmt into the regular stable jobs
- Always run rustfmt on nightly even if it isn't needed

Geobert:
I did not understand the issue here (came here because you pointed this in the PR)

epage:
Whenever I roll out a new release (like 0.11.1), I tag the release using git.  This is what triggers the CI to deploy a binary to the [github releases page](https://github.com/cobalt-org/cobalt.rs/releases).

The problem is the `deploy` happens for every job that runs on "stable".  `rustfmt` runs on stable but is not meant to be deployed, causing a build error.  This is annoying because I get a build error and have to go check it to make sure something more catastrophic didn't fail.

[example](https://travis-ci.org/cobalt-org/cobalt.rs/jobs/327075801)

I think I linked to this when brainstorming an idea to speed up our builds.  I brought this up as a caution for how things can go wrong when splitting one job into two.

PR #352, opened 2018-01-02
fixes #180 : Enable syntax highlight to all platform

### LINKED ISSUE #180: Enable syntax highlighting by default on all platforms
This requires either forking or patching syntect to pass the `static-libonig` flag to rust-onig, AFAIU.

Geobert:


Cargo.toml
```diff
@@ -45,17 +45,17 @@ serde_json = "1.0"
...
 [dependencies.hyper]
 version = "0.10"
 default-features = false
 
 [dependencies.syntect]
-version = "1.8.0"
+version = "1.8"
 optional = true
-default-features = false
+default-features = true
```
epage: Is there a reason you are enabling the default features of `syntect`?

Cargo.toml
```diff
@@ -64,9 +64,9 @@ difference = "1.0"
 tempdir = "0.3"
 
 [features]
-default = []
-unstable = []
-dev = []
+default = ["syntax-highlight"]
+unstable = ["syntax-highlight"]
+dev = ["syntax-highlight"]
```
epage: Why specify it in `dev` and `unstable`?  `default` should be sufficient

Cargo.toml
```diff
@@ -64,9 +64,9 @@ difference = "1.0"
 tempdir = "0.3"
 
 [features]
-default = []
-unstable = []
-dev = []
+default = ["syntax-highlight"]
```
epage: The CI should do a test build with default features disabled.

Also, the CI should no longer explicitly call out `syntax-highlight` if its enabled by default.

Cargo.toml
Geobert: I admit it was kind of trials and errors here ^^' Not sure this one is needed.

epage:
Thanks for owning this!

Cargo.toml
Geobert: I didn't test with only `default`, I test now :)

Cargo.toml
Geobert: Oh, I forgot to edit the CI scripts, going right now :)

Cargo.toml
epage: Thanks

Cargo.toml
epage: Thanks

Cargo.toml
epage: Oh, missed the deploy script.  Thanks for getting that.  So just .travis.yml and appveyor.yml left?

Geobert: appveyor has no mention of the feature, I've changed travis :)

epage: > appveyor has no mention of the feature, I've changed travis :)

My concern is with us accidentally breaking the no-default build.

> The CI should do a test build with default features disabled.

Geobert: hm, I don't know much about appveyor yet, what do you mean?

Geobert:
Shall we keep the gnu target?

appveyor.yml
```diff
@@ -22,6 +22,7 @@ environment:
 install:
   - ps: >-
       $Env:PATH += ';C:\msys64\usr\bin'
+  - if "%TARGET%" == "x86_64-pc-windows-msvc" call "%VS140COMNTOOLS%\..\..\VC\vcvarsall.bat" x86_amd64
```
epage: You'll probably need to copy the rest of the setup for this from gutenberg
https://github.com/Keats/gutenberg/blob/master/appveyor.yml

Cargo.toml
epage: This is a Cargo feature and not an appveyor vs travis thing.

`syntax-highlight` is a default feature, meaning cargo will auto-add it to the build when calling `cargo build`.  Users can disable this though, with `--no-default-features`.  We should make sure we don't break the build when no features are enabled.

So its just a matter of adding to both CIs
`cargo test --no-default-features`.

appveyor.yml
Geobert: yeah, I'm trying to figure out what's wrong ^^ bear with me, it will be a git mess in the end x)

Cargo.toml
Geobert: I see! :D I will add this after fixing appveyor :)

Geobert:
I don't get why appveyor is failing, cargo test runs perfectly fine on my computer :(

epage:
It seems like you made it the furthest when you had `--target %TARGET` passed to cargo.

Geobert:
with `--target %TARGET` it does not work on my computer :-/ I'm trying the `os` line which is present in gutenberg

epage:
> with --target %TARGET it does not work on my computer :-/ 

How come `--target x86_64-pc-windows-msvc` fail on your computer?

Geobert:
Hm, it did failed but I think I forgot the call to the batch file… ^^' it works so I put back the --target

Geobert:
the `cli` tests are all failing, don't know why :(

epage:
My guess is that its something to do with the tempdir. 

I'd recommend
- Disabling travis within your branch (to be nicer to travis while you iterate)
- Moving down to one appveyor target for now (to be nicer to appveyor while you iterate)
- Try setting RUST_BACKTRACE to get more info.

epage:
> Shall we keep the gnu target in appveyor?

Forgot to respond to this

This goes back to #99 for which I have no clue how important that target is within the rust ecosystem.  Is there a reason you disabled it besides faster iteration?

Geobert:
how do we disable travis?

epage:
One option is to rename the file.

Geobert:
Ok I found what's going on: passing `--target` change the path of `cobalt.exe` from `target/debug/cobalt` to `target/x86_64-pc-windows-msvc/debug/cobalt`

I'm going to add an instruction in `appveyor.yml` to copy the exec file just under `target` and change the cli test to launch `target/cobalt` instead

Geobert:
Travis failed here because of https://travis-ci.org/cobalt-org/cobalt.rs/jobs/324590284#L578 (error while downloading rust)

It should work. How do we force a rebuild?

epage:
> Ok I found what's going on: passing --target change the path of cobalt.exe from target/debug/cobalt to target/x86_64-pc-windows-msvc/debug/cobalt

Alternatively, we could use `main_binary` as long as we don't need extra features enabled (which these tests shouldn't need)

https://docs.rs/assert_cli/0.5.4/assert_cli/struct.Assert.html#method.main_binary

appveyor.yml
```diff
@@ -9,29 +11,39 @@ environment:
 
   matrix:
     # Stable channel
-    - TARGET: x86_64-pc-windows-gnu
+    #- TARGET: x86_64-pc-windows-gnu
```
epage: Whats the status of re-enabling this?

epage:
Things remaining
- Consider using `main_binary`
- window-gnu?
- running tests with default features disabled

And then once all looks good on that, cleaning up commits.

appveyor.yml
Geobert: Is does not compile, failing on miniz

Geobert:
- What is main_binary?
- I'm for abandoning gnu target, what's the point if we support Windows through MSVC?
- oh yeah forgot about this, I'll tackle this tomorrow :)

epage:
> What is main_binary?

[main_binary](https://docs.rs/assert_cli/0.5.4/assert_cli/struct.Assert.html#method.main_binary) is feature in `assert_cli` that will find the binary for you.

> I'm for abandoning gnu target, what's the point if we support Windows through MSVC?

*grumble* Ok.

Geobert:
replacing all `command` by `main_binary` and moving the parameters into `with_args(&[…])` makes two tests to fail, only 2 >_< : 
```
failures:

---- init_project_can_build stdout ----
        thread 'init_project_can_build' panicked at 'CLI assertion failed: (command `cargo run --bin cobalt -- init --trace` expected to succeed)
status=failed
stdout=``````
stderr=```error: could not find `Cargo.toml` in `C:\Users\Geob\AppData\Local\Temp\init.T2gAKVXg42pr` or any parent directory
```', C:\Users\Geob\.cargo\registry\src\github.com-1ecc6299db9ec823\assert_cli-0.5.4\src\assert.rs:368:12
stack backtrace:

---- log_levels_silent stdout ----
        thread 'log_levels_silent' panicked at 'CLI assertion failed: `cargo run -- build --silent -d C:\Users\Geob\AppData\Local\Temp\silent.AVccq2wIlRDv` StdErr mismatch: diff:

+Blocking waiting for file lock on build directory
    Finished dev [unoptimized + debuginfo] target(s) in 0.1 secs
     Running `I:\Dev\cobalt.rs\target\debug\cobalt.exe build --silent -d C:\Users\Geob\AppData\Local\Temp\silent.AVccq2wIlRDv`
', C:\Users\Geob\.cargo\registry\src\github.com-1ecc6299db9ec823\assert_cli-0.5.4\src\assert.rs:368:12
stack backtrace:
   0:     0x7ff6aefe95b1 - std::sys_common::backtrace::_print
                               at C:\projects\rust\src\libstd\sys_common\backtrace.rs:92
   1:     0x7ff6aeff5b2c - std::panicking::default_hook::{{closure}}
                               at C:\projects\rust\src\libstd\panicking.rs:380
   2:     0x7ff6aeff57a7 - std::panicking::default_hook
                               at C:\projects\rust\src\libstd\panicking.rs:391
   3:     0x7ff6aeff6195 - std::panicking::rust_panic_with_hook
                               at C:\projects\rust\src\libstd\panicking.rs:577
   4:     0x7ff6aeff6028 - std::panicking::begin_panic<alloc::string::String>
                               at C:\projects\rust\src\libstd\panicking.rs:538
   5:     0x7ff6aeff5f01 - std::panicking::begin_panic_fmt
                               at C:\projects\rust\src\libstd\panicking.rs:522
   6:     0x7ff6aef2db43 - assert_cli::assert::Assert::unwrap
                               at C:\Users\Geob\.cargo\registry\src\github.com-1ecc6299db9ec823\assert_cli-0.5.4\src\assert.rs:368
   7:     0x7ff6aef22f7a - cli::log_levels_silent
                               at I:\Dev\cobalt.rs\tests\cli.rs:138
   8:     0x7ff6aefa943d - test::{{impl}}::call_box<(),closure>
                               at C:\projects\rust\src\libtest\lib.rs:141
   9:     0x7ff6aeff74d1 - panic_unwind::__rust_maybe_catch_panic
                               at C:\projects\rust\src\libpanic_unwind\lib.rs:99
  10:     0x7ff6aef97342 - std::sys_common::backtrace::__rust_begin_short_backtrace<closure,()>
                               at C:\projects\rust\src\libstd\sys_common\backtrace.rs:134
  11:     0x7ff6aef98687 - std::panicking::try::do_call<std::panic::AssertUnwindSafe<closure>,()>
                               at C:\projects\rust\src\libstd\panicking.rs:476
  12:     0x7ff6aeff74d1 - panic_unwind::__rust_maybe_catch_panic
                               at C:\projects\rust\src\libpanic_unwind\lib.rs:99
  13:     0x7ff6aefa22d2 - alloc::boxed::{{impl}}::call_box<(),closure>
                               at C:\projects\rust\src\liballoc\boxed.rs:726
  14:     0x7ff6aeff4a5b - std::sys::imp::thread::{{impl}}::new::thread_start
                               at C:\projects\rust\src\libstd\sys\windows\thread.rs:54
  15:     0x7ffd978c8363 - BaseThreadInitThunk


failures:
    init_project_can_build
    log_levels_silent

test result: FAILED. 6 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out

error: test failed, to rerun pass '--test cli'
```
I don't get it, it should do the exact same thing.

Geobert:
here the gist: https://gist.github.com/Geobert/a989ea06281b0fad39ee2bc67269ad06

epage:
Huh, didn't realize `cargo run` will print out that text with `--quiet`.

See more info at the issue I created on assert_cli: https://github.com/killercup/assert_cli/issues/79

epage:
Ok, so that assert_cli issue will probably not be resolved in time to unblock this issue.

The main care abouts: people can easily run the tests on their own (no extra steps) and the CI provides value.

`main_binary` fails the former because it requires knowing that you need to pass a flag to limit the number of threads

The original solution fails because it requires copying the binary first, even outside of the `--target` case (see `.travis.yml`).

Another idea I have is that instead of `--target %TARGET%`, would `rustup run %CHANNEL%-%TARGET% cargo ...` work and not modify the path the binary is written to?

epage:
> Another idea I have is that instead of --target %TARGET%, would rustup run %CHANNEL%-%TARGET% cargo ... work and not modify the path the binary is written to?

I guess the shorthand would also work: `cargo +%CHANNEL%-%TARGET% ...`

epage:
> rustup-init.exe -y --default-host %TARGET% --default-toolchain %RUST_VERSION%

*sigh* why are we downloading stable (RUST_VERSION) rather than beta (CHANNEL)?

Looks like the latest version of trust gets rid of `CHANNEL` and only uses `RUST_VERSION`
See https://github.com/japaric/trust/blob/master/appveyor.yml

That should fix it.

epage:
So it looks like enabling syntect breaks `cargo check` on windows somehow?

At least (for my machine) it still works on Linux though the serialized builds of appveyor tend to make it the bottle neck.

Geobert:
I'll switch to CHANNEL, doesn't make sense otherwise. Wondering how it works before though…

as onig is relying on some C/C++ code, I think check does not support this yet.

appveyor.yml
```diff
@@ -1,6 +1,8 @@
 # Based on the "trust" template v0.1.1
 # https://github.com/japaric/trust/tree/v0.1.1
 
+os: Visual Studio 2015
+
 environment:
   global:
     RUST_VERSION: stable
```
epage: This is dead code now, please remove it

Geobert:
:o everything passes! need to clean up commits now

Geobert:
WTF, everything is fine but status is 1???

Geobert:
SIGKILL on a compilation? https://travis-ci.org/cobalt-org/cobalt.rs/jobs/325410957#L796

appveyor.yml
```diff
@@ -1,37 +1,49 @@
...
     APPVEYOR_CACHE_SKIP_RESTORE: true
 
   matrix:
     # Stable channel
-    - TARGET: x86_64-pc-windows-gnu
-      CHANNEL: stable
-    - TARGET: i686-pc-windows-msvc
-      CHANNEL: stable
+    #- TARGET: x86_64-pc-windows-gnu
+    #  CHANNEL: stable
```
epage: If we're not supporting gnu, we should remove it (Rather than leave it commented out) and give a reason in the commit message.

appveyor.yml
```diff
@@ -1,37 +1,49 @@
...
+    #  TOOLCHAIN: msvc
+    #  PLATFORM: i686
     # Beta channel
-    - TARGET: x86_64-pc-windows-msvc
-      CHANNEL: beta
+    #- TARGET: x86_64-pc-windows-msvc
+    #  CHANNEL: beta
+    #  TOOLCHAIN: msvc
+    #  PLATFORM: x86_64
```
epage: These targets are still commented out.

Geobert:
I don't get this error: https://ci.appveyor.com/project/johannhof/cobalt-rs/build/1.0.571/job/ptn77qeecg26q01i#L972

this happens time to time without apparent reason :-/

epage:
Whats weird is this doesn't fail as often on other builds.

Geobert:
https://ask.electric-cloud.com/questions/6390/link-fatal-error-lnk1318-unexpected-pdb-error.html

Geobert:
and now it on x86_64 instead on i686… (╯°□°）╯︵ ┻━┻

Geobert:
https://github.com/rust-lang/cargo/issues/4655
https://github.com/rust-lang/cargo/issues/3161

seems that we can't do anything about it? Maybe adding syntax-highlight makes the build consume more RAM?

appveyor.yml
```diff
@@ -29,6 +29,7 @@ install:
   - if "%TOOLCHAIN%" == "msvc" if "%PLATFORM%" == "x86_64" "C:\Program Files\Microsoft SDKs\Windows\v7.1\Bin\SetEnv.cmd" /x64
   - if "%TOOLCHAIN%" == "msvc" if "%PLATFORM%" == "x86_64" call "%VS140COMNTOOLS%\..\..\VC\vcvarsall.bat" x86_amd64
   - if "%TOOLCHAIN%" == "msys" set PATH=C:\msys64\mingw%MSYS_BITS%\bin;C:\msys64\usr\bin;%PATH%
+  - set MSBUILD_NODE_CONNECTION_TIMEOUT=0
```
epage: Should this instead be under `environment: global:`?

epage:
Now I'm trying to weigh out dropping 32-bit support.

The main use case of keeping it afaik is MS's new x86 on ARM stuff which is 32-bit only.  Unsure if we'd be able to cross compile to it at some point.

Geobert:
I agree to get rid of 32bits build, I'll try that asap, holidays are finished…

epage:
`i686-pc-windows-msvc`  builds are already broken for release builds  due to issues with GetLastError?
-One build back with 1.22.1, see https://ci.appveyor.com/project/johannhof/cobalt-rs/build/1.0.560
- Another build with 1.23, see https://ci.appveyor.com/project/johannhof/cobalt-rs/build/1.0.586

epage:
If the build succeeds, feel free to squash the commits, documenting in them why you removed what you did, and let me know so I can merge this.

Geobert:
https://ci.appveyor.com/project/tak1n/cobalt-rs/build/1.0.587/job/cw095bwrnbsdop0x#L976

I don't know what to do about this.

epage:
We're still getting sporadic failures?

This is not fun.

Geobert:
I've asked reddit for help.

appveyor.yml
```diff
@@ -1,37 +1,41 @@
...
   - rustc -Vv
   - cargo -V
+
+test_script:
   # we don't run the "test phase" when doing deploys
-  - if [%APPVEYOR_REPO_TAG%]==[false] ( cargo check --verbose )
-  - if [%APPVEYOR_REPO_TAG%]==[false] ( cargo test --verbose )
+  - if [%APPVEYOR_REPO_TAG%]==[false] ( cargo +%CHANNEL%-%TARGET% build )
+  - if [%APPVEYOR_REPO_TAG%]==[false] ( cargo +%CHANNEL%-%TARGET% test )
+  - if [%APPVEYOR_REPO_TAG%]==[false] ( cargo +%CHANNEL%-%TARGET% build --no-default-features )
```
epage: Do these still need to be `build` instead of `check`?

appveyor.yml
```diff
@@ -1,37 +1,41 @@
...
-      CHANNEL: stable
     - TARGET: x86_64-pc-windows-msvc
       CHANNEL: stable
+      TOOLCHAIN: msvc
+      PLATFORM: x86_64
     # Beta channel
     - TARGET: x86_64-pc-windows-msvc
       CHANNEL: beta
+      TOOLCHAIN: msvc
+      PLATFORM: x86_64
```
epage: What is `PLATFORM`?

Should we set it globally?

appveyor.yml
```diff
@@ -1,37 +1,41 @@
...
 # https://github.com/japaric/trust/tree/v0.1.1
 
+os: Visual Studio 2017
+
 environment:
   global:
-    RUST_VERSION: stable
     CRATE_NAME: cobalt
     APPVEYOR_CACHE_SKIP_RESTORE: true
+    MSBUILD_NODE_CONNECTION_TIMEOUT: 0
```
epage: What value does this give us?

.travis.yml
```diff
@@ -47,9 +47,11 @@ script:
   - rustc -Vv
   - cargo -V
   - cargo check --verbose
-  - cargo check --verbose --features "syntax-highlight,sass"
   - cargo test  --verbose
-  - cargo test  --verbose --features "syntax-highlight,sass"
+  - cargo check --no-default-features --verbose
+  - cargo test --no-default-features --verbose
+  - cargo check --verbose --features "sass"
+  - cargo test  --verbose --features "sass"
```
epage: At some point we should make it so different feature combinations run as different jobs so they can run in parallel rather than series.

Granted that can get messy, particularly when we already have problems with our rustfmt and clippy jobs (see #305)

appveyor.yml
Geobert: This was one of the workaround for the PDB issue, I'll try without it :)

appveyor.yml
Geobert: As said on gitter: because of natives C/C++ libs we need the build as `check` doesn't support them yet.

appveyor.yml
Geobert: Good catch, it was used in the `install` part to call the correct config `.bat` file according to i686 vs x86_64. To be deleted :)

.travis.yml
Geobert: Is this possible with travis?

epage: Yup, we just define more jobs in the matrix with a different variable set to control what happens in the job, maybe a `FEATURE` variable that is `--feature "sass"` vs `--no-default-features`.

Not something to include in this because it'd require a lot more experimenting and we shouldn't hold this up :)

appveyor.yml
epage: The annoying thing is that this doesn't seem to be a problem on Linux :(

epage:
Sweet! Squash the commits and I'll get this merged!

Thanks for all your effort on this one.  Its been long and frustrating.  I'm glad this didn't burn you out.  Side projects should reinvigorate us and not tear us down.

PR #367, opened 2018-01-18
#241

Geobert:
Attempt to fix all this EOL mess for the tests.
+ split of build() function into smaller pieces for readability


Geobert:
bear with me... I've just realised under the shower that I have submit this PR *without * compiling on my computer first... 😅 do not PR while being so tired... let me sort this out ^^' sorry about that

Geobert:
Why after a squash, I have this conflict???

Geobert:
It seems that squashing above the merging point lose the merge… git is kinda dark magic sometime

epage:
This is three changes in one, could you split this into distinct PRs to make it easier for me to review?

Geobert:
3? One is refactoring of build() and the other changes are about solving #241.
I'll split this asap (rush at work)

epage:
The third was fixing the `dev` feature.

epage:
(while the dev feature is small, that also means it can probably go in a lot faster)

Geobert:
Done :)

epage:
Thanks!  This will be a big help

epage:
Hey, I'm looking at documenting best practices with rustfmt, including issues with autocrlf.  Mind giving more background on the problems you had with it?

See https://github.com/crate-ci/crate-ci.github.io/issues/4

Geobert:
If I recall properly, if we have `autocrl=true`, then git checkout the files to the EOL format of the OS.
CRLF for Windows, will break every tests that involved excepts as the default excerpt separator is LF.

On rustfmt, it will change all the CRLF to LF, changing unchanged file EOL and so they will appear as modified even if we didn't touch them :)

epage:
Could you post that on the issue, so I don't have to act as a middleman? :)

PR #370, opened 2018-01-23
Fix EOL issue with tests

Geobert:
Normalizing everything cobalt read to LF and and outputing LF as well. 
excerpts_CRLF test has fixtures as CRLF, and target as LF

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# force LF on checkout to avoid rustfmt messing up the EOL
+* text eol=lf
--- a/tests/fixtures/excerpts_CRLF/.gitattributes
+++ b/tests/fixtures/excerpts_CRLF/.gitattributes
@@ -1,0 +1,2 @@
+# force tests to keep their Windows EOL
+* text eol=crlf
--- a/tests/mod.rs
+++ b/tests/mod.rs
@@ -239,6 +239,11 @@ pub fn excerpts() {
 }
 
 #[test]
+pub fn excerpts_crlf() {
+    run_test("excerpts_CRLF").unwrap();
+}
+
+#[test]
 pub fn posts_in_subfolder() {
     run_test("posts_in_subfolder").unwrap();
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -230,6 +230,7 @@ dependencies = [
  "lazy_static 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "liquid 0.13.7 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.3.9 (registry+https://github.com/rust-lang/crates.io-index)",
+ "normalize-line-endings 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "notify 4.0.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "pulldown-cmark 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "regex 0.2.5 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -657,6 +658,11 @@ dependencies = [
  "bitflags 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "libc 0.2.35 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
+
+[[package]]
+name = "normalize-line-endings"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "notify"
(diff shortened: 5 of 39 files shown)


src/document.rs
```diff
@@ -2,6 +2,7 @@ use std::collections::HashMap;
 use std::collections::hash_map::Entry;
 use std::default::Default;
 use std::path::{Path, PathBuf};
+use std::iter::FromIterator;
```
epage: (arbitrary place)
```
Fix EOL issue with tests

Normalizing everything cobalt read to LF and and outputing LF as well. 
excerpts_CRLF test has fixtures as CRLF, and target as LF



typo
```
Could you get this to conform to [conventional changelist format](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-angular), including calling out `Fixes #241`?

I didn't point it out in the other PRs because they were purely internal and unrelated to Issues.

.gitattributes
```diff
@@ -0,0 +1,2 @@
+* text eol=lf
+**/tests/fixtures/*_CRLF/** text eol=crlf
```
epage: Would it be better to have a separate `.gitattributes` file right there in the test?

Having it local like that seems like it'd help in terms of moving the code around, etc.

tests/mod.rs
```diff
@@ -50,6 +53,8 @@ fn assert_dirs_eq(expected: &Path, actual: &Path) {
             .read_to_string(&mut created)
             .expect("Could not read to string");
 
+        let created = String::from_iter(normalized(created.chars()));
```
epage: if `.gitattributes` is saying `lf` for everything, and the test should be generating `lf` based output, why is normalization needed here?

.gitattributes
```diff
@@ -0,0 +1,2 @@
+* text eol=lf
```
epage: Forgot to ask, why is this needed?  Is it just for the tests?  If so, should it just live in in `tests` and target the `targets` and `fixtures` folders?

epage: Also, should we be added an editorconfig file(s) to specify our line ending policies?

epage:
So let me see if I can summarize our long conversation from #241

- Issue is about getting tests to work regardless of someone's local line ending setting in git
- In the investigation, we found
  - cobalt's output could produce mixed line endings if the user uses Windows line endings because some sections are converting or inserting line endings (markdown, our markdown reference code)
  - cobalt can miss excerpts if the user uses Windows line endings and the default excerpt separator because it specifies a specific line ending policy

There are a mixture of solutions we can apply
- Normalize line endings in what we output
- Normalize line endings in our tests (see #335)
  - There was concern that this just masks behavior
- Detect the user's line endings and convert the output to match it
- Use `.gitattributes` and `.editorconfig` to encourage certain behavior within cobalt's tests (doesn't impact users)
- Encourage repos with users in different environments to use a non-default `excerpt_separator` like `<!--more-->`.
  - What percent of clients will have users in different environments?
- Change the default newline separator
  - The one nice thing about our current one is that it is more "natural"

Traits for us to weigh out in evaluating a solution
- Preserving the user's line endings
- Consistent line endings
- Performance
- UX Simplicity
- Implementation simplicity

My thoughts over time:
- Initially concerned about us unintentionally munging the line endings and thinking that needs to be fixed so we can preserve the user's data as-is
  - Can you tell I work in an industry where the number one rule is "don't corrupt data" (people might die)?
  - I'm also concerned about the users pushing cobalt in directions we never originally considered and needing precise control over their output
- Upon realizing some of it was out of our control (markdown), instead of trying to auto-detect or anything, wondering if we should just leave line endings as-is.
  - Most environments accept mixed line endings
  - It lets the user opt-out of things messing up their line endings, so they can have precise control over their output
- If we are to have any ambition for wide adoption, we need to "just work" which means we can't just tell windows users "Oh, go change the excerpt separator"

So while I'm still concerned about giving advanced users the power to do what they want, I think it is reasonable for this to go in now.  In the future, we can either re-evaluate and pull this out or add a config flag for people to opt-out.  Neither should be done at this moment though.

.gitattributes
Geobert: I didn't know we can have multiple `.gitattributes`, good idea!

.gitattributes
Geobert: This is to make sure that the project is checked out with Linux EOL in source code and test (with exception of the *_CRLF folders) so `cargo fmt` won't messed up the files being different

tests/mod.rs
Geobert: uh, good point, I… don't know, and without, it does not work so it seems I failed here… need more work :-/

Geobert: I forgot one normalization when reading the layout, it makes the situation better but my test still not pass :-/

Geobert: Found multiples occurences of `read_file` some might need normalization as well, I'll see that later

.gitattributes
epage: Oh right, the problem of aligning rustfmt with git.

This seems less obvious enough that it might deserve a comment in the file.

Geobert:
The route I choose is to normalize the EOL when reading files. 3 normalizations are needed (see diffs) and it make the tests pass :D
I've put .gitattributes in the excerpt_CRLF folder but I don't know how the pattern works: can I put *=crlf here?

tests/fixtures/excerpts_CRLF/.gitattributes
```diff
@@ -0,0 +1,2 @@
+# force tests to keep their Windows EOL
+* text eol=crlf
```
epage: Should there be a non-root editorconfig here?

tests/mod.rs
```diff
@@ -5,6 +5,7 @@ extern crate cobalt;
 extern crate error_chain;
 extern crate tempdir;
 extern crate walkdir;
+extern crate normalize_line_endings;
```
epage: Unused?

tests/mod.rs
```diff
@@ -54,16 +55,13 @@ fn assert_dirs_eq(expected: &Path, actual: &Path) {
...
         .filter_map(|e| e.ok())
         .filter(|e| e.file_type().is_file());
     for entry in walker {
-        let extra_file = entry
-            .path()
-            .strip_prefix(expected)
-            .expect("Comparison error");
-        let src_file = Path::new(actual).join(&extra_file);
+        let extra_file = entry.path().strip_prefix(result).expect("Comparison error");
+        let src_file = Path::new(expected).join(&extra_file);
```
epage: What was the intent of changing these variables around?

src/document.rs
```diff
@@ -194,6 +196,7 @@ impl Document {
                  -> Result<Document> {
         trace!("Parsing {:?}", rel_path);
         let content = files::read_file(src_path)?;
+        let content = String::from_iter(normalized(content.chars()));
```
epage: I'm sorry I didn't notice this before but ... what do you think of doing the normalization in `read_file`?

I could see it being nice that `read_file` is a general abstraction for loading data throughout cobalt.  If there are cases where we shouldn't normalize (like maybe `cobalt migrate`), then we just don't use it.

Geobert: … I didn't notice read_file was ours… ^^'

tests/fixtures/excerpts_CRLF/.gitattributes
Geobert: we can have multiple editorconfig as well? I'll them tomorrow :D

tests/mod.rs
Geobert: I found the original name confusing, I forgot to rollback them, it was for me reasoning, I'll rollback this ^^' sorry

.editorconfig
```diff
@@ -0,0 +1,3 @@
+[*]
+end_of_line = lf
+insert_final_newline = true
```
epage: This file should specify `root`

> root: special property that should be specified at the top of the file outside of any sections. Set to "true" to stop .editorconfig files search on current file.

http://editorconfig.org/

any nested ones shouldn't.

Geobert:
Once the CI are done, I think this is merge-ready :D

bddenhartog:
The tests `excerpts` and `excerpts_crlf` **still** fail for me with a fresh clone of the repository on Linux, on the master branch, with no `core.autocrlf` gitconfig setting.

PR #373, opened 2018-01-27
Rename some refactored fct

Geobert:
Here is the follow up of https://github.com/cobalt-org/cobalt.rs/pull/369 applying your comments!

### DIFF
--- a/src/cobalt.rs
+++ b/src/cobalt.rs
@@ -1,7 +1,7 @@
 use std::fs;
 use std::collections::HashMap;
 use std::io::Write;
-use std::path::Path;
+use std::path::{PathBuf, Path};
 use std::ffi::OsStr;
 use liquid;
 use rss;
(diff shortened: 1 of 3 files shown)


Geobert:
I'll wait for https://github.com/cobalt-org/cobalt.rs/pull/370 to merged, otherwise I have issues with tests

Geobert:
@epage : Changing `generate_posts` code to call the factorised version `generate_doc` makes excerpt generation fail.
Doesn't make any sense as I've compare the code, and it's the same. Can it be the cloning?

Geobert:
yeah, the cloning was not a good idea, this is done while rendering excerpt: https://github.com/cobalt-org/cobalt.rs/blob/master/src/document.rs#L312
hence the error.

epage:
Looks good.  let me know when its squashed.

PR #378, opened 2018-01-29
fix(serve): Fix site.base_url by adding http://

Geobert:


src/bin/cobalt/serve.rs
```diff
@@ -44,7 +44,7 @@ pub fn serve_command(matches: &clap::ArgMatches) -> Result<()> {
 
     let mut config = args::get_config(matches)?;
     debug!("Overriding config `site.base_url` with `{}`", ip);
-    config.site.base_url = Some(ip.clone());
+    config.site.base_url = Some(format!("http://{}", ip.clone()));
```
epage: Shouldn't need the `.clone()` anymore.

PR #391, opened 2018-03-05
add option to bind the serve to a different host

tglman:
As the title, this is useful when you are testing/coding your site somewhere else maybe on ssh.

Regard

### DIFF
--- a/src/bin/cobalt/serve.rs
+++ b/src/bin/cobalt/serve.rs
@@ -33,6 +33,14 @@ pub fn serve_command_args() -> clap::App<'static, 'static> {
                 .takes_value(true),
         )
         .arg(
+            clap::Arg::with_name("host")
+                .long("host")
+                .value_name("host-name/IP")
+                .help("Host to serve from")
+                .default_value("localhost")
+                .takes_value(true),
+        )
+        .arg(
             clap::Arg::with_name("no-watch")
                 .long("no-watch")
                 .help("Disable rebuilding on change")
@@ -42,8 +50,9 @@ pub fn serve_command_args() -> clap::App<'static, 'static> {
 }
 
 pub fn serve_command(matches: &clap::ArgMatches) -> Result<()> {
+    let host = matches.value_of("host").unwrap().to_string();
     let port = matches.value_of("port").unwrap().to_string();
-    let ip = format!("127.0.0.1:{}", port);
+    let ip = format!("{}:{}", host, port);
 
     let mut config = args::get_config(matches)?;
     debug!("Overriding config `site.base_url` with `{}`", ip);


src/bin/cobalt/serve.rs
```diff
@@ -30,6 +30,13 @@ pub fn serve_command_args() -> clap::App<'static, 'static> {
                  .help("Port to serve from")
                  .default_value("3000")
                  .takes_value(true))
+        .arg(clap::Arg::with_name("host")
+                 .short("H")
```
epage: Unsure whether there should be an abbreviation.

Cases when to have one
- The flag is used frequently enough that it is (1) easily assumed and (2) helping most people
- The abbreviation is a common convention

So far this seems like a less common case.  I don't have the familiarity with other applications to say there is a strong precedence for `-H`.  So I'm leaning against it.

Thoughts?

src/bin/cobalt/serve.rs
```diff
@@ -39,7 +46,8 @@ pub fn serve_command_args() -> clap::App<'static, 'static> {
 
 pub fn serve_command(matches: &clap::ArgMatches) -> Result<()> {
     let port = matches.value_of("port").unwrap().to_string();
-    let ip = format!("127.0.0.1:{}", port);
+    let host = matches.value_of("host").unwrap().to_string();
+    let ip = format!("{}:{}", host, port);
```
epage: Maybe its just me but it reads better here and above to have `host` before `port`.  Mind swapping the order?

epage:
> add option to bind the serve to a different host

Also, do you mind putting the commit message in [conventional style](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-angular)?  It'll make it easier for me to remember this when creating the changelog.

An example

```
feat(server): Allow binding to different host

This is useful when you are testing/coding your site somewhere else maybe on ssh.
```

src/bin/cobalt/serve.rs
tglman: yes sometime -H stands for human, i may use another (if any is good) or completely remove it.

tglman:
umm I may hacked too much with git and github is not able to merge it, let me know if you prefer a new PR.

Regards

epage:
Oh, sorry, I forgot there where outstanding PRs when I made those very ... disruptive changes.

It might be easiest re-implement it since its such a small change.  I leave it up to you on whether you want to reuse this PR or start a new one.  The former is doable but I don't want the darker corners of git to get in the way of you getting your changes in!

tglman:
Ok managed to merge :)

epage:
Do you feel comfortable squashing the commits?

Any questions or concerns about the formatting issues the CI is complaining about?

tglman:
Hi @epage,
what version of rust fmt do you use ? you know there where few breaking changes recently.

Regards

epage:
I upgraded cobalt to rustfmt from 1.24

See https://github.com/cobalt-org/cobalt.rs/blob/master/.travis.yml#L25

tglman:
Hi @epage,

Now should be fine, I'm still fighting with rustfmt, without rust up is kind of impossible to install the correct version.

Regards

epage:
Thanks!

> Now should be fine, I'm still fighting with rustfmt, without rust up is kind of impossible to install the correct version.

Huh, I assumed switching to the rustfmt from rustup would simplify things for people :(

tglman:
Hi, 
It's fine for the majority so it's a right change, the issue is actually understand what version rustup install and be able to install it manually, it's just me that i do not want to rely too much on .sh downloaded on the fly from Internet.

PR #402, opened 2018-03-19
 feat(new): Load default files from disk 

### LINKED ISSUE #355: Support overriding the default pages for `cobalt new`
`cobalt new <title>` creates a page.  it uses a built-in template based on the location.

We want to support users overriding the built-in template.

## Proposal

`cobalt new` will check for `<source>/_defaults/<collection_slug>.<template_ext>` and fallback to the builtin template if it doesn't exist

## Prior Art

epage:
`cobalt new` looks for a file that matches `_defaults/<collection>.<ext>` where:
- `collection` is the name of the collection (like `pages` or `posts`)
- `ext` is
  - from `--with-ext`
  - from `--file`
  - otherwise `md`

If the file is not found, a built in is used.

I dislike the name `_defaults` but couldn't come up with a better.  The
only alternative I had was `_new`.

`cobalt init` will also now create the `_defaults` folder to help guide
users to this new feature.

Fixes #355

### DIFF
--- a/src/bin/cobalt/new.rs
+++ b/src/bin/cobalt/new.rs
@@ -164,7 +164,7 @@ pub fn create_new_document(
     title: &str,
     file: path::PathBuf,
 ) -> Result<()> {
-    let file = if file.extension().is_none() {
+    let file = if file.extension().is_none() || file.is_dir() {
         let file_name = format!("{}.md", cobalt_model::slug::slugify(title));
         let mut file = file;
         file.push(path::Path::new(&file_name));
--- a/src/bin/cobalt/new.rs
+++ b/src/bin/cobalt/new.rs
@@ -1,3 +1,4 @@
+use std::collections;
 use std::env;
 use std::fs;
 use std::io::Write;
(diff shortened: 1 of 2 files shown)


src/bin/cobalt/new.rs
```diff
@@ -175,20 +206,52 @@ pub fn create_new_document(
...
+    } else {
+        bail!(
+            "Target file wouldn't be a member of any collection: {:?}",
+            file
+        );
+    };
+
+    let source_path = config
+        .source
+        .join(format!("_defaults/{}.{}", file_type, extension));
```
epage: As I said elsewhere, I hate this directory name.  Any other thoughts?

jimrybarski: I propose `_default_templates`. It's a bit long, but its purpose can be inferred without looking at any documentation.

Geobert: This is though, I like gutenberg's choice for it, a bit obscure at first, but archetype names very well what's inside this folder. It will teach a word to people ;)

epage: Gutenberg copied Hugo's use of archetype?  I couldn't find anything in the gutenberg docs related to this.

I forgot another option I thought of is "_prototypes" like the Prototype Pattern or Javascript Prototypes.  The word, at least among programmers, is more common than archetype and the connotation, at least to programmers, is slightly more accurate.

Geobert: Sorry, I made a mistake, I meant Hugo ^^'
_prototypes is not bad :)

PR #407, opened 2018-04-04
feat: add runtime configuration to built in syntax highlighting

### LINKED ISSUE #343: Is it possible to disable syntax highlighting in config?
Here's my use case. I'm using gitlab ci with the recommended cobalt container. It has syntax-highlighting feature turned on, but I was doing all the highlighting myself using highlight.js.  I want to keep using the same CI, but I don't want cobalt's default syntax highlighting.

Thanks!

z2oh:
Add an enabled option to the `syntax_highlight` category of `_cobalt.yml`. When set to `false`, cobalt will not perform syntax highlighting in Markdown files. The default is `true`, so this should be totally backwards compatible with the current functionality.

Do we also want to remove the compile time options for syntax-highlighting? If compiling with the feature adds extra dependencies, there are some potential benefits to keeping the compile time options in, however this could cause some confusion.

I'm not overly familiar with the architecture of this project, so feel free to let me know if there is a way you would rather me implement this or if there is some way to improve this PR.

Closes #343

### DIFF
--- a/src/cobalt_model/config.rs
+++ b/src/cobalt_model/config.rs
@@ -19,12 +19,14 @@ use super::template;
 #[serde(deny_unknown_fields, default)]
 pub struct SyntaxHighlight {
     pub theme: String,
+    pub enabled: bool,
 }
 
 impl Default for SyntaxHighlight {
     fn default() -> Self {
         Self {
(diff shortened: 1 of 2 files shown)


epage:
Thanks!

Issue #436, opened 2018-05-01
[Refactor] rename parse_pages to parse_docs


Geobert:
I'm trying to get knowledge of how cobalt works in order to implement pagination and noticed that `parse_pages` is not really well named (my bad, I was the one who made this function when splitting `build()`).

I think `parse_docs` will be a better name.

In the same idea, the field `pages` in `Collections` is may be not so well chosen, as `context.pages.pages` can show.

I propose to rename it `docs` as well.

Geobert:
We can even change the signature of `parse_pages` as the 1st param is a field of the second parameter

Geobert:
Also, optimization wise, maybe put the call to `Document::parse` inside [the condition](https://github.com/cobalt-org/cobalt.rs/blob/master/src/cobalt.rs#L296) as we keep the parse result only if this condition is verified

epage:
In some respects, pages and docs both work.  In parts of cobalt, I still call them documents for historical reasons but in the user-facing documentation, I refer to them as pages, even if they aren't in the `pages` collection.

epage:
Currently its presented on the site as

> Posts are special pages that live in posts. You can use a different directory for posts with the posts.dir in your _cobalt.yml.

https://cobalt-org.github.io/docs/posts.html

Creating a proposal for the website for terminology would probably be the first thing to do so we refactor the code to match what we're telling users.

Geobert:
Agree, I'll do a pass on the doc and propose a terminology here :)

Geobert:
I propose to change in the documentation:

- under existing `Content` section, create a page called `Documents` which says:

> A document can be a `page`, a  `post` or a `draft`.

then we put below the current `Pages` documentation and then the current `Posts` documentation (including the `Drafts` part)

- Move `Syntax highlighting` on its own page

So in the code we an rename the field `Collection::pages` to `Collection::documents` or `Collection::docs`.
And avoid this weird `pages.pages` and rename the function I'm talking about in the title :)

To optimise the processing a bit, we may pass to `parse_pages` `collection.include_drafts` so we can stop the parsing after the front being created if the flag `front.is_draft` is true.

Geobert:
Are breaking changes allowed? Because to be coherent, I need to rename how we access the collection documents: `collections.posts.pages` become `collections.posts.documents`

epage:
> I propose to change in the documentation:

Why is this here and not in the repo for documentation?

> Are breaking changes allowed? 

[...]

PR #437, opened 2018-05-04
refactor(collection): Rename Collection::pages to Collection::documents

Geobert:
and renamed parse_pages to parse_documents
for the [refactoring proposal](https://github.com/cobalt-org/cobalt.rs/issues/436)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,13 +27,13 @@ matrix:
     - cmake --version
     - cargo coverage || true
   - env: RUSTFMT
-    rust: 1.24.0  # `stable`: Locking down for consistent behavior
+    rust: 1.25.0  # `stable`: Locking down for consistent behavior
     install:
     - rustup component add rustfmt-preview
     script:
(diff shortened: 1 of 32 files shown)


Geobert:
Why I keep forgetting running these tests… I've run fmt test and clippy but not the more important one… sorry, working on it

Geobert:
It seems that `rustfmt` behaves differently between 1.24 and 1.25. I took the liberty to bump to the latest stable version

epage:
I would have recommended on waiting until the proposal is accepted in case it isn't.

Geobert:
It was not hard work so no bother if it's rejected. It made me understand better how cobalt works :)

epage:
> It was not hard work so no bother if it's rejected. It made me understand better how cobalt works :)

Glad you are looking at it like that :)

One of my big concerns is ensuring contributors don't burn out.  That is one of the reasons I encourage communication so much.

Geobert:
let's cancel this

PR #447, opened 2018-05-08
feat(sass): Enable by default

Geobert:
Trying by keeping VS2017 and bump Rust stable version

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,14 +13,16 @@ environment:
     - TARGET: x86_64-pc-windows-msvc
       CHANNEL: stable
       TOOLCHAIN: msvc
+      PLATFORM: x86_64
 
     # Beta channel
     - TARGET: x86_64-pc-windows-msvc
       CHANNEL: beta
       TOOLCHAIN: msvc
+      PLATFORM: x86_64
 
 install:
-  - if "%TOOLCHAIN%" == "msvc" call "C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvars64.bat"
+  - if "%TOOLCHAIN%" == "msvc" if "%PLATFORM%" == "x86_64" call "C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvarsall.bat" x86_amd64
   - if "%TOOLCHAIN%" == "msys" set PATH=C:\msys64\mingw%MSYS_BITS%\bin;C:\msys64\usr\bin;%PATH%
   - curl -sSf -o rustup-init.exe https://win.rustup.rs/
   - rustup-init.exe -y --default-host %TARGET% --default-toolchain %CHANNEL%
@@ -30,11 +32,11 @@ install:
 
 test_script:
   # we don't run the "test phase" when doing deploys
-  - if [%APPVEYOR_REPO_TAG%]==[false] ( cargo +%CHANNEL%-%TARGET% build )
-  - if [%APPVEYOR_REPO_TAG%]==[false] ( cargo +%CHANNEL%-%TARGET% test )
-  - if [%APPVEYOR_REPO_TAG%]==[false] ( cargo +%CHANNEL%-%TARGET% build --no-default-features )
-  - if [%APPVEYOR_REPO_TAG%]==[false] ( cargo +%CHANNEL%-%TARGET% test --no-default-features )
-
+  - if [%APPVEYOR_REPO_TAG%]==[false] ( cargo build --all )
+  - if [%APPVEYOR_REPO_TAG%]==[false] ( cargo test --all )
+  - if [%APPVEYOR_REPO_TAG%]==[false] ( cargo build --all --no-default-features )
+  - if [%APPVEYOR_REPO_TAG%]==[false] ( cargo test --all --no-default-features )
+  
 before_deploy:
   - cargo rustc --target %TARGET% --release --bin cobalt -- -C lto
   - ps: ci\before_deploy.ps1
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 /target
 /build
+/rls
 
 .DS_Store
 *.rs.bk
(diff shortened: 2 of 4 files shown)


Geobert:
Ok, I figured out for the compilation, we were missing PLATFORM flag and --all, don't really know why.

Test wise, cli test are failing if we do `static ref _BIN: PathBuf = CWD.join("target/debug/cobalt");` but works if we do `static ref _BIN: PathBuf = CWD.join("target\\debug\\cobalt");` which does not make any sense as Rust is supposed to abstract this for us. 

I tried to remove lazy_static to see if it's tampering somehow, but we need it.

Geobert:
And sometime, we have this error: https://ci.appveyor.com/project/johannhof/cobalt-rs/build/1.0.775

disable.travis.yml
```diff
@@ -27,22 +27,22 @@ matrix:
     - cmake --version
     - cargo coverage || true
   - env: RUSTFMT
-    rust: 1.24.0  # `stable`: Locking down for consistent behavior
+    rust: 1.25.0  # `stable`: Locking down for consistent behavior
```
epage: FYI If we move forward with this, I'd expect an explanation in the commit message and PR for why we need to bump our minimum version

appveyor.yml
```diff
@@ -30,10 +33,10 @@ install:
 
 test_script:
   # we don't run the "test phase" when doing deploys
-  - if [%APPVEYOR_REPO_TAG%]==[false] ( cargo +%CHANNEL%-%TARGET% build )
-  - if [%APPVEYOR_REPO_TAG%]==[false] ( cargo +%CHANNEL%-%TARGET% test )
+  - if [%APPVEYOR_REPO_TAG%]==[false] ( cargo +%CHANNEL%-%TARGET% build --all )
```
epage: So there were a lot of commits with a lot of failures.  Do you happen to remember what failure you were resolving by adding this?

appveyor.yml
```diff
@@ -13,14 +13,17 @@ environment:
     - TARGET: x86_64-pc-windows-msvc
       CHANNEL: stable
       TOOLCHAIN: msvc
+      PLATFORM: x86_64
 
     # Beta channel
     - TARGET: x86_64-pc-windows-msvc
       CHANNEL: beta
       TOOLCHAIN: msvc
+      PLATFORM: x86_64
```
epage: Yeah, in #398 I forgot to copy these over for some reason.

epage:
> Test wise, cli test are failing if we do static ref _BIN: PathBuf = CWD.join("target/debug/cobalt"); but works if we do static ref _BIN: PathBuf = CWD.join("target\\debug\\cobalt"); which does not make any sense as Rust is supposed to abstract this for us.

That seems very suspect.  Is this locally or on appveyor?  Do you have a appveyor log to point to?

> And sometime, we have this error: https://ci.appveyor.com/project/johannhof/cobalt-rs/build/1.0.775
I got it twice in my trials but the second time it made no sense at all since the change was only on cli.rs

Maybe there is something here to help
https://github.com/rust-lang/rust/issues/33145

disable.travis.yml
Geobert: Oh, I'm the kind of guy that always try to use the latest stable to benefit from code generation improvement and performance. As I'm on 1.25, I noticed that rustfmt behave differently. If we don't bump, I'll install 1.24 for cobalt locally :)

appveyor.yml
Geobert: I'll track back the errors, I forgot to keep a link -_-

disable.travis.yml
epage: Oh, right, this is just for locking down rustfmt for consistent output.  Similar for warnings.,  Ideally these would be upgraded in an independent PR.

Geobert:
ok using VS2015 we have this random errors, that' s why I've updated to VS2017 when activating syntec on all platform.

The issue is that sass-sys does not build with VS2017. I reported the issue: https://github.com/compass-rs/sass-rs/issues/22#issuecomment-387876161

epage:
Thanks for doing that investigation!

Geobert:
Another solution is to see if my PR on libsass is accepted: https://github.com/sass/libsass/pull/2648

epage:
> and it's true that gutenberg has many submodule compiled independantly :-/ Is that something we were planning to do?

I'm trying to avoid splitting up for the sake of compile times.

We already want to split jekyll out.  I think I want to split cobalt_model out but was waiting on further refactorings to better define what is or isn't in cobalt_model.

Geobert:
ok, let's hope my PR will make it

Geobert:
Another lead (long term) is this project: https://github.com/kaj/rsass

Geobert:
I'm trying different leads: https://github.com/compass-rs/sass-rs/pull/25

this can also make sass-rs build on VS2107

Geobert:
I'm waiting for the merge of https://github.com/compass-rs/sass-rs/pull/26 to update our dependency to sass-rs and we should be done with this at last!

epage:
> I'm waiting for the merge of compass-rs/sass-rs#26 to update our dependency to sass-rs and we should be done with this at last!

Awesome job owning this to completion!

Geobert:
My PR has been merged, waiting for a release with it now.
Can be delayed because of https://github.com/rust-lang/rust/issues/50778 though :-/

Geobert:
Testing how Rust 1.27 behaves on appveyor, I didn't see the error on my computer for a while

Geobert:
It works! 🎉 🎈

epage:
My one concern is if this will stay working :(

epage:
Now that we think we got this, could you clean up the history?

Geobert:
I'll do a loop test tomorrow on my computer first ^^

Geobert:
If I squash the merge, I have a conflict with `master` :-/ I think we can merge this, 1.27 may have solved the linker issue.

epage:
I'd rather not have merges in PRs.

Geobert:
How do I solve this?

epage:
Have you tried to rebase your branch against master locally?

Geobert:
yay! Thanks! :)

Geobert:
I'm not use to rebase yet, need to read a bit more on this…

epage:
Thanks!

PR #473, opened 2018-06-26
refactor(permalink): extract permalink parsing into its own module

Geobert:
To be used in pagination

### DIFF
--- a/src/cobalt_model/permalink.rs
+++ b/src/cobalt_model/permalink.rs
@@ -1,0 +1,99 @@
+use liquid;
+use std::path::{Path, PathBuf};
+
+use error::*;
+
+pub fn explode_permalink<S: AsRef<str>>(
+  permalink: S,
+  attributes: &liquid::Object,
+) -> Result<String> {
+  explode_permalink_string(permalink.as_ref(), attributes)
+}
+
+fn explode_permalink_string(permalink: &str, attributes: &liquid::Object) -> Result<String> {
+  lazy_static! {
+    static ref PERMALINK_PARSER: liquid::Parser = liquid::Parser::new();
+  }
+  let p = PERMALINK_PARSER.parse(permalink)?;
+  let mut p = p.render(attributes)?;
+
+  // Handle the user doing windows-style
+  p = p.replace("\\", "/");
+
+  // Handle cases where substutions were blank
+  p = p.replace("//", "/");
+
+  if p.starts_with('/') {
+    p.remove(0);
+  }
+
+  Ok(p)
+}
+
+pub fn format_url_as_file<S: AsRef<str>>(permalink: S) -> PathBuf {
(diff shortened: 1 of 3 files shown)


src/cobalt_model/permalink.rs
```diff
@@ -0,0 +1,55 @@
+use liquid;
+use std::path::{Path, PathBuf};
+
+use error::*;
+
+pub fn explode_permalink<S: AsRef<str>>(
```
epage: Eventually we can consider refactoring this into a newtype so we can directly reference it in the frontmatter, etc and have everything work out.

src/document.rs
```diff
@@ -13,6 +13,7 @@ use rss;
 
 use cobalt_model;
 use cobalt_model::files;
+use cobalt_model::permalink::{explode_permalink, format_url_as_file};
```
epage: Could you switch to `use cobalt_model::permalink`?

First, I prefer that for the sake of making it easier to see where the code is coming from.

Second, this masked the fact that certain tests weren't moved over (github doesn't let me comment on code outside of the diff)

src/cobalt_model/permalink.rs
Geobert: You mean, a proper `struct Permalink`?

src/document.rs
Geobert: Oh yeah, sorry I didn't noticed that it was not the adopted coding style, correcting this tomorrow :D (bed time here ^^')

epage: I wasn't going to care enough but it caused us to miss the tests.

src/cobalt_model/permalink.rs
epage: Yup.  It'll take some iterating on to figure out a design but I don't think its something you should worry about for this.

Geobert:
I have a question on this one: if `permalink_attributes` in `document.rs` is not moved, how to I call it in the pagination code to process the `{{parent}}` and others already processed in `permalink_attributes`?

epage:
> I have a question on this one: if permalink_attributes in document.rs is not moved, how to I call it in the pagination code to process the {{parent}} and others already processed in permalink_attributes?

I'm trying to move to iteratively design the `cobalt_model` API.  Anything not in there is something I don't have a good path forward on.  I think we might want to split or have all of the pagination logic outside of cobalt_model.  Feel free to make `permalink_attributes` `pub(crate)` but only use it outside of `cobalt_model`.

Geobert:
Ok for the `pub(crate)` (can't wait for edition 2018 so we can write `crate` instead). Atm I have `pagination.rs` in `cobalt_model`, should I move it to `src/` then?

epage:
Anything related to config or forntmatter should stay in cobalt_model.  Probably move the processing, including attributes, out.

Issue #395, opened 2018-03-11
[RFC] Pagination feature


Geobert:
Activation of pagination feature will also say on which index pagination will be enabled
In `_cobalt.yml`:
```
index:
    date:
        per_page: 10
        permalink: "/{{ year }}/{{ month }}/p{{ num }}"
    tags:
        per_page: 20
        permalink: "/tags/{{ tag }}/p{{ num }}"
    categories:
        per_page: 20
        permalink: "/categories/{{ categories }}/p{{ num }}"
```
where categories are defined like in https://github.com/cobalt-org/cobalt.rs/issues/131

so we'll have `/categories/MainCat1/SubCat1/p{{ num }}`, `/categories/MainCat1/SubCat2/p{{
num }}`, `/categories/MainCat2/SubCat3/p{{ num }}`…

Pagination would be accessible through a `paginator` object (to match current convention in the static web site generator world):

```
paginator.date.posts : all posts ordered by dates
paginator.date.next_page/prev_page : permalink to next/prev page

paginator.date.[year].posts : all posts for year
paginator.date.[year].next_page/prev_page

paginator.date.[year][month] : all posts for month
paginator.date.[year][month].next_page/prev_page

paginator.tags.[tag].posts : all posts for tag
paginator.tags.[tag1][tag2].posts : all posts that have both tag1 and tag2
paginator.tags.[tag].next_page/prev_page
paginator.tags._all : all the tags
```

I don't know if `[tag1][tag2]` is feasible or not but this is a first try :)

epage:
I've added some sections we'll want to fill out to help guide the conversation

> Activation of pagination feature will also say on which index pagination will be enabled
In _cobalt.yml:

So the things you can paginate on and their permalink are being defined in `_cobalt.yml` but how do I associate my `index.md` with one of these?

epage:
> permalink: "/{{ year }}/{{ month }}/p{{ num }}"

When we crate an index by `published_date`, are they all only available by year/month or are we parsing the permalink to see what variables are being used?

Geobert:
added prior art, still thinking about your questions :)

epage:
Thanks!

Some feedback
- For jekyll, that form of pagination is deprecated.  Seems like https://github.com/sverrirs/jekyll-paginate-v2 might be better to compare
- For gutenberg, that only talks about what variables are available but not how you set it up.
  - https://www.getgutenberg.io/documentation/content/section/
[...]

PR #480, opened 2018-07-06
Pagination on all posts

Geobert:
Start implementation of https://github.com/cobalt-org/cobalt.rs/issues/395
This PR can:
* paginate on `all` which takes all the posts and paginate them
* change sort order
* sort by any field from frontmatter but sub field


src/cobalt_model/pagination_config.rs
```diff
@@ -0,0 +1,108 @@
...
+//   after: i32,
+// }
+
+#[derive(Clone, Debug)]
+pub(crate) struct PaginationCfg {
+  pub(crate) include: String,
+  pub(crate) per_page: i32,
+  pub(crate) permalink: String,
+  pub(crate) order: SortOrder,
+  pub(crate) sort_by: Vec<String>,
```
epage: This seems very suspicious to me that all the members are `pub(crate)`

src/cobalt_model/pagination_config.rs
```diff
@@ -0,0 +1,108 @@
...
+  pub(crate) post_include: Option<String>
```
epage: Whats the difference between an `include` and a `post_include`?

src/cobalt_model/pagination_config.rs
```diff
@@ -0,0 +1,108 @@
...
+// struct Trails {
+//   before: i32,
+//   after: i32,
+// }
+
+#[derive(Clone, Debug)]
+pub(crate) struct PaginationCfg {
+  pub(crate) include: String,
+  pub(crate) per_page: i32,
+  pub(crate) permalink: String,
```
epage: Isn't this duplicating the `permalink` in the frontmatter?

src/cobalt_model/pagination_config.rs
```diff
@@ -0,0 +1,108 @@
...
+use super::SortOrder;
+
+// struct Trails {
+//   before: i32,
+//   after: i32,
+// }
+
+#[derive(Clone, Debug)]
+pub(crate) struct PaginationCfg {
+  pub(crate) include: String,
```
epage: Should this be an `Option` for when pagination is disabled?

Or better, make the frontmatter tracks `PaginationCfg` in an `Option` so we move that to a `None` when this is disabled.

src/cobalt.rs
```diff
@@ -139,13 +152,18 @@ fn generate_doc(posts_data: &[liquid::Value], doc: &mut Document, context: &Cont
         "page".to_owned(),
         liquid::Value::Object(doc.attributes.clone()),
     );
-    let doc_html = doc.render(&globals, &context.liquid, &context.layouts)
+    let doc_html = doc
+        .render(&globals, &context.liquid, &context.layouts)
         .chain_err(|| format!("Failed to render for {:?}", doc.file_path))?;
+    trace!(
+        "doc.file_path: {:?}",
```
epage: Is this a temp debug aid you plan to remove?

src/cobalt.rs
```diff
@@ -188,7 +237,11 @@ fn generate_posts(posts: &mut Vec<Document>, context: &Context) -> Result<()> {
             .unwrap_or(liquid::Value::Nil);
         post.attributes.insert("next".to_owned(), next);
 
-        generate_doc(&simple_posts_data, post, context)?;
+        generate_doc(
+            post,
+            context,
+            generate_collections_var(&simple_posts_data, &context),
```
epage: Whats the reason for moving the collection stuff to this level?

src/cobalt.rs
```diff
@@ -154,11 +172,42 @@ fn generate_pages(posts: Vec<Document>, documents: Vec<Document>, context: &Cont
...
-        trace!("Generating {}", doc.url_path);
-        generate_doc(&posts_data, &mut doc, context)?;
+    for mut doc in pages {
+        trace!("Generating {} / {:?}", doc.url_path, doc.file_path.to_str());
+        let mut pagination_cfg = pagination_config::PaginationCfg::new(&doc.front.pagination);
+        if pagination_cfg.is_pagination_enable() {
+            trace!("It's an index page {}", doc.url_path);
+            let paginators =
+                pagination::generate_paginators(&mut doc, &posts_data, &mut pagination_cfg);
+            // page 1 is not in "_p" folder
```
epage: Ok, so we're baking in the difference between the permalinks?  This was something I brought up earlier but I thought you were planning to leave it alone.  I'm not criticizing, just trying to better understand.

Could you elaborate on how you expect this to work?

src/pagination.rs
```diff
@@ -0,0 +1,279 @@
...
+  posts.sort_by(|a, b| {
+    let keys = config.sort_by.clone();
+    let mut cmp = Ordering::Less;
+    for k in keys {
+      cmp = if let Some(a) = extract_value(a, &k) {
+        if let Some(b) = extract_value(b, &k) {
+          match config.order {
+            SortOrder::Desc => b.partial_cmp(&a).unwrap_or(Ordering::Equal),
+            SortOrder::Asc => a.partial_cmp(&b).unwrap_or(Ordering::Equal),
+            SortOrder::None => unreachable!("We should always have a sort order default value"),
```
epage: And do you protect against someone setting it?

Ideally, that would be an error when converting the config to the in-memory model

src/pagination.rs
```diff
@@ -0,0 +1,279 @@
...
+        .as_scalar()
+        .map_or(None, |value| Some(value.clone()))
+    })
+  })
+}
+
+// sort posts by multiple criteria
+fn sort_posts(posts: &mut Vec<liquid::Value>, config: &PaginationCfg) {
+  posts.sort_by(|a, b| {
+    let keys = config.sort_by.clone();
```
epage: Why is this clone needed?  I've also seen several others that seemed suspect

src/pagination.rs
```diff
@@ -0,0 +1,279 @@
...
+
+pub(crate) fn extract_page_path(p: &liquid::Object, config: &PaginationCfg) -> PathBuf {
+  PathBuf::from(match config.include.as_str() {
+    "all" => extract_page_path_from(&p),
+    _ => "".to_owned(),
+  })
+}
+
+fn extract_value(a: &liquid::Value, key: &String) -> Option<liquid::Scalar> {
+  a.as_object().map_or(None, |attr| {
```
epage: Why aren't you using `Value::get`?

tests/fixtures/pagination_all_reverse_date/posts/my-twelfth-blogpost.md
```diff
@@ -0,0 +1,10 @@
+layout: posts.liquid
```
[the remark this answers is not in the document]
Geobert: I will reduce them, you're right, I've done this much to have 4 per page at least, but I could make 1 per page and prove it works :)

src/document.rs
```diff
@@ -132,6 +132,10 @@ fn document_attributes(
             liquid::Value::scalar(&front.collection),
         ),
         ("data".to_owned(), liquid::Value::Object(front.data.clone())),
+        (
+            "pagination".to_owned(),
+            liquid::Value::Object(front.pagination.clone()),
```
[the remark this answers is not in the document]
Geobert: This part is about gathering the pagination configuration from the front

src/cobalt_model/pagination_config.rs
Geobert: Same as the other `pub(crate)` but maybe I'm overdoing it ^^'

src/cobalt_model/pagination_config.rs
Geobert: `post_include` shouldn't been there, it's a left over from my attempt to `tags` pagination. I needed it to add the tag value after the `include` in the permalink

src/cobalt_model/pagination_config.rs
Geobert: It is indeed a struct representation of the `pagination` field from the frontmatter.

src/cobalt_model/pagination_config.rs
Geobert: You're right, I'm using `"None"` instead of `None` -_-

src/cobalt.rs
Geobert: oops, yes it's a forgotten debug log

src/cobalt.rs
Geobert: So I can pass either the `collections` object or a `paginator` object

src/cobalt.rs
Geobert: For the `all`, it works like this: I'm creating a `paginator`  per page filling them with correct `paqe` number and links to next/previous page.

For the `tags` it does not work now that I see better what we need, so this will be reworked.

src/pagination.rs
Geobert: In the reading of the frontmatter configuration, I'm defaulting to "Desc" if something's wrong

src/pagination.rs
Geobert: I don't remember, but I think I was fighting the borrow checker there

src/pagination.rs
Geobert: Because I don't understand the method when I saw it: it returns an `Option<&Self>` and `Self` is a `Value`. But I want the `Scalar`.

Maybe I'm misunderstanding here, I'm still learning Rust so sorry for all these mistakes.

Geobert:
Thank you for the review!

tests/fixtures/pagination_all_reverse_date/posts/my-twelfth-blogpost.md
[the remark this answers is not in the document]
Geobert: Oh, I remember now why so many: by default, `per_page` is `10`. So I wanted to makes sure that with more than 10 and only `include: "all"` set in the front, we still have pagination working. Can I keep the 2 tests without `per_page` like this? I'll reduce the others.

src/cobalt_model/pagination_config.rs
Geobert: Shall I remove this struct?

src/pagination.rs
Geobert: Fixed it by using a reference :)

src/pagination.rs
Geobert: Trying to use `get` like this:
```
a.get(key.clone()).map_or(None, |sort_key| {
    sort_key
      .as_scalar()
      .map_or(None, |value| Some(value.clone()))
  })
```

and got: 
```
error[E0277]: the trait bound `&liquid::Index: std::convert::From<std::string::String>` is not satisfied
  --> src\pagination.rs:45:5
   |
45 |   a.get(key.clone()).map_or(None, |sort_key| {
   |     ^^^ the trait `std::convert::From<std::string::String>` is not implemented for `&liquid::Index`
   |
             <liquid::Index as std::convert::From<&'a str>>
             <liquid::Index as std::convert::From<std::string::String>>
   = note: required because of the requirements on the impl of `std::convert::Into<&liquid::Index>` for `std::string::String`
```
Which totally lost me.

src/cobalt_model/pagination_config.rs
```diff
@@ -0,0 +1,108 @@
...
+use std::vec::Vec;
+use super::SortOrder;
+
+// struct Trails {
+//   before: i32,
+//   after: i32,
+// }
+
+#[derive(Clone, Debug)]
+pub(crate) struct PaginationCfg {
```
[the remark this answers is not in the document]
Geobert: Renamed to `PaginationConfig`, removing all `pub(crate)` from my code

Geobert:
>  All user-facing cobalt formats are strictly defined using structs / serde

You're refering to `pagination` being an `Object` rather than a struct?

> All user-facing structs are kept separate from in-memory structs

I don't understand this line

> While there are exceptions, hard coded values should be enums instead of strings

You're refering to "all", "tags" etc that are str atm?

tests/fixtures/pagination_all_reverse_date/posts/my-twelfth-blogpost.md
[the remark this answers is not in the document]
epage: Makes sense.  Do we need two tests without `per_page` or only one?

src/cobalt_model/frontmatter.rs
```diff
@@ -80,6 +80,8 @@ pub struct FrontmatterBuilder {
     pub is_draft: Option<bool>,
     #[serde(skip_serializing_if = "liquid::Object::is_empty")]
     pub data: liquid::Object,
+    #[serde(skip_serializing_if = "liquid::Object::is_empty")]
+    pub pagination: liquid::Object,
```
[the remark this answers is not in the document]
epage: `data` is for arbitrary user-driven data.  Everything else is structured.

src/cobalt_model/pagination_config.rs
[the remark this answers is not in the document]
epage: `pub(crate)` is an interesting one.  If we feel we need access to the guts of something, then might the users of the API?  We should have clear reasons for using it (part of a specific abstraction that, for reasons, we didn't just put in a single file).  Of course, transitional refactoring is another valid reason.

tests/fixtures/pagination_all_reverse_date/posts/my-twelfth-blogpost.md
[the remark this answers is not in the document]
Geobert: Maybe not, the 2nd one it to reverse the order, can be tested with a smaller set of posts :)

src/cobalt_model/frontmatter.rs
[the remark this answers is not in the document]
Geobert: Especially when the struct is already there… This is basically `PaginationConfig`

src/cobalt_model/pagination_config.rs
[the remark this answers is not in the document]
Geobert: I keep forgetting that cobalt is also a lib… ^^'

src/document.rs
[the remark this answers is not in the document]
Geobert: There is no paginator in the front, only pagination configuration… I may have misunderstood this part badly :-/

I'm exposing to liquid here https://github.com/Geobert/cobalt.rs/blob/968c68ab3e30c5645d6b455a0e85e85a1d6510a4/src/cobalt.rs#L185

Is that incorrect?

src/cobalt.rs
Geobert: Yeah, I though that if we activate pagination, no need to have the posts twice.

src/cobalt_model/pagination_config.rs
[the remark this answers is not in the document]
epage: Well, we do want to split jekyll migration out into a separate program :)

src/document.rs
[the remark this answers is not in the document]
epage: `document_attributes` is for exposing variables to liquid templates.

So you are exposing both a `paginator` and a `pagination`?  What is the role of the `pagination`?

src/cobalt.rs
epage: Makes sense, we just need to make sure we make this clear in the documentation.

src/pagination.rs
epage: Maybe try
`a.get(&key.clone().into()).flat_map(|sort_key| sort_key.as_scalar())`

`Index` is a bit problematic in liquid's API.  At first, the error sounds weird but I think its complaining about only having ways to convert the param into an `Index` and not an `&Index`.

The last half I think is how to handle an `Option<Option<_>>` when working with `map`.

epage:
> You're refering to pagination being an Object rather than a struct?

Yes.  One problem you'll run into is the "short cuts".  This would be handled by using an `Enum`.

> >   All user-facing structs are kept separate from in-memory structs
> I don't understand this line

Notice we have `SiteConfig`, `SiteBuilder`, and `Site`.  Originally, I had `SiteConfig` just be the `SiteBuilder` but I quickly found problems when conflating the two.

> You're refering to "all", "tags" etc that are str atm?

Yup

epage:
Ok, still got a couple comments to reply to.

src/cobalt_model/pagination_config.rs
epage: I'm a little confused by this comment.  Is `frontmatter.pagination.permalink` duplicate of `frontmatter.permalink`?  If not, what unique role do they fulfill?

src/cobalt.rs
epage: > > Could you elaborate on how you expect this to work?

I'm not as much concerned about the implementation atm but the policy.

src/pagination.rs
epage: What if someone asks for `None`?

src/document.rs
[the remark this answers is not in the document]
Geobert: I see, it's a mistake then, I wasn't intending to expose the pagination configuration, I'll remove this :)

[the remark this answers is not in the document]
Geobert: `pagination ` is the front to activate and configure the pagination process, so no need to expose it. It's removed :)

src/cobalt_model/pagination_config.rs
Geobert: Oh I see! `frontmatter.pagination.permalink` is the one we defined for pagination only, to configure where the different indexes will go, not related to `frontmatter.permalink`.

src/pagination.rs
Geobert: Good question… I don't know ^^'  Defaulting to `Desc`? (As it's the default value I choose for the `order` field)

src/cobalt.rs
Geobert: Policy wise, I'm intending to generate all the different paginator (replacing the `{{ num }}` basically) in this function and use `generate_doc` to create the file, modifying `file_path` according to the `pagination.permalink`.

I don't know if I'm answering properly to your question here ^^'

src/pagination.rs
Geobert: I got:
```rust
error[E0599]: no method named `flat_map` found for type `std::option::Option<&liquid::Value>` in the current scope
  --> src\pagination.rs:54:30
   |
54 |   a.get(&key.clone().into()).flat_map(|sort_key| sort_key.as_scalar())
   |                              ^^^^^^^^
   |
   = note: the method `flat_map` exists but the following trait bounds were not satisfied:
           `&mut std::option::Option<&liquid::Value> : std::iter::Iterator`
```

Then I tried `a.get(&key.clone().into()).map_or(None, |k| k.flat_map(|sort_key| sort_key.as_scalar()))`
```rust
error[E0599]: no method named `flat_map` found for type `&liquid::Value` in the current scope
  --> src\pagination.rs:54:49
   |
54 |   a.get(&key.clone().into()).map_or(None, |k| k.flat_map(|sort_key| sort_key.as_scalar()))
   |                                                 ^^^^^^^^
   |
   = note: the method `flat_map` exists but the following trait bounds were not satisfied:
           `&mut &liquid::Value : std::iter::Iterator`
           `&mut liquid::Value : std::iter::Iterator`
```

Geobert:
Applied all the comments (I think) but the tests and 
> All user-facing structs are kept separate from in-memory structs

About the latter, you're suggesting that I should do a `PaginationBuilder`, `PaginationConfig` and `Paginator` structs? The latter being converted to `liquid::Object` at the very last moment?

src/pagination.rs
epage: We should probably error if someone explicitly asks for an unsupported value

src/pagination.rs
epage: Cloning the key is an annoying part of liquid's API.

Do you need to clone the scalar?

epage:
> About the latter, you're suggesting that I should do a PaginationBuilder, PaginationConfig and Paginator structs? The latter being converted to liquid::Object at the very last moment?

Yes

src/cobalt.rs
```diff
@@ -9,12 +9,14 @@ use std::path;
 
 use cobalt_model;
 use cobalt_model::files;
+// use cobalt_model::pagination_config;
```
epage: Dead code?

src/cobalt_model/frontmatter.rs
```diff
@@ -205,6 +213,10 @@ impl FrontmatterBuilder {
         self.merge(Self::new().set_categories(categories.into()))
     }
 
+    pub fn merge_pagination<S: Into<Option<pagination_config::PaginationConfig>>>(self, pagination: S) -> Self {
+        self.merge(Self::new().set_pagination(pagination.into()))
```
epage: This merge implementation means that we throw out any pagination configuration set in `_cobalt.yml` yet the RFC says you can set default values in there.

src/cobalt_model/frontmatter.rs
```diff
@@ -351,7 +328,7 @@ impl FrontmatterBuilder {
             is_draft: is_draft.or_else(|| other_is_draft),
             collection: collection.or_else(|| other_collection),
             data: merge_objects(data, other_data),
-            pagination: merge_objects(pagination, other_pagination),
+            pagination: pagination.or_else(|| other_pagination),
```
epage: ditto about losing defaults

src/cobalt_model/frontmatter.rs
```diff
@@ -468,7 +445,7 @@ pub struct Frontmatter {
     pub is_draft: bool,
     pub collection: String,
     pub data: liquid::Object,
-    pub pagination: liquid::Object,
+    pub pagination: Option<pagination_config::PaginationConfig>,
```
epage: Should this still hold a `Config` or a `Builder`?

src/cobalt_model/pagination_config.rs
```diff
@@ -1,108 +1,109 @@
...
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
+pub enum Include {
+  None,
+  #[serde(rename = "all")]
+  All,
+  // tags,
+}
+
+impl Into<liquid::Scalar> for Include {
```
epage: I'd personally recommend removing the `liquid` policy from this and have it just return `&'static str`.

src/cobalt_model/pagination_config.rs
```diff
@@ -1,108 +1,109 @@
...
-  //trails: Trails,
+impl Default for Include {
+  fn default() -> Include {
+    Include::None
+  }
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(deny_unknown_fields, default)]
+pub struct PaginationConfig {
```
epage: Somewhere between this and the frontmatter, we are not handling the shortcuts.

src/cobalt.rs
```diff
@@ -174,25 +170,34 @@ fn generate_pages(posts: Vec<Document>, pages: Vec<Document>, context: &Context)
...
     for mut doc in pages {
         trace!("Generating {} / {:?}", doc.url_path, doc.file_path.to_str());
-        let mut pagination_cfg = pagination_config::PaginationCfg::new(&doc.front.pagination);
-        if pagination_cfg.is_pagination_enable() {
+        let pagination_enabled = {
+            doc.front
+                .pagination
+                .as_ref()
+                .map_or(false, |pagination_config| {
+                    pagination_config.is_pagination_enable()
```
epage: If we are accessing a non-config version of the paginator, then we could just check whether `opt.is_some()` or even `if let Some(pagination) = ...`.

src/pagination.rs
```diff
@@ -1,27 +1,28 @@
...
-    "all" => {
+  match config.include {
+    Include::All => {
       sort_posts(&mut all_posts, &config);
       create_all_paginators(&mut all_posts, &doc, &config)
     }
-    _ => {
-      unreachable!();
-    }
+    Include::None => Vec::new(),
```
epage: Shouldn't this be a panic/unreachable?

src/pagination.rs
```diff
@@ -34,35 +35,32 @@ fn extract_page_path_from(p: &liquid::Object) -> String {
     .into_owned()
 }
 
-pub(crate) fn extract_page_path(p: &liquid::Object, config: &PaginationCfg) -> PathBuf {
-  PathBuf::from(match config.include.as_str() {
-    "all" => extract_page_path_from(&p),
+pub fn extract_page_path(p: &liquid::Object, config: &PaginationConfig) -> PathBuf {
+  PathBuf::from(match config.include {
+    Include::All => extract_page_path_from(&p),
     _ => "".to_owned(),
```
epage: Please cover all cases.  Its a big help for when you add new ones, the compiler will tell you when you miss something.

Also, `""` or `panic`/`unreachable`?

src/pagination.rs
```diff
@@ -34,35 +35,32 @@ fn extract_page_path_from(p: &liquid::Object) -> String {
...
     let mut cmp = Ordering::Less;
     for k in keys {
       cmp = if let Some(a) = extract_value(a, &k) {
         if let Some(b) = extract_value(b, &k) {
           match config.order {
             SortOrder::Desc => b.partial_cmp(&a).unwrap_or(Ordering::Equal),
             SortOrder::Asc => a.partial_cmp(&b).unwrap_or(Ordering::Equal),
-            SortOrder::None => unreachable!("We should always have a sort order default value"),
+            SortOrder::None => unreachable!(
+              "Sort order should have default value when constructing PaginationConfig"
```
epage: This message is incorrect or bad.  A default value is not good enough to protect us from reaching this code path.  We need to explicitly check for the user setting it and disallow it.

src/cobalt.rs
```diff
@@ -174,25 +170,34 @@ fn generate_pages(posts: Vec<Document>, pages: Vec<Document>, context: &Context)
...
                 let mut doc_page = doc.clone();
-                doc_page.file_path = pagination::extract_page_path(&p, &pagination_cfg);
+                doc_page.file_path = pagination::extract_page_path(
+                    &p,
+                    &doc.front
+                        .pagination
+                        .as_ref()
+                        .expect("Must have a pagination"),
+                );
                 println!("doc_page.file_path: {:?}", doc_page.file_path);
```
epage: Please remove or find an appropriate way of framing this for logging.

src/cobalt.rs
```diff
@@ -174,25 +170,34 @@ fn generate_pages(posts: Vec<Document>, pages: Vec<Document>, context: &Context)
...
+        let pagination_enabled = {
+            doc.front
+                .pagination
+                .as_ref()
+                .map_or(false, |pagination_config| {
+                    pagination_config.is_pagination_enable()
+                })
+        };
+        if pagination_enabled {
             trace!("It's an index page {}", doc.url_path);
```
epage: Please remove

src/pagination.rs
```diff
@@ -0,0 +1,273 @@
...
+// sort posts by multiple criteria
+fn sort_posts(posts: &mut Vec<liquid::Value>, config: &PaginationConfig) {
+  posts.sort_by(|a, b| {
+    let keys = &config.sort_by;
+    let mut cmp = Ordering::Less;
+    for k in keys {
+      cmp = if let Some(a) = extract_value(a, &k) {
+        if let Some(b) = extract_value(b, &k) {
+          match config.order {
+            SortOrder::Desc => b.partial_cmp(&a).unwrap_or(Ordering::Equal),
```
epage: Its a bit annoying having this match in the inner loop.  It makes the code harder to follow.

Maybe express these as lambdas?

```rust
let cmp = match config.order {
    SortOrder::Desc => |a, b| b.partial_cmp(&a).unwrap_or(Ordering::Equal),
    ...
}
```
(would require renaming `cmp` to `order` or something)

src/pagination.rs
```diff
@@ -0,0 +1,273 @@
...
+  a.get(&key.clone().into())
+    .map_or(None, |sort_key| sort_key.as_scalar().cloned())
+}
+
+// sort posts by multiple criteria
+fn sort_posts(posts: &mut Vec<liquid::Value>, config: &PaginationConfig) {
+  posts.sort_by(|a, b| {
+    let keys = &config.sort_by;
+    let mut cmp = Ordering::Less;
+    for k in keys {
```
epage: atm we aren't allowing dotted keys.  We should check and report errors if we get back a dotted key when converting the Config to a Builder or something.

src/pagination.rs
```diff
@@ -0,0 +1,273 @@
...
+      cmp = if let Some(a) = extract_value(a, &k) {
+        if let Some(b) = extract_value(b, &k) {
```
epage: I kind of dislike the nested ifs, I feel it makes it harder to follow.  Trying to come up with better ideas.

What if we did:
```rust
match (extract_value(a, &k), extract_value(b, &k) {
    (Some(a), Some(b)) => {
        ...
    }
    (None, None) => Ordering::Equal,
    (_, None) => Ordering::Greater,
    (None, _) => Ordering::less,
}
```

src/pagination.rs
```diff
@@ -0,0 +1,273 @@
...
+  let cfg = config.clone();
+  let mut attributes = document::permalink_attributes(&doc.front, &doc.file_path);
+  let pagination_attr = pagination_attributes(page_num, cfg.include);
+  pagination_attr.into_iter().for_each(|(k, v)| {
+    attributes.insert(k, v);
+  });
+  let permalink = permalink::explode_permalink(&config.permalink, &attributes);
+  let p = permalink
+    .and_then(|mut p| {
+      p.push_str(file_name);
```
epage: So the user can set a custom permalink but then we unconditionally add the filename?  Seems like that should be a variable or something.

src/pagination.rs
```diff
@@ -0,0 +1,273 @@
...
+) {
+  paginator.insert(
+    "per_page".to_owned(),
+    liquid::Value::scalar(config.per_page),
+  );
+  paginator.insert("total_posts".to_owned(), liquid::Value::scalar(total_posts));
+  paginator.insert("total_pages".to_owned(), liquid::Value::scalar(total_pages));
+  paginator.insert(
+    "first_page_path".to_owned(),
+    liquid::Value::scalar(format!("/{}", doc.file_path.to_string_lossy())),
```
epage: Why use `file_path` instead of `url_path`?

src/cobalt.rs
epage: That is still implementation and not policy.

From looking at the code, the answer is that
- The "first" page is `front.permalink`
- All other pages use `front.pagination.permalink`

One problem with this code is that the comment does not say this.  Instead the comment makes it sound like somewhere else we have a baked-in policy.

src/cobalt_model/pagination_config.rs
epage: On second thought, I'm wanting to keep this PR as small as possible to enable easier reviewing.  I'd prefer if shortcuts and other types of pagination are explicitly listed in the PR as not included and that we handle those in separate PRs.

Geobert: I mostly agree, but I might need to have "tags" (if we keep it, as it is under discussion in the RFC) as it helps me to see if my code can handle more generic cases. The `all` case is quite simple, but when I started to code `tags` I had to change some things around.

src/cobalt.rs
Geobert: I think I'll end up with the compilation error I've asked about on the Rust forum. I'll try though.

src/pagination.rs
Geobert: What's a "dotted key"?

src/pagination.rs
Geobert: I was thinking the permalink was defining a directory for the file to be, I may be wrong

src/pagination.rs
Geobert: I've overlook `url_path`, it may be more adapted, I'll have a look

src/cobalt.rs
Geobert: You might missed it, I've added it to the RFC, under the paginator fields code block.

src/pagination.rs
epage: "data.custom_field.custom_value`.

src/pagination.rs
epage: THat is correct.

src/cobalt_model/pagination_config.rs
Geobert: I've done this:
```rust
impl Into<&'static str> for Include {
  fn into(self) -> &'static str {
    match self {
      Include::None => "",
      Include::All => "all",
    }
  }
}
```

but with:
```rust
attributes.insert("include".to_owned(), liquid::Value::scalar(include.into()));
```

I got this error:
```rust
error[E0283]: type annotations required: cannot resolve `_: std::convert::Into<liquid::Scalar>`
   --> src\pagination.rs:236:43
    |
236 |   attributes.insert("include".to_owned(), liquid::Value::scalar(include.into()));
    |                                           ^^^^^^^^^^^^^^^^^^^^^
    |
    = note: required by `liquid::Value::scalar`
```

I had to do this to make it compiles:
```rust
let i: &str = include.into();
attributes.insert("include".to_owned(), liquid::Value::scalar(i));
```

I tried `liquid::Value::scalar(include.into() as &str)` with no luck.

Is there a trick to avoid this ugly extra line just to have a type annotation?

src/pagination.rs
Geobert: Thank you so much, this + the lambda idea above taught me a lot on effective use of match!

src/cobalt_model/pagination_config.rs
epage: This error is a bit misleading.

I think the problem is the fact that we have
- `Into<&str> for Include`
- `Into<Scalar> for &str`

It can't know to call `into` twice to convert `Include` into `&str` into `Scalar`.

You could simplify it to
```rust
let i: &str = include.into();
attributes.insert("include".to_owned(), i.into());
```
It should know that the value being inserted is a `Scalar` so it should be able to call the into correctly.

Geobert: It seems that `liquid::Value::scalar(i)` is mandatory. With `i.into()`:
```rust
error[E0277]: the trait bound `liquid::Value: std::convert::From<&str>` is not satisfied
   --> src\pagination.rs:233:45
    |
233 |   attributes.insert("include".to_owned(), i.into());
    |                                             ^^^^ the trait `std::convert::From<&str>` is not implemented for `liquid::Value`
    |
    = note: required because of the requirements on the impl of `std::convert::Into<liquid::Value>` for `&str`
```

src/cobalt_model/frontmatter.rs
Geobert: I'm trying to get my head around the merge process, and if what you said here is correct, all the other fields are impacted as I've basically copied how the other field are merged :-/

Geobert:
All comments have been addressed except the merge issue, I've tested it and indeed, it does not work but I don't understand how `config`'s defaults are merging into frontmatter

Geobert:
> github's review system is terrible. 

yeah especially the auto hidden post feature :(

Thank you again for your review, I'll take care of it asap :)

src/cobalt_model/pagination_config.rs
```diff
@@ -0,0 +1,59 @@
+use std::convert::Into;
+use std::vec::Vec;
+
+use super::SortOrder;
+
+#[derive(Copy, Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
+pub enum Include {
+  None,
+  #[serde(rename = "all")]
```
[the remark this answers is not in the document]
Geobert: Rust styling force capital for the first char. But it feels odd to me to write `include: All` instead of `include: all`. Maybe it's just me though, let me know what you prefer :)

src/cobalt.rs
```diff
@@ -187,22 +187,22 @@ fn generate_pages(posts: Vec<Document>, pages: Vec<Document>, context: &Context)
...
-                doc_page.file_path = pagination::extract_page_path(
-                    &p,
-                    &doc.front
-                        .pagination
-                        .as_ref()
-                        .expect("Must have a pagination"),
-                );
-                println!("doc_page.file_path: {:?}", doc_page.file_path);
+                doc_page.file_path =
+                    permalink::format_url_as_file(paginator.index_permalink.clone());
```
epage: This function takes a reference, you shouldn't need to clone.  You do need to explicitly state you are passing a reference: `permalink::format_url_as_file(&paginator.index_permalink);`

epage:
Sorry for the delay in getting back to you.  I was trying to get some things in for Rust 2018 Preview

epage:
When I did my last pass, I felt like there were issues that weren't addressed.

Trying to find ways to simplify this.  one thought:
- A PR with just the `cargo fmt` fixes
- A PR for creating `permalink.rs`, any renames, passing `collections` to `generate_docs`, etc
- Create a PR that is just the frontmatter.  We can iterate a little on the data model and get that in.
  - We could hide this behind a feature toggle with "unstable" in the name.  A minimal-invasive way to check for it is to error in the frontmatter builder if the feature isn't toggled and paginator is configured.
- We then have a PR for the core logic.


Don't worry about maintaining perfect history with this review.  Feel free to create a branch, dump your code on top of it, and create a PR.

Geobert:
> Sorry for the delay in getting back to you. I was trying to get some things in for Rust 2018 Preview

No bother :) I'm in holidays so I'm stepping back a bit from coding atm :) I'll do all the requests when I come back.

epage:
Good call disconnecting while on holiday

Geobert:
> Trying to find ways to simplify this. one thought:

I was in the middle of doing the changes before my holidays, I'll finish first then split the whole stuff into those 4 PRs. So you will see some pushes here, but don't bother to review them yet, you'll do that on the proper PRs to save some sanity ;)

epage:
> I don't understand the test failure: I don't have it locally :-/

Does `git status` show anything of interest?  The error makes it sound like a directory is missing.

Geobert:
Thanks for your answer!

It's what I though but:
```
> git status
On branch pagination
Your branch is up to date with 'origin/pagination'.

nothing to commit, working tree clean
```

Geobert:
I tried on my server (debian) and no issue neither Oo

Geobert:
It was a new `git clone` to be sure I was not doing anything weird with git.
Good catch though! Trying it right now!

Geobert:
This one is obsolete, another PR is incoming to finish the feature :)

PR #496, opened 2018-08-15
fix(fmt) apply 1.28 rustfmt

Geobert:


epage:
Please also update travis
https://github.com/cobalt-org/cobalt.rs/blob/master/.travis.yml#L29

Geobert:
True, what about nightly version for clippy?

epage:
> True, what about nightly version for clippy?

If you want.  I'd recommend doing it in a separate commit.

> EDIT: anyway to just say "latest stable" in travis like in appveyor?

We intentionally do not say "latest stable".  The concern is that an updated stable will break our build.

epage:
Looks good!

Feel free to squash and let me know when done, and I'll merge.

Geobert:
Squashed! :D

epage:
Thanks!

PR #497, opened 2018-08-25
feat(pagination): pagination frontmatter

Geobert:
Here is the frontmatter part of the pagination feature :)

I'm kinda slow because new game but I'm still in to complete this feature!

You wanted to put a feature flag on this to avoid disruption on the master branch, but I don't know where to put it :-/

`PaginationConfigBuilder::build`?

### DIFF
--- a/src/cobalt_model/frontmatter.rs
+++ b/src/cobalt_model/frontmatter.rs
@@ -11,6 +11,7 @@ use serde_yaml;
 use error::Result;
 
 use super::datetime;
+use super::pagination_config;
 use super::slug;
 
 const PATH_ALIAS: &str = "/{{parent}}/{{name}}{{ext}}";
@@ -80,6 +81,8 @@ pub struct FrontmatterBuilder {
     pub is_draft: Option<bool>,
     #[serde(skip_serializing_if = "liquid::value::Object::is_empty")]
     pub data: liquid::value::Object,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub pagination: Option<pagination_config::PaginationConfigBuilder>,
     // Controlled by where the file is found.  We might allow control over the type at a later
     // point but we need to first define those semantics.
     #[serde(skip)]
@@ -129,6 +132,16 @@ impl FrontmatterBuilder {
     pub fn set_categories<S: Into<Option<Vec<String>>>>(self, categories: S) -> Self {
         Self {
             categories: categories.into(),
+            ..self
+        }
+    }
+
+    pub fn set_pagination<S: Into<Option<pagination_config::PaginationConfigBuilder>>>(
+        self,
+        pagination: S,
+    ) -> Self {
+        Self {
+            pagination: pagination.into(),
             ..self
         }
     }
@@ -214,6 +227,13 @@ impl FrontmatterBuilder {
         self.merge(Self::new().set_published_date(published_date.into()))
     }
 
+    pub fn merge_pagination<S: Into<Option<pagination_config::PaginationConfigBuilder>>>(
+        self,
+        secondary: S,
+    ) -> Self {
+        self.merge(Self::new().set_pagination(secondary.into()))
+    }
+
     #[cfg(test)]
     pub fn merge_format<S: Into<Option<SourceFormat>>>(self, format: S) -> Self {
         self.merge(Self::new().set_format(format.into()))
@@ -247,6 +267,7 @@ impl FrontmatterBuilder {
             is_draft,
             collection,
             data,
+            pagination,
         } = self;
         Self {
             permalink,
@@ -262,6 +283,7 @@ impl FrontmatterBuilder {
             is_draft,
             collection,
             data: merge_objects(data, other_data),
+            pagination,
         }
     }
 
@@ -280,6 +302,7 @@ impl FrontmatterBuilder {
             is_draft,
             collection,
             data,
+            pagination,
         } = self;
         let Self {
             permalink: other_permalink,
@@ -295,6 +318,7 @@ impl FrontmatterBuilder {
             is_draft: other_is_draft,
             collection: other_collection,
             data: other_data,
+            pagination: other_pagination,
         } = other;
         Self {
             permalink: permalink.or_else(|| other_permalink),
@@ -310,6 +334,7 @@ impl FrontmatterBuilder {
             is_draft: is_draft.or_else(|| other_is_draft),
             collection: collection.or_else(|| other_collection),
             data: merge_objects(data, other_data),
+            pagination: merge_pagination(pagination, other_pagination),
         }
     }
 
@@ -366,6 +391,7 @@ impl FrontmatterBuilder {
             is_draft,
             collection,
             data,
+            pagination,
         } = self;
 
         let collection = collection.unwrap_or_else(|| "".to_owned());
@@ -394,9 +420,14 @@ impl FrontmatterBuilder {
             is_draft: is_draft.unwrap_or(false),
             collection,
             data,
+            pagination: pagination.and_then(|p| p.build()),
         };
 
-        Ok(fm)
+        if !cfg!(feature = "pagination-unstable") && fm.pagination.is_some() {
+            Err("Unsupported `pagination` field".into())
+        } else {
+            Ok(fm)
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


Geobert:
It seems that Travis failed to run Oo

src/cobalt_model/frontmatter.rs
```diff
@@ -214,6 +227,13 @@ impl FrontmatterBuilder {
         self.merge(Self::new().set_published_date(published_date.into()))
     }
 
+    pub fn merge_pagination<S: Into<Option<pagination_config::PaginationConfigBuilder>>>(
+        self,
+        secondary: S,
+    ) -> Self {
+        self.merge(Self::new().set_pagination(secondary.into()))
```
epage: Like `merge_data`, should this call `merge_pagination`?

src/cobalt_model/pagination_config.rs
```diff
@@ -0,0 +1,155 @@
...
+
+pub const DEFAULT_PERMALINK: &str = "{{parent}}/{{include}}/_p/{{num}}/";
+pub const DEFAULT_SORT: &str = "published_date";
+pub const DEFAULT_PER_PAGE: i32 = 10;
+
+#[derive(Copy, Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
+pub enum Include {
+    None,
+    #[serde(rename = "all")]
```
epage: Do we nornally do `rename` for our enums?  Whatever we specified, I think I'd rather we just be consistent.

src/cobalt_model/pagination_config.rs
```diff
@@ -0,0 +1,155 @@
...
+        let Self {
+            include,
+            per_page,
+            permalink,
+            order,
+            sort_by,
+        } = self;
+
+        let include = include.unwrap_or(Include::None);
+        if include != Include::None {
```
epage: In cases like this, I'm a fan of early-returns.

src/cobalt_model/pagination_config.rs
```diff
@@ -0,0 +1,155 @@
...
+    pub include: Include,
+    pub per_page: i32,
+    pub permalink: String,
+    pub order: SortOrder,
+    pub sort_by: Vec<String>,
+}
+
+impl Default for PaginationConfig {
+    fn default() -> PaginationConfig {
+        PaginationConfig {
```
epage: Couldn't you reduce duplication by `PaginationConfigBuilder::default().build()`?

src/cobalt_model/frontmatter.rs
```diff
@@ -394,6 +420,7 @@ impl FrontmatterBuilder {
             is_draft: is_draft.unwrap_or(false),
             collection,
             data,
+            pagination: pagination.and_then(|p| p.build()),
         };
 
         Ok(fm)
```
epage: I think this would be a great spot for putting in the feature flag.  If `pagination` is not `None`, then we error.  It shouldn't cause warnings when the flag is not enabled and this function already returns errors.

It could allow someone to configure pagination with `include: None`.  Oh well.

epage:
Sorry for the delay.

Geobert:
> Sorry for the delay.

Don't be sorry, nothing urgent here :D I hope you're well. 

I'll tackle the change requests asap :)

src/cobalt_model/pagination_config.rs
Geobert: You mean this?
```
if include == Include::None {
  return None;
}
```

epage: Yup

src/cobalt_model/frontmatter.rs
epage: At first I commented on this, thinking extra processing is needed for deep members but I'm not even sure `merge_data` needs to be special cased.

src/cobalt_model/pagination_config.rs
Geobert: Removing the `impl Default` makes the `Deserialize` derivation fail:
```
--> src\cobalt_model\pagination_config.rs:133:35
    |
133 | #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
    |                                   ^^^^^^^^^^^ the trait `std::default::Default` is not implemented for `cobalt_model::pagination_config::PaginationConfig`
    |
```

epage: I'm not suggesting removing it, I'm suggesting changing how its implemented.

Right now, it initializes `PaginationConfig` to match the defaulting applied in `build`.  You can reduce this duplication by

```rust
impl Default for PaginationConfig {
    fn default() -> PaginationConfig {
        PaginationConfigBuilder::default().build()
    }
}
```

Hmm, now that I look at things, maybe that won't work because `build()` will turn it into a `None`.

Which gets me thinking, why does this support serde?  I'm assuming its for doing cleaner debug dumps.  We only need `Deserialize` then and don't need `#[serde(deny_unknown_fields, default)]`

At that point, we don't need a default here which I don't think will actually serve any purpose.

src/cobalt_model/frontmatter.rs
Geobert: I don't know if I'm doing it properly:
```rust
    pagination: pagination.and_then(|p| p.build()),
        }; 

        if !cfg!(feature = "pagination-unstable") && fm.pagination.is_some() {
            Err("Unsupported `pagination` field".into())
        } else {
            Ok(fm)
        }        
```

src/cobalt_model/pagination_config.rs
Geobert: > Hmm, now that I look at things, maybe that won't work because build() will turn it into a None.
That's why I tried to removed it in the first place :)

> Which gets me thinking, why does this support serde? 

Same question for https://github.com/cobalt-org/cobalt.rs/blob/8fbca2590ec8d9a7c6490990ff5f8c937db76f88/src/cobalt_model/frontmatter.rs#L441

I think I've put `Serialize` and `Deserialize` because `Frontmatter` has them

epage: Mind doing a PR real quick that removes `Serialize` from anything not loaded from disk?

src/cobalt_model/frontmatter.rs
epage: Looks about right.  Running into problems with it?

Geobert: Nope, just first time using `cfg!` :p

src/cobalt_model/pagination_config.rs
Geobert: Ok :)

Geobert: To get it right: `FrontmatterBuilder` *is* loaded from disk but not `Frontmatter` (as it's being created from the builder) same goes for any struct created by a builder. Am I right?

epage: I got fixed up, we need serialize for to_string / printing debug information.

Deserialize is what we don't need.  Only config stuff gets loaded from disk.  The rest should only ever be used programmatically.

Geobert: Sorry for the delay, I got another try this morning and I don't see how to do it: it breaks a lot of stuff.

`Frontmatter` doesn't need `Deserialize` but `FrontmatterBuilder` does and both `impl Front` which has this `for<'de> serde::Deserialize<'de>` which I don't quite understand.

epage: Sorry, I forgot about `Front`.  Feel free to ignore dropping `Serialize`.  I can worry about it in a future cleanup.

Geobert: Roger that, I'll go back on the PR for pagination :)

src/cobalt_model/frontmatter.rs
Geobert: So nothing to be done here in the end?

src/cobalt_model/pagination_config.rs
Geobert: > Hmm, now that I look at things, maybe that won't work because `build()` will turn it into a `None`.

So is there still something to do here?

src/cobalt_model/frontmatter.rs
epage: Yeah, this should be fine for now.

src/cobalt_model/pagination_config.rs
epage: Yeah, this is fine for now.

Geobert:
I've squashed the commits, it's ready to merge :D

epage:
Thanks!

Geobert:
Thank you! :) I'll start the final PR for the feature ASAP, less time for coding these days :)

PR #516, opened 2018-09-07
Replace unidecode with deunicode

kornelski:
deunicode a in improved fork of unidecode. It adds support for emoji. Uses compact representation of the mappings, which reduce compilation time, memory usage and binary size.


epage:
> assert_eq!(deunicode("🦄☣"), "unicorn face biohazard");

Nice example :)

kornelski:
I see the build failure is caused by rustfmt being annoyingly picky. I hate rustfmt :(

PR #517, opened 2018-09-12
chore: Replace unidecode with deunicode

epage:


epage:
Supersedes https://github.com/cobalt-org/cobalt.rs/pull/516

Thanks @kornelski

PR #528, opened 2018-10-05
Update pinned liquid sub-crates

chippers:
Because the `liquid-{compiler, interpreter, value}` crates were still pinned to `0.16.0`, the fix from cobalt-org/liquid-rust#207 , wasnt passed down to #527.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -186,7 +186,7 @@ dependencies = [
 
 [[package]]
 name = "cobalt-bin"
-version = "0.13.0"
+version = "0.13.1"
 dependencies = [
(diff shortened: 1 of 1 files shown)


epage:
Sorry about that

chippers:
it's no problem :)

epage:
Just pushed out the new tag, so you'll have a binary when thats done, ~20 minutes.

chippers:
dont feel pressured to put out binaries too quick if it's just for me, I have both cobalt and liquid checked out on my machine and all hooked up with cargo replaces from when i was messing around with liquid, and I can build from a clean release in a few minutes.

epage:
Eh, its all automated once I tag a commit.

PR #529, opened 2018-10-06
Add "http://" prefix to server address

KyrychukD:
Makes it possible to open the site in a browser with Ctrl+click.


KyrychukD:
The end result looks like this:
selection_086

epage:
Thanks for doing this and providing the motivation!

Issue #431, opened 2018-04-25
Support jekyll-style frontmatter in the content


berkus:
I don't know why decision was made to not follow the jekyll frontmatter syntax, but some editors, notably Sublime Text include highlighting schemes that rely on it, compare:

1. Cobalt front matter
![one-separator-st3](https://user-images.githubusercontent.com/80306/39259459-20acd7aa-48bf-11e8-9fc7-0f05eb8c3fed.png)

2. Jekyll front matter
![two-separators-st3](https://user-images.githubusercontent.com/80306/39259466-255a14de-48bf-11e8-9900-fa3c9079af41.png)

And I believe some other markdown editors might be doing the same - e.g. github is able to extract jekyll front matter in a .md page and show it in tabular form, see for example [here](https://github.com/metta-systems/metta.systems/blob/80a0f5a42e966eba95e22277bf69b026a2fa0627/content/post/2012-07-16-graphic-dependency-resolution.md) while with cobalt front matter things look [much much worse by default](https://github.com/cobalt-org/cobalt-org.github.io/blob/source/help.md).

epage:
I think we could implement this by checking for `"^---\n.*---\n"` in addition to just `"---\n"`.

The code is here
https://github.com/cobalt-org/cobalt.rs/blob/master/src/cobalt_model/document.rs#L46

We could look to the jekyll import for inspiration
https://github.com/cobalt-org/cobalt.rs/blob/master/src/jekyll_model/document.rs#L49

freiguy1:
Would this be a requirement for documents or optional? In other words, if both triple-dashes were required, in this example:

```
title: The Title
---
The content
```

There would be no frontmatter since there's only one triple-dash?

I have a feeling implementing both triple-dashes as a requirement would be easier however, it's a breaking change for anyone using cobalt. If both triple-dashes were optional, that's also tricky since `---` is markdown for horizontal rule...

berkus:
Both were supported really, so it's not a problem of implementing, it's just a problem on settling on one - and I believe jekyll-style frontmatter is wider present than Cobalt one. However, *sigh*

epage:
Yeah, my thought was to support both.  The only limitation to detecting this situation is if someone has an empty frontmatter and uses `---` later in their document.  Thats a relative low risk that if we switch to encouraging the new form and deprecate the old one, I doubt it'll be a problem.

berkus:
@freiguy1 I believe the `---` should be on separate lines

PR #531, opened 2018-10-09
Document frontmatter can now be surrounded by ---

freiguy1:
#431 
Documents can optionally start with a `---` now.  The code I added essentially checks if the document has front matter surrounded by `---` and if so, it removes the first line and then the logic for separating the front matter stays the same. I added a couple tests as well.

Questions and suggestions are welcome and encouraged! I tried a few ideas before settling on this.

### DIFF
--- a/src/cobalt_model/document.rs
+++ b/src/cobalt_model/document.rs
@@ -46,14 +46,33 @@ impl<T: frontmatter::Front> fmt::Display for DocumentBuilder<T> {
 fn split_document(content: &str) -> Result<(Option<&str>, &str)> {
     lazy_static! {
         static ref FRONT_MATTER_DIVIDE: regex::Regex = regex::Regex::new(r"---\s*\r?\n").unwrap();
-        static ref FRONT_MATTER: regex::Regex = regex::Regex::new(r"\A---\s*\r?\n[\s\S]*---\s*\r?\n").unwrap();
+        static ref FRONT_MATTER: regex::Regex = regex::Regex::new(r"\A---\s*\r?\n([\s\S]*\n)?---\s*\r?\n").unwrap();
     }
 
-    // remove divide at beginning of file if exists
-    let content = if FRONT_MATTER.is_match(content) {
-        content.splitn(2, '\n').skip(1).next().unwrap()
-    } else { content };
+    if FRONT_MATTER.is_match(content) {
+        // skip first empty string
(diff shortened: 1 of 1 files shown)


src/cobalt_model/document.rs
```diff
@@ -74,6 +74,8 @@ fn deprecated_split_front_matter(content: &str) -> Result<(Option<&str>, &str)>
         static ref FRONT_MATTER_DIVIDE: regex::Regex = regex::Regex::new(r"(\A|\n)---\s*\r?\n").unwrap();
     }
     if FRONT_MATTER_DIVIDE.is_match(content) {
+        warn!("Front matter should have a separator above and below. Using just a trailing separator is deprecated.");
```
epage: Recommendation: "Trailing separators are deprecated. We recommend frontmatters be surrounded, above and below, with `---`".

freiguy1:
How & where would I do this?
> Could you call out that this is a breaking change for anyone who had an empty frontmatter that later had --- showing up in their document?

epage:
> How & where would I do this?

Sorry I didn't specify.  I'd recommend putting it in the commit message.  My preferred style so I'm less likely to miss it is [conventional](https://github.com/conventional-changelog/conventional-changelog/blob/a5505865ff3dd710cf757f50530e73ef0ca641da/conventions/angular.md)

epage:
@freiguy1 I've not seen an update in a while (and that is fine if you are busy).  I just want to check in if there is anything you'd like help with, including completing this if you can't get to it.  Whatever works for you; I just don't want to block you using this, frustrate you with steps to get in, etc.

freiguy1:
Thanks for being so accommodating! You all have been a pleasure to work with as I'm still unfamiliar with this process. When you say

> Could you change your commit style to match conventional? It'll help remind me to include this in the changelog.

Does that mean I can change the commit messages for previous commits or just going forward? I didn't know it was possible to change previous commit messages.

freiguy1:
I must admit I got a little lost while trying to change the commit message on my first commit - so I added the BREAKING CHANGE indicator to the most recent commit (which just had some code formatting fixes).

epage:
> I must admit I got a little lost while trying to change the commit message on my first commit - so I added the BREAKING CHANGE indicator to the most recent commit (which just had some code formatting fixes).

its ok, I'll see about cleaning it up.

epage:
Thanks for doing all that work and being patient with my suggestions!

berkus:
🥇

PR #532, opened 2018-10-14
feat(pagination): core logic

Geobert:
And here is the last PR for the feature.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -72,3 +72,4 @@ unstable = []
 syntax-highlight = ["syntect"]
 sass = ['sass-rs']
 pagination-unstable = []
+
--- a/src/cobalt.rs
+++ b/src/cobalt.rs
@@ -1,3 +1,4 @@
+use std::borrow;
 use std::collections::HashMap;
 use std::fs;
 use std::io::Write;
@@ -10,10 +11,12 @@ use rss;
 
 use cobalt_model;
 use cobalt_model::files;
+use cobalt_model::permalink;
 use cobalt_model::Collection;
 use cobalt_model::{Config, SortOrder};
 use document::Document;
 use error::*;
+use pagination;
 
 struct Context {
     pub source: path::PathBuf,
@@ -101,13 +104,10 @@ pub fn build(config: Config) -> Result<()> {
     Ok(())
 }
 
-fn generate_doc(
+fn generate_collections_var(
     posts_data: &[liquid::value::Value],
-    doc: &mut Document,
     context: &Context,
-) -> Result<()> {
-    // Everything done with `globals` is terrible for performance.  liquid#95 allows us to
-    // improve this.
+) -> (borrow::Cow<'static, str>, liquid::value::Value) {
     let mut posts_variable = context.posts.attributes.clone();
     posts_variable.insert(
         "pages".into(),
@@ -118,15 +118,25 @@ fn generate_doc(
         liquid::value::Value::Object(posts_variable),
     )].into_iter()
     .collect();
+    (
+        "collections".into(),
+        liquid::value::Value::Object(global_collection),
+    )
+}
+
+fn generate_doc(
+    doc: &mut Document,
+    context: &Context,
+    global_collection: (borrow::Cow<'static, str>, liquid::value::Value),
+) -> Result<()> {
+    // Everything done with `globals` is terrible for performance.  liquid#95 allows us to
+    // improve this.
     let mut globals: liquid::value::Object = vec![
         (
             "site".into(),
             liquid::value::Value::Object(context.site.clone()),
         ),
-        (
-            "collections".into(),
-            liquid::value::Value::Object(global_collection),
-        ),
+        global_collection,
     ].into_iter()
     .collect();
     globals.insert(
@@ -162,9 +172,41 @@ fn generate_pages(posts: Vec<Document>, documents: Vec<Document>, context: &Cont
     trace!("Generating other documents");
     for mut doc in documents {
         trace!("Generating {}", doc.url_path);
-        generate_doc(&posts_data, &mut doc, context)?;
+        if doc.front.pagination.is_some() {
+            let paginators = pagination::generate_paginators(&mut doc, &posts_data)?;
+            // page 1 uses frontmatter.permalink instead of paginator.permalink
+            let mut paginators = paginators.into_iter();
+            let paginator = paginators
+                .next()
+                .expect("We detected pagination enabled but we have no paginator");
+            generate_doc(
+                &mut doc,
+                context,
+                (
+                    "paginator".into(),
+                    liquid::value::Value::Object(paginator.into()),
+                ),
+            )?;
+            for paginator in paginators {
+                let mut doc_page = doc.clone();
+                doc_page.file_path = permalink::format_url_as_file(&paginator.index_permalink);
+                generate_doc(
+                    &mut doc_page,
+                    context,
+                    (
+                        "paginator".into(),
+                        liquid::value::Value::Object(paginator.into()),
+                    ),
+                )?;
+            }
+        } else {
+            generate_doc(
+                &mut doc,
+                context,
+                generate_collections_var(&posts_data, &context),
+            )?;
+        };
     }
-
     Ok(())
 }
 
@@ -194,7 +236,11 @@ fn generate_posts(posts: &mut Vec<Document>, context: &Context) -> Result<()> {
         .unwrap_or(liquid::value::Value::Nil);
         post.attributes.insert("next".into(), next);
 
-        generate_doc(&simple_posts_data, post, context)?;
+        generate_doc(
+            post,
+            context,
+            generate_collections_var(&simple_posts_data, &context),
+        )?;
     }
 
     Ok(())
--- a/src/pagination.rs
+++ b/src/pagination.rs
@@ -1,0 +1,274 @@
+use std::cmp::Ordering;
+
+use liquid;
+
+use cobalt_model::pagination_config::Include;
+use cobalt_model::pagination_config::PaginationConfig;
+use cobalt_model::permalink;
+use cobalt_model::SortOrder;
+use document;
+use document::Document;
+use error::*;
+
+pub struct Paginator {
+    pub pages: Option<Vec<liquid::value::Value>>,
+    pub indexes: Option<Vec<String>>,
+    pub index: usize,
+    pub index_permalink: String,
+    pub previous_index: usize,
+    pub previous_index_permalink: Option<String>,
+    pub next_index: usize,
+    pub next_index_permalink: Option<String>,
+    pub first_index_permalink: String,
+    pub last_index_permalink: String,
+    pub total_indexes: usize,
+    pub total_pages: usize,
+}
+
+impl Paginator {
+    pub fn new(total_indexes: usize, total_pages: usize) -> Paginator {
+        Paginator {
+            pages: None,   // pages in current index
+            indexes: None, // list of the available indexes, use when include is tags for instance
+            index: 0,
+            index_permalink: String::new(),
+            previous_index: 0,
+            previous_index_permalink: None,
+            next_index: 0,
+            next_index_permalink: None,
+            first_index_permalink: String::new(),
+            last_index_permalink: String::new(),
+            total_indexes,
+            total_pages,
+        }
+    }
+
+    pub fn set_first_last(
+        &mut self,
+        doc: &Document,
+        config: &PaginationConfig,
+        total_pages: usize,
+    ) -> Result<()> {
+        self.first_index_permalink = doc.url_path.to_string();
+        self.last_index_permalink = interpret_permalink(&config, &doc, total_pages)?;
+        Ok(())
+    }
+
+    fn set_current_index_info(
+        &mut self,
+        index: usize,
+        all_pages: &[&liquid::value::Value],
+        config: &PaginationConfig,
+        doc: &Document,
+    ) -> Result<()> {
+        self.index = index;
+        self.pages = Some(all_pages.into_iter().map(|p| (*p).clone()).collect());
+        self.index_permalink = interpret_permalink(&config, &doc, index)?;
+        Ok(())
+    }
+
+    fn set_previous_next_info(
+        &mut self,
+        index: usize,
+        total_indexes: usize,
+        doc: &Document,
+        config: &PaginationConfig,
+    ) -> Result<()> {
+        if index > 1 {
+            // we have a previous index
+            self.previous_index_permalink = Some(interpret_permalink(&config, &doc, index - 1)?);
+            self.previous_index = index - 1;
+        }
+
+        if index < total_indexes {
+            // we have a next index
+            self.next_index = index + 1;
+            self.next_index_permalink = Some(interpret_permalink(&config, &doc, index + 1)?);
+        }
+        Ok(())
+    }
+}
+
+impl Into<liquid::value::Object> for Paginator {
+    fn into(self) -> liquid::value::Object {
+        let mut object = liquid::value::Object::new();
+        // if no pages, means we have indexes instead, `tags` like cases for exemple
+        if let Some(pages) = self.pages {
+            object.insert("pages".into(), liquid::value::Value::Array(pages));
+        }
+        // list of available indexes, in `tags` like cases
+        if let Some(indexes) = self.indexes {
+            object.insert(
+                "indexes".into(),
+                liquid::value::Value::Array(
+                    indexes.iter().map(liquid::value::Value::scalar).collect(),
+                ),
+            );
+        }
+        object.insert(
+            "index".into(),
+            liquid::value::Value::scalar(self.index as i32),
+        );
+        object.insert(
+            "index_permalink".into(),
+            liquid::value::Value::scalar(self.index_permalink),
+        );
+        if let Some(previous_index_permalink) = self.previous_index_permalink {
+            object.insert(
+                "previous_index".into(),
+                liquid::value::Value::scalar(self.previous_index as i32),
+            );
+            object.insert(
+                "previous_index_permalink".into(),
+                liquid::value::Value::scalar(previous_index_permalink),
+            );
+        }
+        if let Some(next_index_permalink) = self.next_index_permalink {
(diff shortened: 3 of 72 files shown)


Geobert:
It's ready! :champagne:

epage:
Past discussion: https://github.com/cobalt-org/cobalt.rs/pull/480

epage:
The fun of my copy buffer unexpectedly getting new stuff

src/pagination.rs
```diff
@@ -0,0 +1,312 @@
...
+        );
+        object
+    }
+}
+
+pub fn generate_paginators(
+    doc: &mut Document,
+    posts_data: &[liquid::value::Value],
+) -> Result<Vec<Paginator>> {
+    let config = doc.front.pagination.as_ref().unwrap();
```
epage: Please replace the `unwrap` with an `expect`

`unwrap`s are ok if the condition that guarentees and explains the unwrap is close to the unwrap.  As the distance grows, the need for an `expect` grows.

src/pagination.rs
```diff
@@ -0,0 +1,312 @@
...
+    let mut all_posts = posts_data.to_vec();
+    match config.include {
+        Include::All => {
+            sort_posts(&mut all_posts, &config);
+            create_all_paginators(&mut all_posts, &doc, &config)
+        }
+        Include::None => Ok(Vec::new()), // user can set `include: None`
```
epage: Don't we guarantee elsewhere that `Include::None` leads to `doc.front.pagination` being `None`?

src/pagination.rs
```diff
@@ -0,0 +1,312 @@
...
+
+fn extract_value(a: &liquid::value::Value, key: &str) -> Option<liquid::value::Scalar> {
+    a.get(&key.into())
+        .and_then(|sort_key| sort_key.as_scalar().cloned())
+}
+
+// sort posts by multiple criteria
+fn sort_posts(posts: &mut Vec<liquid::value::Value>, config: &PaginationConfig) {
+    // Boxing needed to compile, don't really understand why
+    let order: Box<Fn(liquid::value::Scalar, liquid::value::Scalar) -> Ordering> = match config
```
epage: Each closure has a unique, compiler-defined type.  What you are doing is forcing the unnameable structs into being trait objects.

Alternatively, you should be able to decay these into function pointers (`fn(liquid::value::Scalar, liquid::value::Scalar) -> Ordering`, emphasis on the lower-case `fn`) and avoid the boxing entirely.

src/pagination.rs
```diff
@@ -0,0 +1,312 @@
...
+        Include::All => {
+            sort_posts(&mut all_posts, &config);
+            create_all_paginators(&mut all_posts, &doc, &config)
+        }
+        Include::None => Ok(Vec::new()), // user can set `include: None`
+    }
+}
+
+fn extract_value(a: &liquid::value::Value, key: &str) -> Option<liquid::value::Scalar> {
+    a.get(&key.into())
```
epage: What is the `into` doing?

src/pagination.rs
```diff
@@ -0,0 +1,312 @@
...
+        .and_then(|sort_key| sort_key.as_scalar().cloned())
```
epage: Can we get away with this function returning a reference?

src/pagination.rs
```diff
@@ -0,0 +1,312 @@
...
+
+fn create_all_paginators(
+    mut all_posts: &mut Vec<liquid::value::Value>,
+    doc: &Document,
+    pagination_cfg: &PaginationConfig,
+) -> Result<Vec<Paginator>> {
+    let total_pages = all_posts.len();
+    // f32 used here in order to not lose information to ceil the result,
+    // otherwise we can lose an index
+    let total_indexes = (total_pages as f32 / pagination_cfg.per_page as f32).ceil() as usize;
```
epage: Alternatively, you could use div/rem or [`div_rev`](https://rust-num.github.io/num/num/trait.Integer.html#tymethod.div_rem)

src/pagination.rs
```diff
@@ -0,0 +1,312 @@
...
+    let mut paginators = Vec::with_capacity(total_indexes);
+    // keeping `for` loop instead of `map` as asked in PR because I need `i`
+    // anyway and it keep the code more understandable. The issue was
+    // pre-allocating  which is done with `with_capacity`
+    for i in 0..total_indexes {
```
epage: Personally I find the map version clearer.  It better expresses intent that this is mostly a pure operation.

```rust
let paginators: Result<Vec<_>> = (0..total_indexes).iter().map(|i| {
    create_paginator(
        i,
        total_indexes,
        total_pages,
        &pagination_cfg,
        &doc,
        &mut all_posts,
});
paginators
```

Unsure if that `&mut` will cause you problems though.

src/pagination.rs
```diff
@@ -0,0 +1,312 @@
...
+
+    Ok(paginator)
+}
+
+fn pagination_attributes(index_num: i32, include: Include) -> liquid::value::Object {
+    let mut attributes = liquid::value::Object::new();
+    attributes.insert("num".into(), liquid::value::Value::scalar(index_num));
+    let i: &str = include.into();
+    attributes.insert("include".into(), liquid::value::Value::scalar(i));
+    attributes
```
epage: You can do this in a more declarative way which might be easier to read.  Trying to remember how this goes
```rust
[
    ("num".into(), liquid::value::Value::scalar(index_num)),
    ("include".into(), liquid::value::Value::scalar(include.into())),
].iter().collect()
```
For exact syntax, you can find this in other places I setup attributes for permalinks

src/pagination.rs
```diff
@@ -0,0 +1,312 @@
...
+    index_num: usize,
+) -> Result<String> {
+    Ok(if index_num == 1 {
+        doc.url_path.clone()
+    } else {
+        let mut attributes = document::permalink_attributes(&doc.front, &doc.file_path);
+        let pagination_attr = pagination_attributes(index_num as i32, config.include);
+        pagination_attr.into_iter().for_each(|(k, v)| {
+            attributes.insert(k, v);
+        });
```
epage: Can't you use [`extend`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#impl-Extend%3C(K%2C%20V)%3E) to merge them `Object`s?

src/pagination.rs
```diff
@@ -0,0 +1,312 @@
...
+pub fn generate_paginators(
+    doc: &mut Document,
+    posts_data: &[liquid::value::Value],
+) -> Result<Vec<Paginator>> {
+    let config = doc.front.pagination.as_ref().unwrap();
+    let mut all_posts = posts_data.to_vec();
+    match config.include {
+        Include::All => {
+            sort_posts(&mut all_posts, &config);
+            create_all_paginators(&mut all_posts, &doc, &config)
```
epage: So the reason you use `&mut` seems to be
- We divide `all_posts` into chunks
- We move those chunks out rather than clone them,  for performance

My concerns
- mutable data makes the code harder to reason about (non-local reasoning)
- Prevents things like rayon which I want to eventually sprinkle in

One idea I have is to make `all_posts` a `Vec<&Value>`.  That might make sorting faster (just moving around a reference but at the cost of an extra indirection) and then the `clone` becomes explicit in `set_current_index_info` rather than implicit in here.
```rust
let all_posts: Vec<_> = posts_data.iter().collect();
```

When I wrap up my work with CLI-WG and Liquid refactorings, I want to come back to Cobalt and switch its architecture to an ECS.  It'd solve all the awkward architecture issues I've been having and allow us some nice reductions in cloning like we do here.

src/pagination.rs
```diff
@@ -0,0 +1,312 @@
...
+fn interpret_permalink(
+    config: &PaginationConfig,
+    doc: &Document,
+    index_num: usize,
+) -> Result<String> {
+    Ok(if index_num == 1 {
+        doc.url_path.clone()
+    } else {
+        let mut attributes = document::permalink_attributes(&doc.front, &doc.file_path);
+        let pagination_attr = pagination_attributes(index_num as i32, config.include);
```
epage: *(Observation only)*

Liquid's new `Globals` trait would help a lot here.  We'd no longer need to be creating a liquid Object.  Once I switch it to returning `Cow`, it'll be even better because we can then lazily create the items we need since most people probably don't use most of the attributes we create.

src/cobalt.rs
```diff
@@ -162,9 +172,43 @@ fn generate_pages(posts: Vec<Document>, documents: Vec<Document>, context: &Cont
...
+                context,
+                (
+                    "paginator".into(),
+                    liquid::value::Value::Object(paginator.into()),
+                ),
+            )?;
+            for paginator in paginators {
+                let mut doc_page = doc.clone();
+                doc_page.file_path =
+                    permalink::format_url_as_file(paginator.index_permalink.clone());
```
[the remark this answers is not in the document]
Geobert: without it:
```
error[E0382]: use of partially moved value: `paginator`
   --> src\cobalt.rs:200:54
    |
194 |                     permalink::format_url_as_file(paginator.index_permalink);
    |                                                   ------------------------- value moved here
...
200 |                         liquid::value::Value::Object(paginator.into()),
    |                                                      ^^^^^^^^^ value used here after move
    |
    = note: move occurs because `paginator.index_permalink` has type `std::string::String`, which does not implement the `Copy` trait
```

src/pagination.rs
Geobert: Indeed! In the builder!

src/pagination.rs
Geobert: I didn't knew about function pointers before this :D Thanks!

src/pagination.rs
Geobert: Transforms `key` into a `String`, otherwise:
```
error[E0277]: the trait bound `&liquid::value::Index: std::convert::From<&str>` is not satisfied
   --> src\pagination.rs:180:7
    |
180 |     a.get(key)
    |       ^^^ the trait `std::convert::From<&str>` is not implemented for `&liquid::value::Index`
    |
              <liquid::value::Index as std::convert::From<isize>>
              <liquid::value::Index as std::convert::From<&'a str>>
              <liquid::value::Index as std::convert::From<std::string::String>>
    = note: required because of the requirements on the impl of `std::convert::Into<&liquid::value::Index>` for `&str`
```

src/pagination.rs
Geobert: How so? 
```rust
assert_eq!(( 8).div_rem( &3), ( 2,  2));
```
when in this case I want 3 as quotient.

`div_ceil` would be what I want, but it does not exist.

src/pagination.rs
Geobert: it's complaining about `include.into()`, it can't infer the type. How do I specify the type without having the `let i: &str = include.into();` line?

src/pagination.rs
Geobert: If use this
```rust
let all_posts: Vec<_> = posts_data.iter().collect();
```
`sort_posts` will need to return a new `Vec`, am I right?

Geobert:
I've put pagination-unstable in default feature for the moment, I'll remove it on the last commit

Geobert:
uh, appveyor had a hiccup, and travis complaining about something I don't get on my desktop:
```
error[E0621]: explicit lifetime required in the type of `key`
   --> src/pagination.rs:185:16
    |
184 | fn extract_value<'a>(a: &'a liquid::value::Value, key: &str) -> Option<&'a liquid::value::Scalar> {
    |                                                   --- consider changing the type of `key` to `&'static str`
185 |     a.get(&key.into()).and_then(|sort_key| sort_key.as_scalar())
    |                ^^^^ lifetime `'static` required
error: aborting due to previous error
```

Geobert:
I don't get it :(

src/pagination.rs
epage: Might need to be a `vec` because of problems with borrows?

See https://github.com/cobalt-org/cobalt.rs/blob/master/src/cobalt.rs#L121

src/pagination.rs
epage: Liquid no longer accepts `Object`  but `Globals` (`Object` just happens to implement it).

We could have a struct that contains all the relevant root objects and implements `Globals::get`.  It then has to manually map names to struct members (I'd love to make a derive macro for this).

src/pagination.rs
epage: I'd say not to worry too hard about it right now.

epage:
> uh, appveyor had a hiccup, and travis complaining about something I don't get on my desktop:

Rebase against master.  I changed `.get()` to accept a `Scalar` and completely removed `Index`.  You'll probably have to `clone` into the `Scalar` for now.  Another thing I want to improve.  Soo many... :)

src/pagination.rs
Geobert: No offence taken :) I know you're way better than me so I tend to learn from you :D I also feel that casting is a bit overkill. I though you got a function unknown from me to do that in a one-liner. Maybe it's more efficient to test the remainder, so we stay out of floating type?

src/pagination.rs
Geobert: Oops, I've done something, let me know if it's not too ugly, otherwise I'll revert

src/pagination.rs
epage: For now, its fine to stay with float.

epage:
Try explicitly creating a `Scalar` that you pass into `get`.  I think the problem is `get` expects a reference.

src/pagination.rs
```diff
@@ -279,34 +286,19 @@ fn interpret_permalink(
...
-fn range_for_page(per_page: i32, nb_posts: usize) -> Range<usize> {
-    let nb_posts = nb_posts as i32;
+fn range_for_page(index: usize, per_page: i32, nb_posts: usize) -> Range<usize> {
+    let start = (index - 1) * per_page as usize;
+    let end = start + per_page as usize;
     // make sure `end` is not beyond capacity
-    0..if per_page < nb_posts {
-        per_page as usize
+    trace!("start = {}, end = {}, nb_posts = {}", start, end, nb_posts);
+    start..if end < nb_posts {
```
epage: could we make this a `let end = ` instead?  It feels weird having the end part of the range be an entire conditional expression

src/pagination.rs
```diff
@@ -0,0 +1,304 @@
...
+) -> Result<Vec<Paginator>> {
+    let config = doc
+        .front
+        .pagination
+        .as_ref()
+        .expect("Front should have pagination here.");
+    let all_posts: Vec<_> = posts_data.iter().collect();
+    match config.include {
+        Include::All => {
+            let all_posts = sort_posts(&all_posts, &config);
```
epage: Its fine for sort to take ` &mut`.  No reason for us to make a clone just for that.

src/pagination.rs
```diff
@@ -0,0 +1,304 @@
...
+                return cmp;
+            }
+        }
+        cmp
+    });
+    posts
+}
+
+fn create_all_paginators(
+    all_posts: &Vec<&liquid::value::Value>,
```
epage: Sorry if I wasn't clear.  I think my intent is that this would now be a slice

src/pagination.rs
```diff
@@ -0,0 +1,304 @@
...
+    doc: &Document,
+    pagination_cfg: &PaginationConfig,
+) -> Result<Vec<Paginator>> {
+    let total_pages = all_posts.len();
+    // f32 used here in order to not lose information to ceil the result,
+    // otherwise we can lose an index
+    let total_indexes = (total_pages as f32 / pagination_cfg.per_page as f32).ceil() as usize;
+    let paginators: Result<Vec<_>> = (0..total_indexes)
```
epage: I think the idea is we could do something like
```rust
let paginators: Result<Vec<_>> = all_posts.chunks(pagination_cfg.per_page).enumerate().map(|(i, chunk)| {
            create_paginator(
                i,
                total_indexes,
                total_pages,
                &pagination_cfg,
                &doc,
                &chunk,
            )
});
```
So we are passing a `&[&Value]` to create_paginator that contains only the posts relevant for that paginator.

Geobert:
Oh, as `pagination-unstable` is in front of the feature, the `no-default-features` test can't pass. What can we do?

Geobert:
Yay! This time it seems ok :D

Cargo.toml
```diff
@@ -66,9 +66,10 @@ predicates = "0.9"
 difference = "2.0"
 
 [features]
-default = ["syntax-highlight", "sass"]
+default = ["syntax-highlight", "sass", "pagination-unstable"]
```
epage: Don't forget to remove this from default

src/pagination.rs
```diff
@@ -0,0 +1,275 @@
...
+        }
+        Include::None => {
+            unreachable!("PaginationConfigBuilder should have lead to a None for pagination.")
+        }
+    }
+}
+
+fn extract_value<'a>(a: &'a liquid::value::Value, key: &str) -> Option<&'a liquid::value::Scalar> {
+    let key = liquid::value::Value::scalar(key.to_owned());
+    let key = key.as_scalar().expect("Should have scalar");
```
epage: Why not create a Scalar directly?

Geobert: When trying this:
```rust
   let key = liquid::value::Scalar::new(key);
    a.get(key).and_then(|sort_key| sort_key.as_scalar())
```

I got:
```rust
error[E0277]: the trait bound `&liquid::value::Scalar: std::convert::From<liquid::value::Scalar>` is not satisfied
   --> src\pagination.rs:181:7
    |
181 |     a.get(key).and_then(|sort_key| sort_key.as_scalar())
    |       ^^^ the trait `std::convert::From<liquid::value::Scalar>` is not implemented for `&liquid::value::Scalar`
    |
              <liquid::value::Scalar as std::convert::From<f64>>
              <liquid::value::Scalar as std::convert::From<std::string::String>>
              <liquid::value::Scalar as std::convert::From<&'a std::string::String>>
              <liquid::value::Scalar as std::convert::From<&'static str>>
            and 3 others
    = note: required because of the requirements on the impl of `std::convert::Into<&liquid::value::Scalar>` for `liquid::value::Scalar`
```

Cargo.toml
Geobert: yup, I'll do this once everything is passing the tests :)

PR #548, opened 2018-11-04
feat(pagination): pagination on tags

Geobert:
Here is my implementation of pagination on tags (https://github.com/cobalt-org/cobalt.rs/issues/395). 

* `paginator.indexes` is populated with the 1st paginator of each tag.
* I support multiple pages per tag.
* I've added `index_title` in the paginator so we know which tag we are in.

I still need to write some tests, but I'd like to start the review process as we know it will take some time reach a stable code ;)

### DIFF
--- a/src/cobalt_model/frontmatter.rs
+++ b/src/cobalt_model/frontmatter.rs
@@ -15,6 +15,7 @@ use super::pagination_config;
 use super::slug;
 
 const PATH_ALIAS: &str = "/{{parent}}/{{name}}{{ext}}";
+
 lazy_static! {
     static ref PERMALINK_ALIASES: HashMap<&'static str, &'static str> = [("path", PATH_ALIAS),]
         .iter()
--- a/src/pagination.rs
+++ b/src/pagination.rs
@@ -5,15 +5,20 @@ use liquid;
 use cobalt_model::pagination_config::Include;
 use cobalt_model::pagination_config::PaginationConfig;
 use cobalt_model::permalink;
+use cobalt_model::slug;
 use cobalt_model::SortOrder;
 use document;
 use document::Document;
 use error::*;
 
+use std::collections::HashMap;
+
+#[derive(Default, Clone, Debug)]
 pub struct Paginator {
     pub pages: Option<Vec<liquid::value::Value>>,
-    pub indexes: Option<Vec<String>>,
+    pub indexes: Option<Vec<Paginator>>,
     pub index: usize,
+    pub index_title: Option<liquid::value::Value>,
     pub index_permalink: String,
     pub previous_index: usize,
     pub previous_index_permalink: Option<String>,
@@ -32,6 +37,7 @@ impl Paginator {
             indexes: None, // list of the available indexes, use when include is tags for instance
             index: 0,
             index_permalink: String::new(),
+            index_title: None,
             previous_index: 0,
             previous_index_permalink: None,
             next_index: 0,
@@ -48,9 +54,10 @@ impl Paginator {
         doc: &Document,
         config: &PaginationConfig,
         total_pages: usize,
+        index_title: Option<&liquid::value::Value>,
     ) -> Result<()> {
         self.first_index_permalink = doc.url_path.to_string();
-        self.last_index_permalink = interpret_permalink(&config, &doc, total_pages)?;
+        self.last_index_permalink = interpret_permalink(&config, &doc, total_pages, index_title)?;
         Ok(())
     }
 
@@ -60,10 +67,12 @@ impl Paginator {
         all_pages: &[&liquid::value::Value],
         config: &PaginationConfig,
         doc: &Document,
+        index_title: Option<&liquid::value::Value>,
     ) -> Result<()> {
         self.index = index;
         self.pages = Some(all_pages.into_iter().map(|p| (*p).clone()).collect());
-        self.index_permalink = interpret_permalink(&config, &doc, index)?;
+        self.index_title = index_title.map(|i| i.clone());
+        self.index_permalink = interpret_permalink(&config, &doc, index, index_title)?;
         Ok(())
     }
 
@@ -73,17 +82,20 @@ impl Paginator {
         total_indexes: usize,
         doc: &Document,
         config: &PaginationConfig,
+        index_title: Option<&liquid::value::Value>,
     ) -> Result<()> {
         if index > 1 {
             // we have a previous index
-            self.previous_index_permalink = Some(interpret_permalink(&config, &doc, index - 1)?);
+            self.previous_index_permalink =
+                Some(interpret_permalink(&config, &doc, index - 1, index_title)?);
             self.previous_index = index - 1;
         }
 
         if index < total_indexes {
             // we have a next index
             self.next_index = index + 1;
-            self.next_index_permalink = Some(interpret_permalink(&config, &doc, index + 1)?);
+            self.next_index_permalink =
+                Some(interpret_permalink(&config, &doc, index + 1, index_title)?);
         }
         Ok(())
     }
@@ -102,9 +114,11 @@ impl Into<liquid::value::Object> for Paginator {
                 "indexes".into(),
                 liquid::value::Value::Array(
                     indexes
-                        .iter()
-                        .cloned()
-                        .map(liquid::value::Value::scalar)
+                        .into_iter()
+                        .map(|paginator| {
+                            let v: liquid::value::Object = paginator.into();
+                            liquid::value::Value::Object(v)
+                        })
                         .collect(),
                 ),
             );
@@ -117,6 +131,9 @@ impl Into<liquid::value::Object> for Paginator {
             "index_permalink".into(),
             liquid::value::Value::scalar(self.index_permalink),
         );
+        if let Some(index_title) = self.index_title {
+            object.insert("index_title".into(), index_title);
+        }
         if let Some(previous_index_permalink) = self.previous_index_permalink {
             object.insert(
                 "previous_index".into(),
@@ -170,17 +187,26 @@ pub fn generate_paginators(
     match config.include {
         Include::All => {
             sort_posts(&mut all_posts, &config);
-            create_all_paginators(&all_posts, &doc, &config)
+            create_all_paginators(&all_posts, &doc, &config, None)
         }
+        Include::Tags => create_tags_paginators(&all_posts, &doc, &config),
         Include::None => {
             unreachable!("PaginationConfigBuilder should have lead to a None for pagination.")
         }
     }
 }
 
-fn extract_value<'a>(a: &'a liquid::value::Value, key: &str) -> Option<&'a liquid::value::Scalar> {
+fn extract_value<'a>(a: &'a liquid::value::Value, key: &str) -> Option<&'a liquid::value::Value> {
     let key = liquid::value::Scalar::new(key.to_owned());
-    a.get(&key).and_then(|sort_key| sort_key.as_scalar())
+    a.get(&key)
+}
+
+fn extract_scalar<'a>(a: &'a liquid::value::Value, key: &str) -> Option<&'a liquid::value::Scalar> {
+    extract_value(a, key).and_then(|sort_key| sort_key.as_scalar())
+}
+
+fn extract_tags<'a>(value: &'a liquid::value::Value) -> Option<&'a liquid::value::Array> {
+    extract_value(value, "tags").and_then(|tags| tags.as_array())
 }
 
 // sort posts by multiple criteria
@@ -202,7 +228,7 @@ fn sort_posts(posts: &mut Vec<&liquid::value::Value>, config: &PaginationConfig)
         let keys = &config.sort_by;
         let mut cmp = Ordering::Less;
         for k in keys {
-            cmp = match (extract_value(a, &k), extract_value(b, &k)) {
+            cmp = match (extract_scalar(a, &k), extract_scalar(b, &k)) {
                 (Some(a), Some(b)) => order(a, b),
                 (None, None) => Ordering::Equal,
                 (_, None) => Ordering::Greater,
@@ -216,10 +242,71 @@ fn sort_posts(posts: &mut Vec<&liquid::value::Value>, config: &PaginationConfig)
     })
 }
 
+fn distribute_posts_by_tags<'a>(
+    all_posts: &[&'a liquid::value::Value],
+) -> Result<HashMap<String, Vec<&'a liquid::value::Value>>> {
+    let mut per_tags: HashMap<String, Vec<&'a liquid::value::Value>> = HashMap::new();
+    for post in all_posts {
+        if let Some(tags) = extract_tags(post) {
+            for tag in tags {
+                let tag = tag.as_scalar().ok_or("Should have string tags")?.to_str();
+                // add_to_tag(&tag.to_string(), post, &mut per_tags)
+                let cur_tag = per_tags.entry(tag.to_string()).or_insert(vec![]);
+                cur_tag.push(post);
+            }
+        }
+    }
+    Ok(per_tags)
+}
+
+#[derive(Default, Debug)]
+struct TagPaginators {
+    firsts_of_tags: Vec<Paginator>,
+    paginators: Vec<Paginator>,
+}
+
+fn create_tags_paginators(
+    all_posts: &[&liquid::value::Value],
+    doc: &Document,
+    pagination_cfg: &PaginationConfig,
+) -> Result<Vec<Paginator>> {
+    let mut per_tags = distribute_posts_by_tags(&all_posts)?;
+
+    // create all other paginators
+    let mut tag_paginators: TagPaginators = per_tags
+        .iter_mut()
+        .try_fold(TagPaginators::default(), |mut acc, (tag, posts)| {
+            sort_posts(posts, &pagination_cfg);
+            let cur_tag_paginators = create_all_paginators(
+                posts,
+                doc,
+                &pagination_cfg,
+                Some(&liquid::value::Value::scalar(tag.to_owned())),
+            )?;
+            acc.firsts_of_tags.push(cur_tag_paginators[0].clone());
+            acc.paginators.extend(cur_tag_paginators.into_iter());
+            Ok(acc)
+        })
+        .or_else(|e: Error| Err(e))?;
+
+    tag_paginators.firsts_of_tags.sort_unstable_by_key(|p| {
+        if let Some(ref index_title) = p.index_title {
+            Some(slug::slugify(index_title.to_str()).to_lowercase())
+        } else {
+            None
+        }
+    });
+    let mut first = Paginator::default();
+    first.indexes = Some(tag_paginators.firsts_of_tags);
+    tag_paginators.paginators.insert(0, first);
+    Ok(tag_paginators.paginators)
+}
+
 fn create_all_paginators(
     all_posts: &[&liquid::value::Value],
     doc: &Document,
     pagination_cfg: &PaginationConfig,
+    index_title: Option<&liquid::value::Value>,
 ) -> Result<Vec<Paginator>> {
     let total_pages = all_posts.len();
     // f32 used here in order to not lose information to ceil the result,
@@ -229,7 +316,15 @@ fn create_all_paginators(
         .chunks(pagination_cfg.per_page as usize)
         .enumerate()
         .map(|(i, chunk)| {
-            create_paginator(i, total_indexes, total_pages, &pagination_cfg, &doc, &chunk)
+            create_paginator(
+                i,
+                total_indexes,
+                total_pages,
+                &pagination_cfg,
+                &doc,
+                &chunk,
+                index_title,
+            )
         })
         .collect();
     paginators
@@ -242,22 +337,22 @@ fn create_paginator(
     config: &PaginationConfig,
     doc: &Document,
     all_posts: &[&liquid::value::Value],
+    index_title: Option<&liquid::value::Value>,
 ) -> Result<Paginator> {
     let index = i + 1;
     let mut paginator = Paginator::new(total_indexes, total_pages);
 
-    paginator.set_first_last(&doc, &config, total_indexes)?;
-    paginator.set_current_index_info(index, &all_posts, &config, &doc)?;
-    paginator.set_previous_next_info(index, total_indexes, &doc, &config)?;
+    paginator.set_first_last(&doc, &config, total_indexes, index_title)?;
+    paginator.set_current_index_info(index, &all_posts, &config, &doc, index_title)?;
+    paginator.set_previous_next_info(index, total_indexes, &doc, &config, index_title)?;
 
     Ok(paginator)
 }
 
-fn pagination_attributes(index_num: i32, include: Include) -> liquid::value::Object {
-    let i: &str = include.into();
+fn pagination_attributes(page_num: i32, include: String) -> liquid::value::Object {
     let attributes: liquid::value::Object = vec![
-        ("num".into(), liquid::value::Value::scalar(index_num)),
-        ("include".into(), liquid::value::Value::scalar(i)),
+        ("num".into(), liquid::value::Value::scalar(page_num)),
+        ("include".into(), liquid::value::Value::scalar(include)),
     ]
     .into_iter()
     .collect();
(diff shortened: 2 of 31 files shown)


Geobert:
Hm, I have issue with performance on debug build, on test cases it work well, but on my blog with more than a hundred posts takes a few seconds to run, I don't know what's the expectation on this kind of feature in the statically generated website world. 

It runs fast on release though.

epage:
Sorry I've had a chance to look at this yet

> Hm, I have issue with performance on debug build, on test cases it work well, but on my blog with more than a hundred posts takes a few seconds to run, I don't know what's the expectation on this kind of feature in the statically generated website world.

Have you been able to narrow down what is slowing down debug builds?  Could just be things like bounds checks that get removed but could also be something that impacts it generally.

Maybe for Christmas I'll take a break from WG-CLI and do the re-architecture which should speed things up.

src/pagination.rs
```diff
@@ -166,17 +197,30 @@ pub fn generate_paginators(
...
+    a.get(&key)
+}
+
+fn extract_scalar<'a>(a: &'a liquid::value::Value, key: &str) -> Option<&'a liquid::value::Scalar> {
+    extract_value(a, key).and_then(|sort_key| sort_key.as_scalar())
+}
+
+fn extract_tags(value: &liquid::value::Value) -> Option<&liquid::value::Array> {
+    if let Some(data) = extract_value(value, "data") {
+        extract_value(data, "tags").and_then(|tags| tags.as_array())
```
epage: `data` is meant to be a user-defined.  Part of the intent of this is the concern that we make assumptions on how a field works that the user has created, we could break their site and I'm trying to limit how we break users.

src/pagination.rs
```diff
@@ -212,10 +256,80 @@ fn sort_posts(posts: &mut Vec<&liquid::value::Value>, config: &PaginationConfig)
...
+        |tag: &str,
+         post: &'a liquid::value::Value,
+         per_tags: &mut HashMap<String, Vec<&'a liquid::value::Value>>| {
+            if per_tags.contains_key(tag) {
+                let v = per_tags
+                    .get_mut(tag)
+                    .expect("Tag's presence has been tested, can't happen!");
+                v.push(post)
+            } else {
+                per_tags.insert(tag.to_string(), vec![post]);
```
epage: Can't you simplify this with the [`entry` API](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.entry)?

src/pagination.rs
```diff
@@ -212,10 +256,80 @@ fn sort_posts(posts: &mut Vec<&liquid::value::Value>, config: &PaginationConfig)
...
+            }
+        };
+    all_posts.into_iter().for_each(|post| {
+        if let Some(tags) = extract_tags(post) {
+            tags.iter().for_each(|tag| {
+                add_to_tag(
+                    &tag.as_scalar().expect("Should have str tags").to_str(),
```
epage: Don't trust user input.  If you want to enforce this, make it a proper error.

src/pagination.rs
```diff
@@ -212,10 +256,80 @@ fn sort_posts(posts: &mut Vec<&liquid::value::Value>, config: &PaginationConfig)
     })
 }
 
+fn distribute_posts_by_tags<'a>(
+    all_posts: &[&'a liquid::value::Value],
+) -> HashMap<String, Vec<&'a liquid::value::Value>> {
```
epage: If we're already borrowing the values, should we borrow the keys?

src/pagination.rs
```diff
@@ -212,10 +256,80 @@ fn sort_posts(posts: &mut Vec<&liquid::value::Value>, config: &PaginationConfig)
...
+        if let Some(tags) = extract_tags(post) {
+            tags.iter().for_each(|tag| {
+                add_to_tag(
+                    &tag.as_scalar().expect("Should have str tags").to_str(),
+                    post,
+                    &mut per_tags,
+                )
+            })
+        } else {
+            add_to_tag("_none", post, &mut per_tags)
```
epage: I'd prefer we use a blank string as a sentinel.  We should also have a constant defined for it.

src/pagination.rs
```diff
@@ -212,10 +256,80 @@ fn sort_posts(posts: &mut Vec<&liquid::value::Value>, config: &PaginationConfig)
...
+    let mut per_tags = distribute_posts_by_tags(&all_posts);
+    let mut paginators: Vec<Paginator> = vec![];
+
+    // create all other paginators
+    let mut firsts_of_tags: Vec<_> = per_tags
+        .iter_mut()
+        .filter_map(|(tag, posts)| {
+            sort_posts(posts, &pagination_cfg);
+            if let Ok(cur_tag_paginators) =
+                create_all_paginators(posts, doc, &pagination_cfg, Some(tag))
```
epage: Why do we ignore errors if creating the pagintors fails?

src/pagination.rs
```diff
@@ -212,10 +256,80 @@ fn sort_posts(posts: &mut Vec<&liquid::value::Value>, config: &PaginationConfig)
...
+                let first_of_tag = cur_tag_paginators[0].clone();
+                paginators.extend(cur_tag_paginators.into_iter());
+                Some(first_of_tag)
+            } else {
+                None
+            }
+        }).collect();
+    firsts_of_tags.sort_unstable_by_key(|p| {
+        if let Some(ref index_title) = p.index_title {
+            Some(slug::slugify(index_title).to_lowercase())
```
epage: Our sort criteria for tags should be noted in the RFC.

src/pagination.rs
Geobert: I see, I never understood why `categories` was root field and not `tags`, now I do. So this means that this whole index on tags can't really be a thing if we don't know what the user will put in `tags`, can it? :-/

src/pagination.rs
Geobert: I use this `_none` value to create a directory `tags/_none` if a post has no tag defined in its frontmatter. With an empty string, where will I put such posts?

src/pagination.rs
Geobert: I didn't want to stop processing in case of error, but I should print some error message instead of just returning `None` indeed. Will do that!

src/pagination.rs
```diff
@@ -237,22 +359,22 @@ fn create_paginator(
...
     Ok(paginator)
 }
 
-fn pagination_attributes(index_num: i32, include: Include) -> liquid::value::Object {
-    let i: &str = include.into();
+fn pagination_attributes(index_num: i32, include: String) -> liquid::value::Object {
     let attributes: liquid::value::Object = vec![
         ("num".into(), liquid::value::Value::scalar(index_num)),
-        ("include".into(), liquid::value::Value::scalar(i)),
+        ("include".into(), liquid::value::Value::scalar(include)),
```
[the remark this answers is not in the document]
Geobert: `include` is the one defined in the RFC, on master it can only be `All` and this PR aims to support `Tags` as another `include`.

src/pagination.rs
epage: We can add `tags` to the frontmatter definition and define the behavior of them (probably its own PR?)

Users would have to migrate from their custom tags to the native tags.

src/pagination.rs
```diff
@@ -262,12 +384,24 @@ fn interpret_permalink(
...
+            format!("{}/{}", include, slug::slugify(extra_include))
+        } else {
+            doc.url_path.clone()
+        }
     } else {
         let mut attributes = document::permalink_attributes(&doc.front, &doc.file_path);
-        let pagination_attr = pagination_attributes(index_num as i32, config.include);
+        let include: &str = config.include.into();
+        let include: String = if let Some(extra_include) = extra_include {
+            format!("{}/{}", include, slug::slugify(extra_include))
```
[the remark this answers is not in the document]
Geobert: `include` is the one defined in the RFC, can only be `All` on master. This PR aims to add `Tags`. When using `Tags`, we create a directory call `tags` and `extra_include` represents the tag so we got this directory structure:
```
_site
   |__posts
   |__all <-- only if we have an index with `include: All`
   |__tags
         |__tag1
         |     |__ index.html
         |     |__ _p <-- only if nb of posts with tag1 above `per_page`
         |           |__ 2
         |                 |__ index.html
         |__tag2
```

src/pagination.rs
epage: I'll have to think more on this.  I dislike us coming up with our own thing like this.

src/pagination.rs
epage: Why would we not stop processing on errors?

src/pagination.rs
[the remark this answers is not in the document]
epage: But why are we taking on the policy to add `include` and `extra_include` to the path rather than offering them as variables to the user and they control it?

src/pagination.rs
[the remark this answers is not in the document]
epage: Oh right.

src/pagination.rs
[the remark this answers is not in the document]
epage: Now also trying to figure out what is a good name for `extra_include`.  We should probably use that same name for `index_title`.

src/pagination.rs
Geobert: Ok, I'll suspend this PR then, and write the RFC for that and the PR after.

src/pagination.rs
Geobert: Because failing on one tag does not means the others will too… but as I'm writing this, I realise it does not make sense: the generated website is not what the user is expecting no need to continue.

src/pagination.rs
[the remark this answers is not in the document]
Geobert: > But why are we taking on the policy to add include and extra_include to the path rather than offering them as variables to the user and they control it?

Do you have an example of such variable so I can try to figure out a way to do that. It's true I didn't think of this way to do it. I was coding what we wrote in the RFC :)

src/pagination.rs
Geobert: RFC written: https://github.com/cobalt-org/cobalt.rs/issues/549

src/pagination.rs
```diff
@@ -60,10 +67,12 @@ impl Paginator {
         all_pages: &[&liquid::value::Value],
         config: &PaginationConfig,
         doc: &Document,
+        extra_include: Option<&str>,
     ) -> Result<()> {
         self.index = index;
         self.pages = Some(all_pages.into_iter().map(|p| (*p).clone()).collect());
-        self.index_permalink = interpret_permalink(&config, &doc, index)?;
+        self.index_title = extra_include.map(str::to_string);
```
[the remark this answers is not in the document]
Geobert: I've added `index_title` following your comment:
> We should probably provide a variable to say what the current page's tag / category / date is (ie imagine what the header / title of the page would be, currently hidden in default.liquid).

here:
https://github.com/cobalt-org/cobalt.rs/issues/395#issuecomment-434077056

:)

src/pagination.rs
Geobert: Oh, maybe `try_fold`?

epage: `?` doesn't work well in these kind of closures which is annoying.  One trick I use to work around this is to make a function that returns `Result<Option<_>>` and then have the closure call it, swapping the Result and Option.

Alternatively, you could skip filter_map.  Personally, I try to avoid mutable operations in these things (extending `paginators`).

src/pagination.rs
Geobert: I've tried to borrow the keys as well:
```rust
fn distribute_posts_by_tags<'a>(
    all_posts: &[&'a liquid::value::Value],
) -> Result<HashMap<&'a str, Vec<&'a liquid::value::Value>>> {
    let mut per_tags: HashMap<&'a str, Vec<&'a liquid::value::Value>> = HashMap::new();
    let add_to_tag =
        |tag: &'a str,
         post: &'a liquid::value::Value,
         per_tags: &mut HashMap<&'a str, Vec<&'a liquid::value::Value>>| {
            let cur_tag = per_tags.entry(tag).or_insert(vec![post]);
            cur_tag.push(post)
        };

    for post in all_posts {
        if let Some(tags) = extract_tags(post) {
            for tag in tags {
                let tag = tag.as_scalar().ok_or("Should have string tags")?.to_str();
                add_to_tag(&tag, post, &mut per_tags)
            }
        } else {
            add_to_tag("_none", post, &mut per_tags)
        }
    }
    Ok(per_tags)
}
```
but I'm having lifetime errors:

```rust
error[E0597]: `tag` does not live long enough
   --> src\pagination.rs:276:29
    |
276 |                 add_to_tag(&tag, post, &mut per_tags)
    |                             ^^^ borrowed value does not live long enough
277 |             }
    |             - borrowed value only lives until here
    |
note: borrowed value must be valid for the lifetime 'a as defined on the function body at 260:29...
   --> src\pagination.rs:260:29
    |
260 | fn distribute_posts_by_tags<'a>(
    |                             ^^
```

src/pagination.rs
Geobert: Do you want to discuss this point here or on the RFC there https://github.com/cobalt-org/cobalt.rs/issues/549#issuecomment-438027251 ?

epage: Yeah, we should handle it in https://github.com/cobalt-org/cobalt.rs/issues/550 ... once I have a chance to look over the pagination part.

src/pagination.rs
epage:
```rust
                let tag = tag.as_scalar().ok_or("Should have string tags")?.to_str();
                add_to_tag(&tag, post, &mut per_tags)
```

Oh, the problem is the `to_str`.  It might create a String which lives to the end of the scope rather than for lifetime `'a`.  You'd have to pattern match and pull out the `Borrowed` and error if you get an `Owned` (since we define tags as only strings).

This will be improved as we move tags to the frontmatter because then the code will only accept `Vec<String>`.

Man, took me a good number of passes to spot where the lifetime mismatch was.

src/pagination.rs
Geobert: I keep this conversion open for when we found a solution and it's implemented :)

src/pagination.rs
Geobert: At least, you got it, I'm still quite lost every time lifetime is involved x)

src/pagination.rs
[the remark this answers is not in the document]
Geobert: Are we good on this?

src/pagination.rs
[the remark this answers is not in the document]
Geobert: I'm still trying to find a better name here. `extra_include` is not good at all, I agree.
`include_postfix` is more understandable, but it does not reflect the `index_title` usage.

or maybe use `index_title` here as well?

Geobert:
Only the `index_title` and the `extra_include` points left to discuss :)

src/pagination.rs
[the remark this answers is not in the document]
Geobert: I decided to use `index_title` :)

Geobert:
There's only this https://github.com/cobalt-org/cobalt.rs/pull/548#pullrequestreview-173664525 to discuss. 
I don't know what is your idea from user point of view when you're saying:
> What if instead we used Values and left it to the user? They know what they are indexing over and can make the assumption on the format of that.

src/pagination.rs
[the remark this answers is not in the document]
Geobert: I've changed it to be a Value

Geobert:
@epage : I thinks it's ready :D Let me know if you agree, if so, I'll squash the commits

src/cobalt_model/frontmatter.rs
```diff
@@ -425,11 +426,19 @@ impl FrontmatterBuilder {
...
+            } else {
+                if tags.iter().any(|x| x.trim().is_empty()) {
+                    return Err("Empty strings are not allowed in tags".into());
+                }
+                Some(tags)
             }
-        }
+        } else {
+            // default value
+            Some(vec![DEFAULT_TAGS.to_owned()])
```
epage: Two questions
- If we should have built-in support for a default tag or not.
- Is this the right place for it

So far, cobalt gives you a base set of features and you then opt-in for more.  I somewhat lean towards tags being the same way.  We should discuss this on #549

since this is generic tag behavior and not pagination, shouldn't this be its own PR?

As for if this is the right place or not: this will cause the default to be applied to generic pages and not just posts.   At the moment, that basically is us advertising a feature that does nothing.  I feel like it'd be better if it was on posts only.  Thoughts?

Geobert: > If we should have built-in support for a default tag or not.
https://github.com/cobalt-org/cobalt.rs/issues/550#issuecomment-440687841 ;)

> Is this the right place for it

Do you mean "code wise" or "PR wise"?

> I feel like it'd be better if it was on posts only. 

I didn't realise that here would impact generic pages as well. I agree it should be only for posts, at least for now, I don't see use case to tags pages. But I'm biased by my simple usage of Cobalt.

> since this is generic tag behavior and not pagination, shouldn't this be its own PR? 

I can do another PR for that if you want.

epage: > > Is this the right place for it
> Do you mean "code wise" or "PR wise"?

Code wise.  The section I meant as a follow up to this was

> As for if this is the right place or not: this will cause the default to be applied to generic pages and not just posts. At the moment, that basically is us advertising a feature that does nothing. I feel like it'd be better if it was on posts only. Thoughts?

src/pagination.rs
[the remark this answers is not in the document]
epage: Sounds good.

src/pagination.rs
epage: Now that tags are guaranteed to be strings, can't we switch to borrows?

src/pagination.rs
```diff
@@ -262,12 +365,24 @@ fn interpret_permalink(
     config: &PaginationConfig,
     doc: &Document,
     index_num: usize,
+    index_title: Option<&liquid::value::Value>,
```
epage: This is still a major open question in the RFC issue

epage:
So much easier to review when it is broken up.  Thanks for doing that!

src/cobalt_model/frontmatter.rs
Geobert: Ok, I think I understand better how defaults are managed, basically, it's when we `cobalt new` so if the user wants a default tag, he just needs to put it in the configuration.

My fix for this discussion here is to remove the default alltogether:
```rust
let tags = if let Some(tags) = tags {
            if tags.is_empty() {
                None
            } else {
                if tags.iter().any(|x| x.trim().is_empty()) {
                    return Err("Empty strings are not allowed in tags".into());
                }
                Some(tags)
            }
        } else {
            None // here the change
        };
```

src/pagination.rs
Geobert: If I got it correctly, we're going to use `index` for this, and `page_num` for the current page number in the current paginator

src/pagination.rs
Geobert: I tried:
```rust
fn distribute_posts_by_tags<'a>(
    all_posts: &[&'a liquid::value::Value],
) -> Result<HashMap<&'a str, Vec<&'a liquid::value::Value>>> {
    let mut per_tags: HashMap<&'a str, Vec<&'a liquid::value::Value>> = HashMap::new();
    let add_to_tag =
        |tag: &'a str,
         post: &'a liquid::value::Value,
         per_tags: &mut HashMap<&'a str, Vec<&'a liquid::value::Value>>| {
            let cur_tag = per_tags.entry(tag).or_insert(vec![]);
            cur_tag.push(post)
        };

    for post in all_posts {
        if let Some(tags) = extract_tags(post) {
            for tag in tags {
                let tag = tag.as_scalar().ok_or("Should have string tags")?.to_str();
                add_to_tag(&tag, post, &mut per_tags)
            }
        }
    }
    Ok(per_tags)
}
```

and ended putting `'static` asked by the compiler… is that alright? It feels wrong

Geobert: to avoid `'static` I tried to add `'a` on some functions instead:
```rust
fn create_tags_paginators<'a>(
    all_posts: &[&'a liquid::value::Value],
    doc: &Document,
    pagination_cfg: &PaginationConfig,
) -> Result<Vec<Paginator>> {
```
and 
```rust
pub fn generate_paginators<'a>(
    doc: &mut Document,
    posts_data: &'a [liquid::value::Value],
) -> Result<Vec<Paginator>> {
```

and I get this error:
```rust
error[E0495]: cannot infer an appropriate lifetime for lifetime parameter 'a in function call due to conflicting requirements
   --> src\pagination.rs:278:24
    |
278 |     let mut per_tags = distribute_posts_by_tags(&all_posts)?;
    |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
note: first, the lifetime cannot outlive the lifetime 'a as defined on the function body at 273:27...
   --> src\pagination.rs:273:27
    |
273 | fn create_tags_paginators<'a>(
    |                           ^^
    = note: ...so that the expression is assignable:
            expected &[&liquid::value::Value]
               found &[&'a liquid::value::Value]
    = note: but, the lifetime must be valid for the static lifetime...
    = note: ...so that the types are compatible:
            expected std::convert::From<&str>
               found std::convert::From<&'static str>
```

Geobert: current `pagination.rs`: https://gist.github.com/Geobert/afa5682b4ee65eb3db74b5d843a845cd

Geobert: It seems that my main issue is either not having a `From<&'a str>` to convert a `&'a str` to a `Value`, or not having a `to_string` that gives `Cow<String>` from a `Value`. (Or maybe I'm totally misunderstanding here ^^')

Geobert: I'm really stuck here, I think it's not possible, shall I keep the non borrowing version?

Geobert: I tried with `&'a String` instead of `&'a str` with no luck: https://gist.github.com/Geobert/71ca84f562e4046d5f7716ae1d1b5e3b

I got:
```rust
  Checking cobalt-bin v0.14.0 (I:\Dev\cobalt.rs)
error[E0308]: mismatched types
   --> src\pagination.rs:254:46
    |
254 |                 let cur_tag = per_tags.entry(tag).or_insert(vec![]);
    |                                              ^^^ expected reference, found enum `std::borrow::Cow`
    |
    = note: expected type `&'a std::string::String`
               found type `std::borrow::Cow<'_, str>`

```

src/cobalt_model/frontmatter.rs
epage: Defaults aren't about `cobalt new`.

When we load a page, we load
- the page's frontmatter
- the collection's frontmatter
- the global frontmatter

and merge them.  I'd recommend putting something in the defaults in `_cobalt.yml` and see how it impacts your site.  It is meant to be a simpler approach than jekyll's [default frontmatters](https://jekyllrb.com/docs/configuration/front-matter-defaults/) (no path-based configuration; maybe one day we'll need it).

I'd like to add in a debug flag to showcase what your frontmatter looks like after all of this. I think I want the ECS first since it will make it easier.

src/pagination.rs
epage: Sorry, I don't remember what I was getting at; let's not worry about this for now.

Geobert:
I think there's only this discussion https://github.com/cobalt-org/cobalt.rs/issues/550#issuecomment-443869922 left

Geobert:
@epage When you can, I still need a hand here with this static lifetime issue, thanks :) (mentioning you in case the notification did not make it)

epage:
Sorry for the delays.  The bigger or less defined the reviews, the more time I need to set aside, the more I ignore notifications unless I know I have the time which hurts in cases like this where it is small.

Overall :( sorry.

> After rebasing, I have this lifetime issue which I don't know how to deal with :-/
>
> EDIT: It seems it's because Scalar is a ScalarCow<'static> which force the 'static lifetime :-/

You probably want to look at my commit where I updated liquid to see how I did things
https://github.com/cobalt-org/cobalt.rs/pull/574/commits/b9981489ffec4e39a6e5b2e61a45aedd75911fba#diff-513943c24870c6505c01fe06179781d4R85

Particularly, `Scalar` previously would hide the fact that it allocated when given a reference.  This is no longer the case.  The caller needs to explicitly allocate.

Geobert:
Don't be sorry, Cobalt is a hobby, no pressure :D 

Thanks for the insight, I'll check that asap ^^

Geobert:
@edpage: For me, it's good to go :)

epage:
>  For me, it's good to go :)

Well, we have some design to decisions to still figure out *but* it is behind an unstable feature flag.  So I guess it can be merged now and iterated on later without "breaking".  Feel free to clean up your commits and I'll merge.  When cleaning up commits, please cleanup the message as if this is the new commit.  I rewrote your commit message last time and look what mistakes I make :)

Geobert:
> Well, we have some design to decisions to still figure out but it is behind an unstable feature flag. So I guess it can be merged now and iterated on later without "breaking". 

Agreed, anyway, by adding categories, I think I'll do some refactoring anyway :)

Geobert:
@epage : ok to go! :D

Issue #562, opened 2018-11-28
Add option to make "cobalt publish" add date prefix to filename


Geobert:
To keep files in order by chronology, it would be cool to add the `YYYY-MM-DD-` prefix to the filename on `cobalt publish`

# Proposal

Activation by a configuration field in `_cobalt.yml`:
```yml
posts:
  publish_date_in_filename: true
```
Default:
- publish_date_in_filename true for posts
- publish_date_in_filename false for pages

`cobalt publish` behavior:
- Add or update date prefix on file name
- Set or update published_date in frontmatter

## Impact on inferring `published_date` from filename

For now, we will unconditionally still infer the date from the filename.
- Not everyone will be using `cobalt publish`
- Until we have Collections (#368), we sometimes have to imitate it.

An open question for later is whether we should make the inferring of the date exclusive to this flag being enabled.  That question is irrelevant until we have Collections support.

epage:
Some interesting considerations
- This won't be applicable to all users
- This won't be applicable to all collections

Do we solve this with a command line flag, a global config option, a per-collection config option?

epage:
> Which users?

Any users which want to not have the date in their filename

Geobert:
So it's ok to have flag/option to activate it, isn't it?

epage:
Possibly.  One of my concerns is providing too  many options, overwhelming people.  This is one of the reaosns I merged `cobalt serve` and `cobalt watch`.

Geobert:
If the option are well documented, I don't see a problem, especially in this simple case. Or to simplify this, let's stick to global option only? No CLI flag.

Geobert:
added proposal to the first post

epage:
```yml
posts:
  add_date_prefix: true
```

We need to consider what name makes the intent clear.  This name doesn't help us know it is scoped to `cobalt publish` or that it is the `published_date`.

We can also look at re-framing the problem.  We could instead consider a flag to enable extracting the published_date from the filename and if it is enabled, then `cobalt publish` also prepends it.

Geobert:
> We need to consider what name makes the intent clear.

ok, what about `prepend_date_to_file_name_on_publish`?

> a flag to enable extracting the published_date from the filename and if it is enabled, then cobalt publish also prepends it.

uh? I don't get it: extracting from the filename to add it again to the filename?

Geobert:
According to the doc, we have:
> Recognized name formats include:
[...]

PR #571, opened 2018-12-07
feat(add-date-prefix-on-publish): implementation

Geobert:
addresses https://github.com/cobalt-org/cobalt.rs/issues/562

still need to figure out flag's name and it behaviour (see RFC)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ matrix:
   - env: RUSTFMT
     rust: 1.31.0  # `stable`: Locking down for consistent behavior
     install:
-    - rustup component add rustfmt-preview
+    - rustup component add rustfmt
     script:
     - cargo fmt -- --check
   - env: RUSTFLAGS="-D warnings"
--- a/src/bin/cobalt/new.rs
+++ b/src/bin/cobalt/new.rs
@@ -140,8 +140,10 @@ pub fn publish_command(matches: &clap::ArgMatches) -> Result<()> {
         .value_of("FILENAME")
         .expect("required parameters are present");
     let file = path::Path::new(file);
+    let config = args::get_config(matches)?;
+    let config = config.build()?;
 
-    publish_document(file).chain_err(|| format!("Could not publish `{:?}`", file))?;
+    publish_document(&config, file).chain_err(|| format!("Could not publish `{:?}`", file))?;
 
     Ok(())
 }
@@ -406,7 +408,33 @@ pub fn rename_document(
     Ok(())
 }
 
-pub fn publish_document(file: &path::Path) -> Result<()> {
+fn prepend_date_to_filename(
+    config: &cobalt_model::Config,
+    file: &path::Path,
+    date: &cobalt_model::DateTime,
+) -> Result<()> {
+    // avoid prepend to existing date prefix
+
+    let file_stem = cobalt_model::file_stem(file);
+    let (_, file_stem) = cobalt_model::parse_file_stem(file_stem);
+    let file_name = format!(
+        "{}{}.{}",
+        (**date).format("%Y-%m-%d-"),
+        file_stem,
+        file.extension()
+            .and_then(|os| os.to_str())
+            .unwrap_or_else(|| &config
+                .posts
+                .template_extensions
+                .get(0)
+                .expect("at least one element is enforced by config validator"))
+    );
+    trace!("`publish_date_in_filename` setting is activated, prefix filename with date, new filename: {}", file_name);
+    fs::rename(file, file.with_file_name(file_name))?;
+    Ok(())
+}
+
+pub fn publish_document(config: &cobalt_model::Config, file: &path::Path) -> Result<()> {
     let doc = cobalt_model::files::read_file(file)?;
     let doc = cobalt_model::DocumentBuilder::<cobalt_model::FrontmatterBuilder>::parse(&doc)?;
     let (front, content) = doc.parts();
@@ -417,8 +445,13 @@ pub fn publish_document(file: &path::Path) -> Result<()> {
     let doc =
         cobalt_model::DocumentBuilder::<cobalt_model::FrontmatterBuilder>::new(front, content);
     let doc = doc.to_string();
-
     cobalt_model::files::write_document_file(doc, file)?;
-
+    let posts = config.posts.clone().build()?;
+    let pages = config.pages.clone().build()?;
+    if (posts.pages.includes_file(&file) && config.posts.publish_date_in_filename)
+        || (pages.pages.includes_file(&file) && config.pages.publish_date_in_filename)
+    {
+        prepend_date_to_filename(&config, &file, &date)?;
+    }
     Ok(())
 }
--- a/src/cobalt_model/frontmatter.rs
+++ b/src/cobalt_model/frontmatter.rs
@@ -526,7 +526,7 @@ fn merge_pagination(
 }
 
 /// The base-name without an extension.  Correlates to Jekyll's :name path tag
-fn file_stem<P: AsRef<path::Path>>(p: P) -> String {
+pub fn file_stem<P: AsRef<path::Path>>(p: P) -> String {
     file_stem_path(p.as_ref())
 }
 
@@ -536,7 +536,7 @@ fn file_stem_path(p: &path::Path) -> String {
         .unwrap_or_else(|| "".to_owned())
 }
 
-fn parse_file_stem(stem: String) -> (Option<datetime::DateTime>, String) {
+pub fn parse_file_stem(stem: String) -> (Option<datetime::DateTime>, String) {
     lazy_static! {
         static ref DATE_PREFIX_REF: regex::Regex =
             regex::Regex::new(r"^(\d{4})-(\d{1,2})-(\d{1,2})[- ](.*)$").unwrap();
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -222,7 +222,7 @@ dependencies = [
  "serde_json 1.0.34 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_yaml 0.8.7 (registry+https://github.com/rust-lang/crates.io-index)",
  "syntect 3.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "toml 0.4.9 (registry+https://github.com/rust-lang/crates.io-index)",
+ "toml 0.4.10 (registry+https://github.com/rust-lang/crates.io-index)",
  "walkdir 2.2.7 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
@@ -1536,7 +1536,7 @@ dependencies = [
 
 [[package]]
 name = "toml"
-version = "0.4.9"
+version = "0.4.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "serde 1.0.80 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -1910,7 +1910,7 @@ dependencies = [
 "checksum tokio-executor 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)" = "c117b6cf86bb730aab4834f10df96e4dd586eff2c3c27d3781348da49e255bde"
 "checksum tokio-io 0.1.10 (registry+https://github.com/rust-lang/crates.io-index)" = "7392fe0a70d5ce0c882c4778116c519bd5dbaa8a7c3ae3d04578b3afafdcda21"
 "checksum tokio-reactor 0.1.6 (registry+https://github.com/rust-lang/crates.io-index)" = "4b26fd37f1125738b2170c80b551f69ff6fecb277e6e5ca885e53eec2b005018"
-"checksum toml 0.4.9 (registry+https://github.com/rust-lang/crates.io-index)" = "19782e145d5abefb03758958f06ea35f7b1d8421b534140e0238fd3d0bfd66e3"
+"checksum toml 0.4.10 (registry+https://github.com/rust-lang/crates.io-index)" = "758664fc71a3a69038656bee8b6be6477d2a6c315a6b81f7081f591bffa4111f"
 "checksum traitobject 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)" = "efd1f82c56340fdf16f2a953d7bda4f8fdffba13d93b00844c25572110b26079"
 "checksum treeline 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)" = "a7f741b240f1a48843f9b8e0444fb55fb2a4ff67293b50a9179dfd5ea67f8d41"
 "checksum typeable 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)" = "1410f6f91f21d1612654e7cc69193b0334f909dcf2c790c4826254fbb86f8887"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ ignore = "0.4"
 serde = "1.0"
 serde_yaml = "0.8"
 serde_json = "1.0"
-toml = "0.4.9"
+toml = "0.4.10"
 normalize-line-endings = "0.3.0"
 
 [dependencies.sass-rs]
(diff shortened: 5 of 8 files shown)


Geobert:
It seems that rustfmt has changed again with 1.31, Shall I update Travis and appveyor to 1.31?

epage:
> It seems that rustfmt has changed again with 1.31, Shall I update Travis and appveyor to 1.31?

Go for it.  Also, I believe rustfmt.toml should now be stable, so let's see about locking down the config with that so we get more consistent behavior.

src/bin/cobalt/new.rs
```diff
@@ -409,8 +419,26 @@ pub fn publish_document(file: &path::Path) -> Result<()> {
...
     let doc = doc.to_string();
-
     cobalt_model::files::write_document_file(doc, file)?;
 
+    if config.posts.add_date_prefix {
+        // avoid prepend to existing date prefix
+        let (_, file_name) = cobalt_model::parse_file_stem(
+            file.file_stem()
+                .map(|os| os.to_string_lossy().into_owned())
+                .unwrap_or_else(|| "".to_owned()),
```
epage: So `parse_file_stem` takes an owned value because in its current use case, it *might* be able to reuse the allocation.

In this case, we allocate and throw it away.

On the other hand, this case isn't as performance sensitive.  So maybe we leave `parse_file_stem` alone?

src/bin/cobalt/new.rs
```diff
@@ -409,8 +419,26 @@ pub fn publish_document(file: &path::Path) -> Result<()> {
     let doc =
         cobalt_model::DocumentBuilder::<cobalt_model::FrontmatterBuilder>::new(front, content);
     let doc = doc.to_string();
-
     cobalt_model::files::write_document_file(doc, file)?;
 
+    if config.posts.add_date_prefix {
```
epage: From what little I remember of this code path, don't we need to check if the user is publishing a post?

src/bin/cobalt/new.rs
```diff
@@ -409,8 +419,26 @@ pub fn publish_document(file: &path::Path) -> Result<()> {
...
+                .map(|os| os.to_string_lossy().into_owned())
+                .unwrap_or_else(|| "".to_owned()),
+        );
+
+        let file_name = format!(
+            "{}{}.{}",
+            date.prefix(),
+            file_name,
+            file.extension()
+                .map(|os| os.to_string_lossy().into_owned())
```
epage: We shouldn't make an owned value to just throw it away when we can just borrow instead

src/bin/cobalt/new.rs
```diff
@@ -409,8 +419,26 @@ pub fn publish_document(file: &path::Path) -> Result<()> {
...
+                .unwrap_or_else(|| config.posts.template_extensions[0].to_owned())
```
epage: 1. Is it possible to get in this situation? I don't remember what the CLI looks like for this.
2. We should have an error check somewhere to ensure `template_extensions` has at least one item in it

src/cobalt_model/datetime.rs
```diff
@@ -29,6 +29,10 @@ impl DateTime {
         self.0.format("%Y-%m-%d %H:%M:%S %z").to_string()
     }
 
+    pub fn prefix(&self) -> String {
```
epage: This seems too specialized to go on this struct

src/bin/cobalt/new.rs
Geobert: Where do you see the reuse of the allocation?

I'd like to avoid the `into_owned` if possible but if it's too much of a bother, as you said, it's only done in `publish`, so no big deal.

src/bin/cobalt/new.rs
Geobert: Can we publish something else than a post?

src/bin/cobalt/new.rs
epage: > Where do you see the reuse of the allocation?

This is the current use:
https://github.com/cobalt-org/cobalt.rs/blob/6502731d043d414aa9fdb41a128088898b387cfe/src/cobalt_model/frontmatter.rs#L374

We allocate a String when extracting the `file_stem`.  If there is no prepended date, then `parse_file_stem` returns the original allocation.

Hmm, `file_stem` is then just used to create a slug, if needed, and that will create a new allocation.  I guess we can switch it to only doing borrows and would be decreasing the number of allocation (by 1 in the prepended case)

src/cobalt_model/datetime.rs
Geobert: But I can't access the inside chrono date, shall I make an accessor to it?

epage: Options
1. Create a `Date` object, create a getter on `DateTime` for it, have a `Date::format`
2. Have a `format_date`
3. Use `impl ops::Deref for DateTime ` to access Chrono
4. Use `impl convert::From<DateTime> for chrono::DateTime<chrono::FixedOffset> ` to create a Chrono
5. Do our own formatting, taking advantage of ` impl chrono::Datelike for DateTime`

Geobert: Do you have a preference? I have a slight pref for solution 2, doing a `format_date`, and refactor current `format` to call this new `format_date`.

src/bin/cobalt/new.rs
Geobert: I'm really bad with lifetimes :(

```rust
file.extension()
                .map(|os| os.to_string_lossy().borrow())
```
gives me
```rust
error[E0277]: the trait bound `std::borrow::Cow<'_, str>: std::borrow::Borrow<std::string::String>` is not satisfied
   --> src\bin\cobalt\new.rs:438:48
    |
438 |                 .map(|os| os.to_string_lossy().borrow())
    |                                                ^^^^^^ the trait `std::borrow::Borrow<std::string::String>` is not implemented for `std::borrow::Cow<'_, str>`
    |
              <std::borrow::Cow<'a, B> as std::borrow::Borrow<B>>
```

How do I specify a lifetime here?

src/bin/cobalt/new.rs
Geobert: I don't know if cobalt can accept files without extension, can it?

I didn't check `template_extensions` because default value put `md` and `liquid` in it.

src/bin/cobalt/new.rs
epage: So `to_string_lossy` is returning a value.  You are then trying to take a reference to it but the value is a temp.

You'll need to do
```rust
let ext = file.extension().map(|os| os.to_string_lossy());
ext.map(|s| s.as_ref()).unwrap_or_else("")
```

src/bin/cobalt/new.rs
epage: People can override defaults

src/cobalt_model/datetime.rs
epage: `format` is doing its own formatting and shouldn't be refactored with `format_date`.

When sharing code, the thing I like to point out is if we are sharing requirements or implementation.  If it is just implementation, we shouldn't reduce "duplication".  In this case, there are no shared requirements (nothing dictates that `format`s start match our filename start).

Similarly, this policy for filenames is exclusive to filenames and might not have shared requirements with other code that might want dates.  For this reason, I don't recommend centralizing it on `DateTime`.

I'm leaning towards (3) as the simplest option.

src/bin/cobalt/new.rs
epage: Nothing in the code currently blocks it.  You could add a check or that or support both.

src/cobalt_model/datetime.rs
Geobert: about refactoring, I was thinking of something like that:
```rust
pub fn format(&self) -> String {
        self.format_date("%Y-%m-%d %H:%M:%S %z")
    }

    pub fn format_date(&self, format: &str) -> String {
        self.0.format(format).to_string()
    }
```

but I see the point of not sharing code in this case, I tends to factorise as much as possible, but I can see the issue of bloated classes if followed to strictly

src/bin/cobalt/new.rs
Geobert: If no extension is defined, what shall we do? 
another format for the file name `{}{}` instead of `{}{}.{}`?

src/bin/cobalt/new.rs
Geobert: Doesn't work neither: 
```rust
let ex = ext.map(|os| os.as_ref());
    |                               ^^        - `os` dropped here while still borrowed
    |                               |
    |                               borrowed value does not live long enough
```

src/bin/cobalt/new.rs
Geobert: To detect if it's a post, am I right if I say that I need to check the file path and compare it against the configuration `posts.dir"?

src/bin/cobalt/new.rs
epage: Maybe we should error for now?

I don't thinks can work correctly without an extension, for example, I don't think a file can be included in a collection without one.  This will also be easier to change in the future and less surprising to users (than us adding one).

We add an extension in new and rename but I think that is a CLI usability thing.  In publish, they already have a file and we shouldn't change it behind their back.

src/cobalt_model/frontmatter.rs
```diff
@@ -370,16 +370,11 @@ impl FrontmatterBuilder {
...
             let file_stem = file_stem(relpath);
-            let (file_date, file_stem) = parse_file_stem(file_stem);
-            if self.published_date.is_none() {
-                self.published_date = file_date;
-            }
-            if self.slug.is_none() {
-                let slug = slug::slugify(file_stem);
-                self.slug = Some(slug);
-            }
+            let (_, file_stem) = parse_file_stem(file_stem);
```
Geobert: >  Add or update date prefix on file name
>    Set or update published_date in frontmatter
> BREAKING CHANGE: We will not longer extract dates from pages by default

these three things made a funny effect: `parse_file_stem`'s first result (`Option<DateTime>`) is never used, shall I change the function?

Geobert:
@edpage: Apart from this https://github.com/cobalt-org/cobalt.rs/pull/571#pullrequestreview-185391836 I think I'm done here :)

Geobert:
tests are running fine on my computer… -_-

Geobert:
@edpage: this time, there's only https://github.com/cobalt-org/cobalt.rs/pull/571#pullrequestreview-185391836 left and then it will be ready

src/bin/cobalt/new.rs
```diff
@@ -406,7 +408,31 @@ pub fn rename_document(
...
+    // avoid prepend to existing date prefix
+
+    let file_stem = cobalt_model::file_stem(file);
+    let (_, file_stem) = cobalt_model::parse_file_stem(file_stem);
+    let file_name = format!(
+        "{}{}.{}",
+        (**date).format("%Y-%m-%d-"),
+        file_stem,
+        file.extension()
+            .map(|os| os
```
epage: If I'm understanding this, i think it'd be simplified by using `and_then` instead of `map.

```rust
file.extension()
  .and_then(|os| os.to_str())
  .unwrap_or_else(|| &config.posts.template_extensions[0]))
```

src/bin/cobalt/new.rs
```diff
@@ -406,7 +408,31 @@ pub fn rename_document(
...
+                .to_str()
+                .unwrap_or_else(|| &config.posts.template_extensions[0]))
+            .unwrap_or_else(|| &config.posts.template_extensions[0])
```
epage: Maybe instead of `template_extension[0]`, we should do `template_extensions.get(0).expect("at least one element is enforced by config validator")` so we can note the invariant that we are relying on that is non-local.

src/cobalt_model/frontmatter.rs
epage: I think I'm missing we the date extraction was removed?  Because we assume the date will always be there because of `cobalt publish`?  What about people doing it manually?

src/bin/cobalt/new.rs
```diff
@@ -417,8 +443,17 @@ pub fn publish_document(file: &path::Path) -> Result<()> {
     let doc =
         cobalt_model::DocumentBuilder::<cobalt_model::FrontmatterBuilder>::new(front, content);
     let doc = doc.to_string();
-
     cobalt_model::files::write_document_file(doc, file)?;
-
+    if config.posts.publish_date_in_filename {
+        let posts = config.posts.clone().build()?;
+        if posts.pages.includes_file(&file) {
```
epage: This feels backwards to me.  We should be checking what collection it is in before doing the collection-specific processing.

Yes, it causes the extra clones.  When I get to the mythical day of switching to an ECS, that should be fixed.

epage:
I'm still a little torn on what the right design is for this.

Do we need a config flag?
- The reason in favor of a config flag is this is a per-site and not per-user or per-run option

Should the config flag impact anything besides publish?
- I have a bias against a config flag that is for anything besides `cobalt build`.  
  - cargo handles this with a generic `metadata` table for external tools to have custom config in the same file

Maybe this is good enough.  *shrug*

src/bin/cobalt/new.rs
```diff
@@ -406,7 +408,31 @@ pub fn rename_document(
...
+        "{}{}.{}",
+        (**date).format("%Y-%m-%d-"),
+        file_stem,
+        file.extension()
+            .map(|os| os
+                .to_str()
+                .unwrap_or_else(|| &config.posts.template_extensions[0]))
+            .unwrap_or_else(|| &config.posts.template_extensions[0])
+    );
+    trace!("Prefix with date, new filename: {}", file_name);
```
epage: If we keep this trace, we should note the config setting that caused this so users can be helped along to disabling it if desired.

src/cobalt_model/frontmatter.rs
Geobert: Doing what manually?

`parse_file_stem`'s job it to parse the date out from the filename, but as with this PR, `publish` take total control of it (the breaking change) we don't need to keep the date if any is present (we update the date anyway)

Geobert:
As far as I'm concern, this is ready to merge if you don't have any issue with it

src/cobalt_model/frontmatter.rs
epage: Not everyone might use `publish` (manually publishing).

epage:
Ah, you did a force-push, no wonder I didn't get a notification that there were changes.

Just the one open issue.

Geobert:
> Ah, you did a force-push, no wonder I didn't get a notification that there were changes.

I didn't know that ^^'

I'll mention you next time I force push ^^

src/cobalt_model/frontmatter.rs
Geobert: Oh, you mean, manually editing `is_draft` and call `build`?

I though this:
> BREAKING CHANGE: We will not longer extract dates from pages by default

was about not doing it any more by default?

Geobert: Oh got it ^^' I'm restoring this part of the code

.travis.yml
```diff
@@ -38,7 +38,7 @@ matrix:
     script:
     - cargo check --tests
   - env: CLIPPY
-    rust: 1.31.0  # `stable`: Locking down for consistent behavior
+    rust: 1.31.0
```
epage: btw why is the comment removed?

src/cobalt_model/frontmatter.rs
epage: Weren't we talking about making this behavior conditioned on the flag in the collection?

I'm fine with us re-thinking that.  I'm just wanting to double check the plan

epage:
> Geobert **force-pushed** the Geobert:publish-add-date-prefix branch from 15001b0 to 9b90af1

Randomly came on here just in case :)

.travis.yml
Geobert: uh, I don't remember doing that ^^' I'll restore it

src/cobalt_model/frontmatter.rs
Geobert: I'm kinda lost. What I understand so far is that we stop processing filename by default and the flag is only for the `publish` command. I might misunderstood badly here ^^'

epage: Ok, my recommendation has changed on this based on recent experience and is posted on https://github.com/cobalt-org/cobalt.rs/issues/562#issuecomment-453720805.

I think that means this issue is resolved and everything is good to go?

epage:
I think this is in a state to clean up commit history + force-push so I can merge it.

Geobert:
Squash done! :D

Issue #550, opened 2018-11-10
[RFC] make tags first citizen of frontmatter


Geobert:
# Tags definition

## Current situation

At the moment, `tags` are under `data` so they are totally user defined and no rules apply to them. In order to land https://github.com/cobalt-org/cobalt.rs/pull/548, we need to make tags first citizen of the frontmatter (like the categories).

Long term, we want taxonomies (see #553) but we feel that this subset of taxonomies can be introduced without forcing breaking changes later.

### Declaration

Tags will be populated in the frontmatter as an array of strings:

```yml
---
tags: [tag1, tag2]
---
```

- If `tags` is not present, it will be `nil`
  - Alternative: have a post or global default of `["Untagged"]` that the user can override in `_cobalt.yml`. We are not doing that because generally cobalt features are opt-in.
- If the user wants a catch-all "Untagged" tag, they can set it using defaulting in the config file.
- If `tags` is `[]`, we will treat that as overriding the default frontmatter in `_cobalt.yml` to `nil`

### Liquid Variable Access

Tags can be used in the page object like `page.tags`.

If a page does not have tags, then the `tags` will not be defined on `page`.  This is in contrast to most parts of Cobalt's design which tries to not have people condition on the existence of something
- This is probably what we'll want when we expand this feature into taxonomies (see #553)

### Permalink

Unlike categories, `tags` is not an available variable in permalink.  This is from the following assumptions
- No ordering to tags for us to make permalinks from
- Tags are likely to change after the fact

### Basic Example Usage

```liquid
---
tags: [tag1, tag2]
---
{% if page.tags %} 
  {% for tag in page.tags %}
    {{ tag }}
  {% endfor %}
{% endif %}
```

## Pagination

We will support having a top-level page paginated over the list of tags.  Each tag will then have a page that is paginated over the pages marked with that tag.

To do this, we are extending #395 in the following ways:

### Activation

We are extending #395's frontmatter configuration to add a `Tags` option to `include`:
```yml
---
pagination: 
  include: Tags
---
```

#### Permalink

The pagination-specific permalink will gain the following variables
- `{{ index }}`: For tags, this will be the current tag

This will allow us to have
- Landing page that lists tags
- Additional pages for tags
- Landing page for a tag
- Additional pages for current tag

**Open Questions**:
- How do we map our current `permalink` and `pagination.permalink` to this?

### Paginator

The paginator variable will be extended with:
```yml
# paginator to each tag
[...]

PR #575, opened 2018-12-28
feat(tags): Allow overriding default tag

epage:
Because "no tags" is `None` and "no preference" is `one`, we need to
allow `[]` to also represent `no tags` and convert it to `None`.

See #550


epage:
FYI @Geobert

Geobert:
Ok, no worries, I'll rebase :)

PR #587, opened 2019-01-13
fix(publish): collection detection was failing

Geobert:
fix for https://github.com/cobalt-org/cobalt.rs/issues/562#issuecomment-453863642
sorry for the lack of concentration on this one :(

### DIFF
--- a/src/bin/cobalt/new.rs
+++ b/src/bin/cobalt/new.rs
@@ -136,14 +136,16 @@ pub fn publish_command_args() -> clap::App<'static, 'static> {
 }
 
 pub fn publish_command(matches: &clap::ArgMatches) -> Result<()> {
-    let file = matches
+    let filename = matches
         .value_of("FILENAME")
         .expect("required parameters are present");
-    let file = path::Path::new(file);
+    let mut file = env::current_dir().expect("How does this fail?");
+    file.push(path::Path::new(filename));
+    let file = file;
     let config = args::get_config(matches)?;
     let config = config.build()?;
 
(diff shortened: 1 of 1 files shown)


src/bin/cobalt/new.rs
```diff
@@ -136,14 +136,16 @@ pub fn publish_command_args() -> clap::App<'static, 'static> {
 }
 
 pub fn publish_command(matches: &clap::ArgMatches) -> Result<()> {
-    let file = matches
+    let filename = matches
         .value_of("FILENAME")
         .expect("required parameters are present");
-    let file = path::Path::new(file);
+    let mut file = env::current_dir().expect("How does this fail?");
```
epage: I'm assuming you should use [canonicalize](https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.canonicalize) instead of doing it yourself.

Otherwise the path can be "/home/geobert/mysite/posts/../_drafts/draft.md" and break prefix checks (I think).

epage: Note: you probably copied this from somewhere else like `rename`.  That is because the file doesn't exist yet which would fail a cannonicalize, I think.

Geobert: I tried canonicalize but it doesn't work as I'm ending with an extended length path (Windows specific stuff) that starts with `\\?\<Drive>:\<PATH>`. I've only copied what's done in `rename`

epage: Bleh, one of those annoying things to resolve to make paths easier.

PR #605, opened 2019-02-15
feat(pagination): indexation by categories

Geobert:
More pagination!
related RFC: https://github.com/cobalt-org/cobalt.rs/issues/395

### DIFF
--- a/src/pagination.rs
+++ b/src/pagination.rs
@@ -190,6 +190,7 @@ pub fn generate_paginators(
             create_all_paginators(&all_posts, &doc, &config, None)
         }
         Include::Tags => create_tags_paginators(&all_posts, &doc, &config),
+        Include::Categories => create_categories_paginators(&all_posts, &doc, &config),
         Include::None => {
             unreachable!("PaginationConfigBuilder should have lead to a None for pagination.")
         }
@@ -207,6 +208,10 @@ fn extract_scalar<'a>(a: &'a liquid::value::Value, key: &str) -> Option<&'a liqu
 
 fn extract_tags<'a>(value: &'a liquid::value::Value) -> Option<&'a liquid::value::Array> {
     extract_value(value, "tags").and_then(|tags| tags.as_array())
+}
+
+fn extract_categories<'a>(value: &'a liquid::value::Value) -> Option<&'a liquid::value::Array> {
+    extract_value(value, "categories").and_then(|categories| categories.as_array())
 }
 
 // sort posts by multiple criteria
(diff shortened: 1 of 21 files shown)


Geobert:
I'm high, why did I put "publish" in my second commit?? ^^'

Will disappear with the squash anyway :)

src/pagination.rs
```diff
@@ -242,6 +247,142 @@ fn sort_posts(posts: &mut Vec<&liquid::value::Value>, config: &PaginationConfig)
...
+    parent: &'b mut Category<'a>,
+    cur_idx: usize,
+    categories: &[liquid::value::Value],
+    post: &'a liquid::value::Value,
+) -> Result<()> {
+    if cur_idx <= categories.len() {
+        let cat_full_name = categories[..cur_idx].to_vec();
+        let mut cur_cat = if let Ok(idx) = parent
+            .sub_cats
+            .binary_search_by_key(&liquid_array_to_string(&cat_full_name), |c| {
```
epage: imo it'd be better to use `binary_search_by` and provide our own comparator than doing a lot of allocations in key-generation.

src/pagination.rs
```diff
@@ -242,6 +247,142 @@ fn sort_posts(posts: &mut Vec<&liquid::value::Value>, config: &PaginationConfig)
...
+        })
+        .collect();
+    s.pop(); // remove last '/'
+    s
+}
+
+// construct a hierarchy of Categories with their posts from a list of categories
+fn parse_categories_list<'a, 'b>(
+    parent: &'b mut Category<'a>,
+    cur_idx: usize,
```
epage: This function does a lot of conditional logic based on `cur_idx`.  Both for reviewing and for future maintenance of this code, could you make the intent clearer independent of the rest of the code base?

It takes time and is error prone to build up a global state in one's mind to see how all the parts interact.  Making a function make sense on its own is a big help.

src/pagination.rs
```diff
@@ -242,6 +247,142 @@ fn sort_posts(posts: &mut Vec<&liquid::value::Value>, config: &PaginationConfig)
...
+
+// walk the categories tree and construct Paginator for each node,
+// filling `pages` and `indexes` accordingly
+fn walk_categories<'a, 'b>(
+    category: &'b mut Category<'a>,
+    config: &PaginationConfig,
+    doc: &Document,
+) -> Result<Vec<Paginator>> {
+    let mut cur_cat_paginators_holder: Vec<Paginator> = vec![];
+    if !category.cat_name.is_empty() {
```
epage: Why would a category's name be empty?

src/pagination.rs
```diff
@@ -242,6 +247,142 @@ fn sort_posts(posts: &mut Vec<&liquid::value::Value>, config: &PaginationConfig)
     })
 }
 
+// Categories
+
+#[derive(Debug)]
+struct Category<'a> {
+    cat_name: liquid::value::Array,
```
epage: I'm assuming this is the "path" to this category?  I feel like a different name for the variable would be appropriate because this one, implies to me, the leaf categories' name (being singular).

src/pagination.rs
```diff
@@ -242,6 +247,142 @@ fn sort_posts(posts: &mut Vec<&liquid::value::Value>, config: &PaginationConfig)
...
+    let mut cur_cat_paginators_holder: Vec<Paginator> = vec![];
+    if !category.cat_name.is_empty() {
+        sort_posts(&mut category.posts, &config);
+        let cur_cat_paginators = create_all_paginators(
+            &category.posts,
+            doc,
+            &config,
+            Some(&liquid::value::Value::array(category.cat_name.clone())),
+        )?;
+        if !cur_cat_paginators.is_empty() {
```
epage: Under what condition would a category exist but have no paginators?

Or does this mean that all of the items for the category fit on one page?  Can't quite remember how this was set up.

Geobert:
Thank you for the review!

src/pagination.rs
Geobert: Imagine a post in `/cat1/cat2`, `cat1` has no post so after `create_all_paginators`, we have none. Hence creation of a Paginator with no post.

src/pagination.rs
Geobert: I'm confused. Here is my comparison function:
```rust
fn compare_category_path(p1: &liquid::value::Array, p2: &liquid::value::Array) -> Ordering {
    p1.iter()
        .partial_cmp(p2.iter())
        .expect("Arrays of same hierarchy level should be fully comparable")
}
```

I'm using it in a sort:
```rust
parent.sub_cats
          .sort_by(|c1, c2| compare_category_path(&c1.cat_path, &c2.cat_path));
```

And in the `binary_search_by`:
```rust
parent
            .sub_cats
            .binary_search_by(|c| 
                compare_category_path(&cat_full_path, &c.cat_path))
```

It does not work.

But if I change the order of the parameters in `binary_search_by`, it works…
How so? What am I missing?

src/pagination.rs
epage: Ah, yes.

src/pagination.rs
```diff
@@ -271,50 +271,73 @@ impl<'a> Category<'a> {
...
+//             s
+//         })
+//         .collect();
+//     s.pop(); // remove last '/'
+//     s
+// }
+
+fn compare_category_path(p1: &liquid::value::Array, p2: &liquid::value::Array) -> Ordering {
+    p1.iter()
+        .partial_cmp(p2.iter())
```
epage: Oh cool!

src/pagination.rs
```diff
@@ -545,3 +568,20 @@ fn interpret_permalink(
...
+mod test {
+    use super::*;
+
+    #[test]
+    fn compare_category_path_test() {
+        let a = vec![liquid::value::Value::scalar("A")];
+        let b = vec![liquid::value::Value::scalar("B")];
+        let ar = liquid::value::Value::array(a);
+        let ar = ar.as_array().unwrap();
+        let br = liquid::value::Value::array(b);
```
epage: btw liquid now has a macro for creating stuff in a cleaner manner.  Not too many examples of it but https://github.com/cobalt-org/liquid-rust/blob/2b01636c754a70248b1ee61b43b8369ba502414c/liquid-value/src/values.rs#L588

src/pagination.rs
epage: So `compare_category_path(&cat_full_path, &c.cat_path))` fails but `compare_category_path(&c.cat_path, &cat_full_path))` succeeds?

Geobert: Yes, which bother me a lot ^^'

src/pagination.rs
Geobert: Looking cool, but I don't know if it's the correct way to write the comparison function because of the issue above ^^'

src/pagination.rs
Geobert: Thanks!

Geobert: How to use it outside of liquid.rs? I'm seeing it being defined in liquid.rs' `macros.rs` which is not `pub` in `lib.rs`

src/pagination.rs
Geobert: By looking at https://doc.rust-lang.org/std/vec/struct.Vec.html#method.binary_search_by
It seems that the `seek` is on second position as well. 
So… this is normal?

src/pagination.rs
epage: Did you try a `macro_use` on `liquid_value`?

epage:
btw the macro stuff doesn't have to be a blocker.  All looks good.  If you want,. feel free to cleanup the commits and let me know, and I'll merge.

src/pagination.rs
Geobert: I tried like this:
```rust
#[cfg(test)]
mod test {
    use super::*;

    #[macro_use]
    use liquid::liquid_value;

    #[test]
    fn compare_category_path_test() {
      let ar = liquid_value!(["A"]);
        let br = liquid_value!(["B"]);
        assert_eq!(Ordering::Less, compare_category_path(&ar, &br));
        assert_eq!(Ordering::Greater, compare_category_path(&br, &ar));
    }
}
```

and in `lib.rs` as 
```rust
#[macro_use]
extern crate liquid;
``` 
with no success

Geobert: I don't even know if this test is worth it, it was for me to test the comparison function but it's trivial enough I think. Though?

epage: Either way.  I don't think using the macros should be blocking though.  Thats something we can fix separately from this.

Geobert:
@epage: ready to go! :D

PR #609, opened 2019-02-21
Add script for deploying from a git `post-update` hook

rotty:
I hope scripts like these are in the scope of the `cobalt.rs` repository. I have an accompanying PR for the cobalt website in preparation, but I first need to know whether this PR is acceptable, so I can reference the script from the website PR.


epage:
Sure, I'm for supporting people in wanting to do things like this.  I am a bit unsure if I want it applied to the cobalt website.  What would be the pros / cons?

The only thing I'm unsure of is the folder.   It is probably fine.

Issue #622, opened 2019-03-31
[Bug] Pagination's permalink index are embedded into {{include}}


Geobert:
Continuation of https://github.com/cobalt-org/cobalt.rs/issues/618#issuecomment-478261676

Ok, after reviewing my code, something is wrong with how I treat the pagination's
permalink (vs front's permalink, as atm, both have the same same).

I have different "1st": 
* the first paginator, which uses fron't permalink
* the page 1 of an index.

In the case of `All` they are the same.

In the case of `Tags`, we have the 1st paginator using front's permalink which has
`indexes` field. Then we have a paginated index per tag, page 1 being forced to
`tags/my_tag`, other pages being interpreted with the following caveat: `{{include}}` is
forced to `tags/my_tag`. 

In the case of `Categories`, it's the same, I embed the needed categories into
`{{include}}` which contains, for exemple, `categories/category1/subcategoryA`.

So I didn't interpret how we treat `<index>` in the permalink, I embedded it into
`include`, which is wrong.

So, let's specify it properly now that I have more knowledge on the subject ^^' and I
think I'll take @epage proposal.
# Current Proposal
## All
```
permalink: /
pagination:
  include: All
  permalink_suffix: ./{{num}}/
```

* `/index.html`
* `/all/2/index.html`

### Example 1
```
permalink: /tags_page.html
pagination:
  include: Tags
  permalink_suffix: ./{{num}}/
```

* `/tags_page.html`
* `/tags_page/my_tag/index.html`
* `/tags_page/my_tag/2/index.html`

### Example 2
```
permalink: /tags
pagination:
  include: Tags
  permalink_suffix: ./{{num}}/
```

* `/tags/index.html`
* `/tags/my_tag/index.html`
* `/tags/my_tag/2/index.html`

## Spec

* calculate `pagination_root` using `permalink`: if directory nothing to do, if file,
  remove extension
* All: 1st paginator uses `permalink`, the rest uses
  `pagination_root + index + permalink_suffix` (`index` = "all")
* Tags/Categories: meta paginator uses `permalink`, 1st paginator for one `index` uses
  `paginator_root + index`, the rest uses `paginator_root + index + permalink_suffix`
* Dates: same, meta paginator lists the biggest time unit, `index` span defined by
  `date_index`
---
# Former proposal below
## Proposal 1
If frontmatter's `permalink` is a folder, we use it as index "root". If frontmatter's
[...]

PR #631, opened 2019-04-06
fix(pagination): revamp pagination permalink management

Geobert:
fixes https://github.com/cobalt-org/cobalt.rs/issues/622


src/cobalt_model/permalink.rs
```diff
@@ -24,7 +24,7 @@ fn explode_permalink_string(permalink: &str, attributes: &liquid::value::Object)
     // Handle cases where substutions were blank
     p = p.replace("//", "/");
 
-    if p.starts_with('/') {
+    if p.starts_with('/') && p.len() > 1 {
```
Geobert: I've added this to avoid empty `doc.url_path` which broke some pagination navigation

src/pagination/mod.rs
```diff
@@ -136,23 +135,45 @@ fn interpret_permalink(
...
+    let permalink_path = std::path::Path::new(&permalink);
+    let mut pagination_root = permalink_path.extension().map_or_else(
+        || permalink.clone(),
+        |os_str| {
+            permalink
+                .trim_end_matches(&format!(".{}", os_str.to_string_lossy()))
+                .to_string()
+        },
+    );
+    // avoid "//"
```
Geobert: `//all/_p/2/` is not working, it gives us a link `all/_p/2` which leads to `http://all/_p/2/`.

Whereas `/all/_p/2/` gives us `http://localhost:3030/all/_p/2/`

Geobert:
Hm, I should have test on my blog before opening the PR, I have a big issue. I'll update the tests to reflect the bug

Geobert:
I don't understand something while serving with cobalt:
`<a href="tags/tag1/_p/2"` gives me a wrong url `http://localhost:3000/tags/tag1/tags/tag1/_p/2` (note the double `/tags/tag1/`)

whereas `<a href="/tags/tag1/_p/2"` (with a leading slash) gives me `http://localhost:3000/tags/tag1/_p/2` which is correct. 

Why is that?

src/pagination/mod.rs
epage: We do not support user-configured relative paths out of simplicity *but* the post-processed permalinks are made relative in case the generated site is not in the user's root.

I recommend doing an explicit `/{{ page.permalink}}` or even better `{{ site.base_url }}/{{ page.permalink }}`.

epage:
> `<a href="tags/tag1/_p/2"`

This is a relative path, so where it leads depends on what page in your site you are starting from.  I assume this link is being generated in `tags/tag1/index.html` and following that leads to `tags/tag1/tags/tag1/_p/2`?

src/pagination/mod.rs
Geobert: Ok I see! Make sense. Removing that!

src/cobalt_model/pagination_config.rs
```diff
@@ -121,13 +121,15 @@ impl PaginationConfigBuilder {
...
-        let permalink = permalink.unwrap_or_else(|| DEFAULT_PERMALINK.to_owned());
+        let permalink_suffix =
+            permalink_suffix.unwrap_or_else(|| DEFAULT_PERMALINK_SUFFIX.to_owned());
         let order = order.unwrap_or(SortOrder::Desc);
         let sort_by = sort_by.unwrap_or_else(|| vec![DEFAULT_SORT.to_owned()]);
         Some(PaginationConfig {
             include,
             per_page,
-            permalink,
+            front_permalink: permalink.to_string(),
```
epage: `to_string` I assume your intention is more so `to_owned`?

They both work, the latter just clarifies intent

src/cobalt_model/permalink.rs
epage: Any idea why?  If so, we should have test cases to clarify..

Geobert:
> 
> 
> > `<a href="tags/tag1/_p/2"`
> 
> This is a relative path, so where it leads depends on what page in your site you are starting from. I assume this link is being generated in `tags/tag1/index.html` and following that leads to `tags/tag1/tags/tag1/_p/2`?

yes! So it's the user's responsibility to add the leading slash again then?

src/pagination/mod.rs
```diff
@@ -136,23 +135,45 @@ fn interpret_permalink(
...
-    Ok(if page_num == 1 {
-        if let Some(index) = index {
-            let include: &str = config.include.into();
-            format!("{}/{}", include, index_to_string(&index))
-        } else {
-            doc.url_path.clone()
-        }
+    let mut attributes = document::permalink_attributes(&doc.front, &doc.file_path);
+    let permalink = permalink::explode_permalink(&config.front_permalink, &attributes)?;
+    let permalink_path = std::path::Path::new(&permalink);
```
epage: Not sure how I feel about using `std::path` for manipulating a URI path

src/pagination/mod.rs
```diff
@@ -136,23 +135,45 @@ fn interpret_permalink(
...
+        || permalink.clone(),
+        |os_str| {
+            permalink
+                .trim_end_matches(&format!(".{}", os_str.to_string_lossy()))
+                .to_string()
+        },
+    );
+    // avoid "//"
+    if pagination_root.len() == 1 && pagination_root.starts_with('/') {
+        pagination_root.remove(0);
```
epage: If `explode` strips the leading `/`, why is this needed?

epage:
> yes! So it's the user's responsibility to add the leading slash again then?

Correct.  We should fix this in the docs.

src/cobalt_model/permalink.rs
Geobert: I'll revert this, I think it's the same issue of the leading slash to be added by the user

src/pagination/mod.rs
Geobert: Is there another tool in the std lib to test for extension?

src/pagination/mod.rs
Geobert: As said above, not needed anymore :D

src/pagination/mod.rs
epage: I'd either handlroll a solution (like we've done with other url stuff) or use [`Url::path_segments`](https://docs.rs/url/1.7.2/url/struct.Url.html#method.path_segments) to at least get the last segment and then check for the extension within that.

epage:
Looks good!  Feel free to cleanup the commits and let me know when you've pushed them

Geobert:
@epage Done!
For the `std::path` issue, I'll think about on how to do it properly

PR #667, opened 2019-08-09
fix(pagination_permalink): fixes #666 and adjust tests

### LINKED ISSUE #666: Pagination suffix work change the href behaviour?
I've noticed something today as I've recompiled a cobalt at work. My pagination navigation is like https://github.com/Geobert/blog/blob/master/src/_includes/nav_pagination.liquid

I'm using `"/{{ paginator.previous_index_permalink }}"` in `href`.
Before https://github.com/cobalt-org/cobalt.rs/pull/631, when serving the blog, I got `http://localhost:3000/all/_p/2/` as link. 
After: `all/_p/2/` which leads nowhere.

So basically, I broke it… 
Opening this issue to fix it ASAP :)

Geobert:
The bug was when pagination root was empty, the `format!` call would put nothing before the /, so testing emptyness of `pagination_root` before deciding which `format!` to call.


Geobert:
uh, what's going on with sass-sys compilation?

epage:
> uh, what's going on with sass-sys compilation?

No idea.  I've been meaning to dif into it since I want to upgrade liquid but been traveling.

epage:
Merging based on Appveyor's success due to sass ... again.

Thanks for fixing this!

Geobert:
We really need a pure Rust sass solution :-/

Geobert:
https://github.com/kaj/rsass but it passes "only" 52.6% of the spec tests

PR #668, opened 2019-08-18
feat(pagination): pagination on dates

### LINKED ISSUE #618: [RFC] Pagination on dates
Continuation of https://github.com/cobalt-org/cobalt.rs/issues/395#issuecomment-443871794

# Activation

```yml
pagination:
  include: Date
  date_index: ["%Y", "%m", "%d"] // optional, this is default value
```

# Ideas

At first, I thought it will behave like the others includes: 
* create a first page following `permalink`
* create a directory named after the include (`all`, `tags`, `categories` and `date` for
  this case)
* create sub pages under this directory (`date/2018/04/02`) with a paginator that contains
  posts from this date 
* create all parent pages paginators as well (`date/2018/04`) for posts within this month,
  and another one just for the year

Then @djwf on gitter suggested that if we go to `post/2018/04/` we should get the index of
the posts of this month instead of a 404 and I quite agree.

## Question

How to approach pagination on date then. De we keep the current pagination logic of sub
directory named after the wanted include, or do we do an exception for dates?

## Leads

I have multiple ideas:
* exception for dates and generates index pages under `/posts/`
* duplicate pages in both `/date/` and `/posts/`
* create indexes under `/dates/` and redirection by generating pages with a `<meta>` under `/posts/`
* create indexes under `/posts/` and redirection by generating pages with a `<meta>` under `/dates/`
* document how to configure your web server to redirect properly

Thoughts?

Geobert:
It took me a while, but here is my implementation proposal to paginate on dates.
Fixes #618


Geobert:
Of course, failing on `sass-sys` on Linux

src/cobalt_model/pagination_config.rs
```diff
@@ -125,13 +144,16 @@ impl PaginationConfigBuilder {
             permalink_suffix.unwrap_or_else(|| DEFAULT_PERMALINK_SUFFIX.to_owned());
         let order = order.unwrap_or(SortOrder::Desc);
         let sort_by = sort_by.unwrap_or_else(|| vec![DEFAULT_SORT.to_owned()]);
+        let mut date_index = date_index.unwrap_or_else(|| DEFAULT_DATE_INDEX.to_vec());
+        date_index.sort_unstable(); // keep the date indexes in logical order
```
epage: Sort or error?

src/cobalt_model/pagination_config.rs
```diff
@@ -156,6 +179,7 @@ impl Default for PaginationConfig {
             front_permalink: Default::default(),
             order: SortOrder::Desc,
             sort_by: vec![DEFAULT_SORT.to_owned()],
+            date_index: DEFAULT_DATE_INDEX.to_vec(), // TODO default value to define
```
epage: Just creating this to track the TODO

src/pagination/dates.rs
```diff
@@ -0,0 +1,159 @@
...
+            field,
+            posts: vec![],
+            sub_date: vec![],
+        }
+    }
+}
+
+fn extract_published_date<'a>(value: &'a liquid::value::Value) -> Option<DateTime> {
+    if let Some(published_date) = extract_scalar(&value, "published_date") {
+        DateTime::parse(published_date.to_str())
```
epage: Why are we parsing it?  Shouldn't we call [`to_date`](https://github.com/cobalt-org/liquid-rust/blob/master/liquid-value/src/scalar.rs#L113)?

src/cobalt_model/pagination_config.rs
Geobert: It's an intended sort, so if the user put month, day, year, we keep a logical order year, month, day to distribute the posts into.

src/cobalt_model/pagination_config.rs
Geobert: Oh, this TODO was to confirmed the value I chose: `Year, Month` as default. If that's ok with you, I'll remove the TODO

src/pagination/dates.rs
```diff
@@ -0,0 +1,159 @@
...
+    if not_found {
+        let mut holder = DateIndexHolder::new(value, Some(*wanted_field));
+        holder.posts.push(post);
+        date_holder.sub_date.push(holder);
+    }
+}
+
+fn get_date_field_value(date: &DateTime, field: &DateIndex) -> u32 {
+    match field {
+        DateIndex::Year => date.year() as u32, // let's say we don't support negative year…
```
epage: Should we at least turn this comment into a panic?

tests/target/pagination_dates/dates/2016/index.html
```diff
@@ -0,0 +1,43 @@
...
+  <div>Date: 2016</div>
+  
+    <div>Sub-dates:</div>
+  
+
+
+All years:
+<ul>
+  
+    <li><a href="/dates/2016/1/">1 (1)</a>
```
epage: What about supporting zero-padding?  At least for me, I prefer it.

epage:
Thanks for your work on this!

src/cobalt_model/pagination_config.rs
epage: But what does it mean for the user to specify them out of order?  Are we just changing what they said, behind their backs?  Thats something I'm not normally a fan of.

src/cobalt_model/pagination_config.rs
epage: Thats a good default, so feel free to remove the TODO.

src/pagination/dates.rs
Geobert: Because I didn't know its existence ^^ gonna change that

src/pagination/dates.rs
Geobert: You're totally right about this, going to do that

tests/target/pagination_dates/dates/2016/index.html
Geobert: I prefer it too, but that's why I asked about the `date` filter, I was counting on it to do that as in the code, it's all `u32`. Shall I try to do it in cobalt directly?

epage: `date` filter won't help with the paths being generated, that is purely in cobalt.

src/cobalt_model/pagination_config.rs
Geobert: That's my issue: I can't think of a meaning to having them out of order. So I've done this to avoid friction but maybe an error is more appropriate, so there's no "magic"?

epage: We can always remove an error without impacting people.  Changing behavior might have an impact.

tests/target/pagination_dates/dates/2016/index.html
Geobert: True, I'll try to figure it out

Geobert:
I've applied your comments, let me know what you think :)

src/pagination/dates.rs
```diff
@@ -46,17 +46,26 @@ pub fn create_dates_paginators(
...
 }
 
-fn date_fields_to_array(date: &Vec<u32>) -> liquid::value::Array {
-    date.iter()
-        .map(|d| liquid::value::Value::scalar(*d as i32))
-        .collect()
+fn format_date_holder(d: &DateIndexHolder) -> liquid::value::Value {
+
+liquid::value::Value::scalar(
+    match d.field.expect("Should not be called with the root DateIndexHolder") {
```
epage: I think I'd prefer this simplified
```
let field = d.field.expect(...);
let formatted = match field {
};
liquid::value::Value::scalar(formatted)
```

epage:
Just the minor style issue and this all looks good! Thanks!

Geobert:
It's done and commit squashed as well :)

Issue #671, opened 2019-08-31
[RFC] sort by weight


Geobert:
Initially, this was first talked about in https://github.com/cobalt-org/cobalt.rs/issues/395 so we can sort the docs by a "weight" field, in order to pin a post for example. 

I'm going to start the implementation of this, but before starting anything, I'm opening this RFC to be sure we are all on the same page on this :)

# My plan

I was going to introduce a new field `weight` (an `i32`) in the frontmatter.

After that, the work on sorting is already done in the pagination feature: https://github.com/cobalt-org/cobalt.rs/blob/ef83cfec90c9cef77f0ac508a7374ce4f4387a1b/src/pagination/mod.rs#L74

epage:
Sounds good.  The only question I have is what the default weight should be.

Geobert:
I was thinking to `0`, using `i32` we are in the middle of all the possible values

PR #678, opened 2019-09-02
feat(pagination): sort by weight

Geobert:
added `weight` field in frontmatter in order to tackle https://github.com/cobalt-org/cobalt.rs/issues/671

### DIFF
--- a/src/cobalt_model/pagination_config.rs
+++ b/src/cobalt_model/pagination_config.rs
@@ -4,8 +4,11 @@ use std::vec::Vec;
 use super::SortOrder;
 
 pub const DEFAULT_PERMALINK_SUFFIX: &str = "{{num}}/";
-pub const DEFAULT_SORT: &str = "published_date";
+
 pub const DEFAULT_PER_PAGE: i32 = 10;
+lazy_static! {
+    static ref DEFAULT_SORT: Vec<String> = vec!["weight".to_string(), "published_date".to_string()];
+}
 
 lazy_static! {
     static ref DEFAULT_DATE_INDEX: Vec<DateIndex> = vec![DateIndex::Year, DateIndex::Month];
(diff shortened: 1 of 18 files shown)


epage:
Should the default sort_by be updated to be weight, then published_date?

Geobert:
> 
> 
> Should the default sort_by be updated to be weight, then published_date?

This is a great idea, I'll add this to this PR right now :)

src/cobalt_model/pagination_config.rs
```diff
@@ -4,8 +4,11 @@ use std::vec::Vec;
 use super::SortOrder;
 
 pub const DEFAULT_PERMALINK_SUFFIX: &str = "{{num}}/";
-pub const DEFAULT_SORT: &str = "published_date";
+
 pub const DEFAULT_PER_PAGE: i32 = 10;
+lazy_static! {
```
epage: Reminds me that I need to go back sometime and move from lazy_static to once_cell

src/cobalt_model/pagination_config.rs
```diff
@@ -150,7 +153,7 @@ impl PaginationConfigBuilder {
         let permalink_suffix =
             permalink_suffix.unwrap_or_else(|| DEFAULT_PERMALINK_SUFFIX.to_owned());
         let order = order.unwrap_or(SortOrder::Desc);
-        let sort_by = sort_by.unwrap_or_else(|| vec![DEFAULT_SORT.to_owned()]);
+        let sort_by = sort_by.unwrap_or_else(|| DEFAULT_SORT.to_vec());
```
epage: I think `clone` would better convey intent but oh well

Geobert: I can change it, I love having your opinion on this kind of stuff, it teach me to better express the intent of my code :)

src/cobalt_model/pagination_config.rs
Geobert: Just discovering `once_cell`, looks really great!

Issue #681, opened 2019-09-02
Should we get rid of every "pages" page being an index?


epage:
Before #395, we gave every page in the "pages" collection a "collections.posts" so people could generate an index.

We now have an opt-in solution with #395, so it seems like an expensive "give it to everyone" option can/should be removed.

epage:
If we make this breaking change
- We'd need to have a clearly documented way to get the same behavior in an alternative form
- Consider if/how we can help people migrate with tools
- Consider what a transition period might look like

Geobert:
Great that you opened this ticket. Before sleeping, I had this conclusion as well: the pagination/indexation feature is quite separated from `collections` which force a migration.

So we either provide means to migrate, or find a way to integrate `paginator` and `collections`.

But they seems to be conflictual by nature.

epage:
> But they seems to be conflictual by nature.

Which parts are conflicting?  Just that different names are provided?
- For "pages", we could enable the pagintor by default
- We could do something like "if the paginator is enabled, provide an alias in `collections`.
  - Maybe provide an opt-out for testing

This might conslidate the code paths for now and give people a transition route (disable the settings).  We can then eventually flip the switch on the defaults.  Later, we remove the `collections` alias.

Geobert:
Not conflictual, I'm still a bit confused by collections as I don't use them ^^' 

But still separated. If we create a collection, we cannot paginate it.

epage:
I'm not referring to the custom Collections feature but that you could create an index page, without pagination, using [`collections.posts`](https://cobalt-org.github.io/docs/variables/).  

We can make `collections.posts` an alias for the paginated index as part of the transition

Geobert:
I agree.

What other default collections to we have? `collection.pages`?

epage:
We don't offer a `collection.pages` variable even though that is the only other collection.

Geobert:
Ok, I'll do a PR to add the alias. Shall I add a warning of some kind?

epage:
I considered a warning but I don't want to unconditionally warn and we can't detect the usage of `collections.posts`.

If we make  compatibility configurable, we can warn on someone having it enabled.

Geobert:
And if compatibility is configurable and disabled, we don't alias?

epage:
Correct, we don't automatically enable the pagination and alias it.

Geobert:
Do you think the aliasing flag should be global or per index only?

epage:
[...]

PR #683, opened 2019-09-05
feat(pagination): aliasing paginator to collections.posts

Geobert:
For https://github.com/cobalt-org/cobalt.rs/issues/681

### DIFF
--- a/src/cobalt.rs
+++ b/src/cobalt.rs
@@ -208,40 +208,34 @@ fn generate_pages(posts: Vec<Document>, documents: Vec<Document>, context: &Cont
         let paginators = pagination::generate_paginators(&mut doc, &posts_data)?;
         // page 1 uses frontmatter.permalink instead of paginator.permalink
         let mut paginators = paginators.into_iter();
-        if let Some(paginator) = paginators.next() {
-            if doc.front.pagination_compat {
-                warn!(
-                    "`collections.posts` is deprecated. Please transition to `paginator` and then \
-                     disable support for `collections.posts` by setting `pagination_compat` to \
-                     `false`"
-                );
-            }
+        let paginator = paginators
+            .next()
+            .expect("Should have at least one paginator.");
+        if doc.front.pagination_compat {
+            warn!(
+                "`collections.posts` is deprecated. Please transition to `paginator` and then \
+                 disable support for `collections.posts` by setting `pagination_compat` to \
+                 `false`"
+            );
+        }
+        generate_doc(
+            &mut doc,
+            context,
+            (
+                "paginator".into(),
+                liquid::value::Value::Object(paginator.into()),
+            ),
+        )?;
+        for paginator in paginators {
+            let mut doc_page = doc.clone();
+            doc_page.file_path = permalink::format_url_as_file(&paginator.index_permalink);
             generate_doc(
-                &mut doc,
+                &mut doc_page,
                 context,
                 (
                     "paginator".into(),
                     liquid::value::Value::Object(paginator.into()),
                 ),
-            )?;
-            for paginator in paginators {
-                let mut doc_page = doc.clone();
-                doc_page.file_path = permalink::format_url_as_file(&paginator.index_permalink);
-                generate_doc(
-                    &mut doc_page,
-                    context,
-                    (
-                        "paginator".into(),
-                        liquid::value::Value::Object(paginator.into()),
-                    ),
-                )?;
-            }
-        } else {
-            // Create empty `collections.posts.pages`
-            generate_doc(
-                &mut doc,
-                context,
-                generate_collections_var(&posts_data, &context),
             )?;
         }
     }
--- a/src/cobalt.rs
+++ b/src/cobalt.rs
@@ -133,15 +133,40 @@ fn generate_doc(
 ) -> Result<()> {
     // Everything done with `globals` is terrible for performance.  liquid#95 allows us to
     // improve this.
-    let mut globals: liquid::value::Object = vec![
-        (
-            "site".into(),
-            liquid::value::Value::Object(context.site.clone()),
-        ),
-        global_collection,
-    ]
-    .into_iter()
-    .collect();
+    let mut globals: liquid::value::Object =
+        if doc.front.pagination.is_some() || doc.front.pagination_compat {
+            warn!(
+            "`collections.posts` is deprecated. Please transition to `paginator` and then disable \
+            support for `collections.posts` by setting `pagination_compat` to `false`"
+            );
+            let collections_posts: liquid::value::Object =
+                vec![("posts".into(), global_collection.1.clone())]
+                    .into_iter()
+                    .collect();
+            vec![
+                (
+                    "site".into(),
+                    liquid::value::Value::Object(context.site.clone()),
+                ),
+                (
+                    "collections".into(),
+                    liquid::value::Value::Object(collections_posts),
+                ),
+                global_collection,
+            ]
+            .into_iter()
+            .collect()
+        } else {
+            vec![
+                (
+                    "site".into(),
+                    liquid::value::Value::Object(context.site.clone()),
+                ),
+                global_collection,
+            ]
+            .into_iter()
+            .collect()
+        };
     globals.insert(
         "page".into(),
         liquid::value::Value::Object(doc.attributes.clone()),
--- a/src/cobalt.rs
+++ b/src/cobalt.rs
@@ -133,40 +133,35 @@ fn generate_doc(
 ) -> Result<()> {
     // Everything done with `globals` is terrible for performance.  liquid#95 allows us to
     // improve this.
-    let mut globals: liquid::value::Object =
-        if doc.front.pagination.is_some() || doc.front.pagination_compat {
-            warn!(
-            "`collections.posts` is deprecated. Please transition to `paginator` and then disable \
-            support for `collections.posts` by setting `pagination_compat` to `false`"
-            );
-            let collections_posts: liquid::value::Object =
-                vec![("posts".into(), global_collection.1.clone())]
-                    .into_iter()
-                    .collect();
-            vec![
-                (
-                    "site".into(),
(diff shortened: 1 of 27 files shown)


src/cobalt.rs
```diff
@@ -133,15 +133,40 @@ fn generate_doc(
...
-        (
-            "site".into(),
-            liquid::value::Value::Object(context.site.clone()),
-        ),
-        global_collection,
-    ]
-    .into_iter()
-    .collect();
+    let mut globals: liquid::value::Object =
+        if doc.front.pagination.is_some() || doc.front.pagination_compat {
```
epage: I feel like this needs to be an `&&` and not an `||`.  If someone has a paginator configured and doesn't have compat set, then we shouldn't warn people / add `collections`

src/cobalt.rs
```diff
@@ -133,15 +133,40 @@ fn generate_doc(
...
+            vec![
+                (
+                    "site".into(),
+                    liquid::value::Value::Object(context.site.clone()),
+                ),
+                global_collection,
+            ]
+            .into_iter()
+            .collect()
+        };
```
epage: (picked arbitrary place because GH limitations)

Shouldn't this change be leading us to no longer need `generate_collections_var`?

src/cobalt.rs
Geobert: Duh, you're right, I didn't do any test yet, it was to show you how I was going to do the aliasing. Good catch, though!

src/cobalt.rs
Geobert: If the pagination is not activated, we still need it, isn't it?

Geobert: I understand now (https://github.com/cobalt-org/cobalt.rs/issues/681#issuecomment-528568537)

Geobert: `generate_collections_var` is used in `generate_pots` and I still use it to manage the case where no paginator was created (because of no posts, test `data_files` was failing)

Geobert:
From now on, I need to activate the feature flag by default, otherwise, it will never pass the CI, can I do it? (No merge to master yet, we still need to agree that the solution is ok and answered to the 2 other issues)

Geobert:
Ignore this updated PR. I was tired and I still need to get rid of generate_collections_var, you were right ^^

src/cobalt.rs
```diff
@@ -133,15 +133,35 @@ fn generate_doc(
...
-            "site".into(),
-            liquid::value::Value::Object(context.site.clone()),
-        ),
-        global_collection,
-    ]
-    .into_iter()
-    .collect();
+    let mut globals: liquid::value::Object = if doc.front.pagination_compat {
+        let collections_posts: liquid::value::Object =
+            vec![("posts".into(), global_collection.1.clone())]
```
epage: Note to self: double check that what we are inserting in here is compatible with what we originally put in here

src/cobalt.rs
```diff
@@ -181,40 +201,43 @@ fn generate_pages(posts: Vec<Document>, documents: Vec<Document>, context: &Cont
...
     for mut doc in documents {
         trace!("Generating {}", doc.url_path);
-        if doc.front.pagination.is_some() {
-            let paginators = pagination::generate_paginators(&mut doc, &posts_data)?;
-            // page 1 uses frontmatter.permalink instead of paginator.permalink
-            let mut paginators = paginators.into_iter();
-            let paginator = paginators
-                .next()
-                .expect("We detected pagination enabled but we have no paginator");
+        if doc.front.pagination.order == SortOrder::None {
```
epage: Note to self: double check the change away from an `Option`

src/cobalt.rs
```diff
@@ -181,40 +201,43 @@ fn generate_pages(posts: Vec<Document>, documents: Vec<Document>, context: &Cont
...
+            // "inherit" sort order from frontmatter if none set in `pagination` section
+            doc.front.pagination.order = context.posts.order;
```
epage: Might not be feasible but can we implement this over in `config.rs` where we merge in the posts-defaults?

src/cobalt.rs
epage: @Geobert could you explain your theory of ops here?  It seems like the paginator is always enabled for all pages, regardless of what collection they are in (posts or pages) when
- I thought pagination is only supported (and so should error otherwise) in pages
- The goal was to be compatible while allowing people to opt-out.  It doesn't look like this provides an opt-out mechanism.

My thought for how this would have been implemented is to modify the `pages` specific defaults to set an active paginator over `all`.  That makes it easy for people to opt-out on a global level while then opting in on a per-page level.

tests/fixtures/excerpts_CRLF/index.liquid
```diff
@@ -1,5 +1,7 @@
 ---
 layout: default.liquid
+pagination:
```
epage: Why is this test modified?  I would assume the goal is that no existing test should need changes because our goal is that this should go unnoticed by users.

tests/fixtures/pagination_compat/index.liquid
```diff
@@ -0,0 +1,25 @@
+---
+permalink: /
+layout: default.liquid
+pagination:
+  include: All
```
epage: What is the goal of this test?

src/cobalt.rs
epage: @Geobert could you explain your theory of ops here?  It seems like the paginator is always enabled for all pages, regardless of what collection they are in (posts or pages) when
- I thought pagination is only supported (and so should error otherwise) in pages
- The goal was to be compatible while allowing people to opt-out.  It doesn't look like this provides an opt-out mechanism.

My thought for how this would have been implemented is to modify the `pages` specific defaults to set an active paginator over `all`.  That makes it easy for people to opt-out on a global level while then opting in on a per-page level.

tests/fixtures/excerpts_CRLF/index.liquid
Geobert: Because the default `per_page` value is 10 and we have 12 posts in this test. So 2 posts were paginated away and the test failed.

tests/fixtures/pagination_compat/index.liquid
Geobert: To make sure that we can use `collections.posts`. Before this PR, if we have pagination enabled, `collections` was not created.

src/cobalt.rs
Geobert: 1. You're right, I messed up, going back to that code.
2. Opt-out is done like this in `_cobalt.yml`:
```yml
pages:
  default:
    pagination_compat: false
```

src/cobalt.rs
Geobert: It's not possible in `config.rs` but I can make it happen in `document.rs` by modifying `Document::parse`: instead of `default_front: cobalt_model::FrontmatterBuilder`, I can pass the whole `Collection` so I have both `default_front` extracted from it and the `posts.order` flag to pass to `FrontmatterBuilder` somehow (either a `set_pagination_order` function or a parameter in `build()`).

Shall I do it?

src/cobalt.rs
Geobert: Trying to sort 1., I'm wondering if the `pagination_compat` flag in the frontmatter is a good idea. Maybe only a global flag in `PageConfig` would make more sense?

tests/fixtures/pagination_compat/index.liquid
epage: Unless legacy support is on, there shouldn't be a `collections` though

src/cobalt.rs
epage: Why can't it be done in `config.rs`?

src/cobalt.rs
epage: I'm mixed.
- Global flag would make sense for the `collections`, previous, and next support in posts
- Somehow we still need to pass the result of that global flag to the processing of each page
- Would it be useful for someone to disable it for all pages and then opt-back-in on a page-by-page basis until they migrate?

tests/fixtures/pagination_compat/index.liquid
Geobert: Yes, but you said it was activated by default. 
This test ensure that if the legacy support is disable, we don't have `collections`: https://github.com/cobalt-org/cobalt.rs/blob/5d3a3bfb882325fef9f084a4041760892a60acfe/tests/mod.rs#L328

epage: Ok, so this is just to ensure `collections.posts` didn't break?  Don't we already have tests for that?

Geobert: >  Before this PR, if we have pagination enabled, collections was not created.

That's why I have this test.
It's for the case "when pagination is enabled, we still have access to collections.posts" .
It wasn't the case before this whole transition discussion: if pagination was enabled, collections.posts was not created for pages.

src/cobalt.rs
Geobert: I can't find my way while reading the file. As far as I understand it, I need access to `DocumentBuilder` which can't be done in `config.rs`?

I might be  wrong though

src/cobalt.rs
Geobert: > Somehow we still need to pass the result of that global flag to the processing of each page

The flag would be in `context.pages`, wouldn't it?

> Would it be useful for someone to disable it for all pages and then opt-back-in on a page-by-page basis until they migrate?

As the legacy support is on by default, I can't see why someone would deactivate it before migrating all to `paginator`.

Geobert: Sorry for the late update, holidays are finished so less time (and I'm trying to get involved in rust-analyzer as well) Shall I proceed with a global flag?

src/cobalt.rs
Geobert: Shall I proceed on moving it in `document.rs` or do we keep it here?

Geobert: @epage (I mention you, as it begins to be long without answer, maybe you didn't get the notif, like the other time :) )

Issue #452, opened 2018-05-19
Automatic Generation Of sitemap.xml


tglman:
Hi,

Would be cool if cobalt could generate the [sitemap.xml](https://en.wikipedia.org/wiki/Sitemaps) with the references to all the pages and resources, in a similar way on how is today is generate the rss.xml, be aware that already exist a crate for generate sitemap [here](https://crates.io/crates/sitemap).

Regards

epage:
Closed by #684

PR #684, opened 2019-09-06
add support for sitemap generation

tglman:
This implement the sitemap support with the use of sitemap crate, as wrote in this issue: https://github.com/cobalt-org/cobalt.rs/issues/452

### DIFF
--- a/src/cobalt.rs
+++ b/src/cobalt.rs
@@ -9,6 +9,7 @@ use jsonfeed;
 use jsonfeed::Feed;
 use liquid;
 use rss;
+use sitemap::writer::SiteMapWriter;
 
 use crate::cobalt_model;
 use crate::cobalt_model::files;
@@ -29,6 +30,7 @@ struct Context {
     pub liquid: cobalt_model::Liquid,
     pub markdown: cobalt_model::Markdown,
     pub assets: cobalt_model::Assets,
+    pub sitemap: Option<String>,
 }
 
 impl Context {
@@ -43,6 +45,7 @@ impl Context {
             liquid,
             markdown,
             assets,
+            sitemap,
         } = config;
 
         let pages = pages.build()?;
@@ -65,6 +68,7 @@ impl Context {
             liquid,
             markdown,
             assets,
+            sitemap,
         };
         Ok(context)
     }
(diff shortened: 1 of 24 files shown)


src/cobalt.rs
```diff
@@ -450,3 +464,37 @@ fn create_jsonfeed(
...
 }
+fn create_sitemap(
+    path: &str,
+    dest: &path::Path,
+    collection: &Collection,
+    documents: &[Document],
+    collection_pages: &Collection,
+    documents_pages: &[Document],
+) -> Result<()> {
+    let sitemap_path = dest.join(path);
```
epage: From an API perspective (even if internal) I think I'd prefer this to be in the caller

src/cobalt_model/config.rs
```diff
@@ -67,7 +67,7 @@ impl PageConfig {
             order: collection::SortOrder::None,
             rss: None,
             jsonfeed: None,
-            base_url: None,
+            base_url: site.base_url.clone(),
```
epage: I'm confused as to why the code didn't do this in the first place.  I'll need to dig into this

src/document.rs
```diff
@@ -264,6 +264,21 @@ impl Document {
...
+    pub fn to_sitemap<T: std::io::Write>(
+        &self,
+        root_url: &str,
+        writer: &mut sitemap::writer::UrlSetWriter<T>,
+    ) -> Result<()> {
+        let link = format!("{}/{}", root_url, &self.url_path);
+        let mut url = sitemap::structs::UrlEntry::builder();
+        url = url.loc(link);
+        if let Some(date) = self.front.published_date {
+            url = url.lastmod(*date);
```
epage: If/when we add a modified_date, we'll need to remember to update this

I think the only issue related to the topic is https://github.com/cobalt-org/cobalt.rs/issues/197

src/cobalt_model/config.rs
epage: Ok, this was previously needed only for RSS, so we just avoided sending along dead data.

tglman:
Hi, 
now should be all fine.

Bye

PR #711, opened 2020-01-26
refactor(serve): Trade actix for tiny_http

Geobert:
Move to tinyhttp


src/bin/cobalt/serve.rs
```diff
@@ -78,6 +77,69 @@ pub fn serve_command(matches: &clap::ArgMatches) -> Result<()> {
...
+    info!("Ctrl-c to stop the server");
+
+    // attempts to create a server
+    let server = match Server::http(ip) {
+        Ok(server) => server,
+        Err(e) => {
+            error!("{}", e);
+            process::exit(1);
+        }
+    };
```
epage: There a reason you don't propagate this up?

src/bin/cobalt/serve.rs
```diff
@@ -78,6 +77,69 @@ pub fn serve_command(matches: &clap::ArgMatches) -> Result<()> {
...
+
+    for request in server.incoming_requests() {
+        if let Err(e) = static_file_handler(&dest, request) {
+            error!("{}", e);
+            process::exit(1);
+        }
+    }
```
epage: ditto about propagating up

epage:
For some reason, the pipeline results aren't coming up this morning.  iirc the goal is compile times.  Any idea how much this improved things?

src/bin/cobalt/serve.rs
Geobert: I don't see any, especially as we process::exit. I think it was like this in cobalt before? I can change it to remove `error!()` :)

Geobert:
Thank you for reviewing that :)

> For some reason, the pipeline results aren't coming up this morning. iirc the goal is compile times. Any idea how much this improved things?

Damned, I was counting on the pipelines… ^^' I'll bench that tonight :)

src/bin/cobalt/serve.rs
epage: The existing process exit is pretty much in response to ctrl-c since we don't need to print extra context and due to how some of the error reporting works..  In some of this code, errors might crop up for other reasons and those are ones we should propagate up to provide additional context.

Geobert: So should I left it there?

epage: I'd recommend getting rid of the exist's and see where that takes us

Geobert:
Bench!

command used (Powershell) `Measure-Command { cargo build --features "pagination-unstable" | Out-Host }`:

# Tiny HTTP
Ticks             : 1469759398
TotalMinutes      : 2.44959899666667
TotalSeconds      : 146.9759398
TotalMilliseconds : 146975.9398

Ticks             : 1432811547
TotalMinutes      : 2.388019245
TotalSeconds      : 143.2811547
TotalMilliseconds : 143281.1547

Ticks             : 1427187356
TotalMinutes      : 2.37864559333333
TotalSeconds      : 142.7187356
TotalMilliseconds : 142718.7356

# Actix
Ticks             : 1712280069
TotalMinutes      : 2.853800115
TotalSeconds      : 171.2280069
TotalMilliseconds : 171228.0069

Ticks             : 1686540462
TotalMinutes      : 2.81090077
TotalSeconds      : 168.6540462
TotalMilliseconds : 168654.0462

Ticks             : 1685488431
TotalMinutes      : 2.809147385
TotalSeconds      : 168.5488431
TotalMilliseconds : 168548.8431

20ish seconds less with tiny_http :)

src/bin/cobalt/serve.rs
epage: Oh, you are right that this should ignore errors because we want to keep running in case updating the site failed once

Geobert:
Ok, not even try to compile before pushing… #tired

src/bin/cobalt/serve.rs
epage: However, if failing to create the HTTP server fails, we should exit out of the program.  We should do this with `?` so the error unwinds the stack and all of our context gets added as we do so.

Geobert: It's giving me hard time as the error is a Box<dyn std::error::Error + Send + Sync>, I can't just do `let server = Server::http(ip)?;` I need to do a `map_err` of some kind

epage:
Thanks! Once CI is passing, I'll merge it!

PR #725, opened 2020-02-11
fix: Do not render `Nil` excerpt

### LINKED ISSUE #724: Having full page contents in RSS feeds is not possible
As described in [this comment on #309](https://github.com/cobalt-org/cobalt.rs/issues/309#issuecomment-584806050), and confirmed by @epage in a reply, setting `excerpt_separator` to the empty string is intended to cause the full page contents appear in the RSS feed, but it currently leads to the feed's item contents being empty.

rotty:
Setting `excerpt_separator` to the empty string is intended to disable
auto-extracted excerpts, and, as a side-effect, cause the full content
to be used in the RSS feed.

Before this change, this did not happen, as
`Document::description_to_str` would render the `Nil` value inserted
into the "excerpt" attribute in all cases.

Fixes #724.

### DIFF
--- a/src/document.rs
+++ b/src/document.rs
@@ -284,9 +284,13 @@ impl Document {
             .description
             .clone()
             .or_else(|| {
-                self.attributes
-                    .get("excerpt")
-                    .map(|s| s.render().to_string())
+                self.attributes.get("excerpt").and_then(|excerpt| {
+                    if excerpt.is_nil() {
+                        None
+                    } else {
+                        Some(excerpt.render().to_string())
+                    }
+                })
             })
             .or_else(|| {
                 self.attributes
@@ -317,7 +321,12 @@ impl Document {
         Ok(html.to_owned())
     }
 
-    /// Renders excerpt and adds it to attributes of the document.
+    /// Renders the excerpt and adds it to attributes of the document.
+    ///
+    /// The excerpt is either taken from the `excerpt` frontmatter setting, if
+    /// given, or extracted from the content, if `excerpt_separator` is not
+    /// empty. When neither condition applies, the excerpt is set to the `Nil`
+    /// value.
     pub fn render_excerpt(
         &mut self,
         globals: &Object,


rotty:
Hmm, it seems my commit message does not fit the style guidelines required by CI, but I cannot find anything in the contribution guidelines on that topic.

epage:
> Hmm, it seems my commit message does not fit the style guidelines required by CI, but I cannot find anything in the contribution guidelines on that topic.

Sorry about that; I forgot to update the guide.  In part, it is guidance only anyways.  cobalt follows [conventional style](https://www.conventionalcommits.org/en/v1.0.0).  If people aren't comfortable rewriting their history, I go ahead and do it by squashing the the change so I can use it for changelog generation.

src/document.rs
```diff
@@ -328,7 +332,7 @@ impl Document {
             let excerpt = self.render_html(excerpt_str, globals, parser, markdown)?;
             Value::scalar(excerpt)
         } else if self.front.excerpt_separator.is_empty() {
-            Value::Nil
+            return Ok(());
```
epage: When writing code, I tend to prefer people making conditionals off of the value of a variable rather than the variable existing or not (like bash or make).  I think my intention of making this `Nil` was so that a `{{ page.excerpt }}` variable would always exist but it would just be `nil`.

The other downside to this route is it is a breaking change.

An alternative is for us to instead change the code that is reading this variable and check `value.is_nil()`.

epage:
Thanks for resolving this!

src/document.rs
rotty: > When writing code, I tend to prefer people making conditionals off of the value of a variable rather than the variable existing or not (like bash or make).
I see, I didn't even think of that being exposed as a variable to the user, to be honest -- I just considered the Rust "layer" of things.

> I think my intention of making this `Nil` was so that a `{{ page.excerpt }}` variable would always exist but it would just be `nil`.
> 
> The other downside to this route is it is a breaking change.
> 
> An alternative is for us to instead change the code that is reading this variable and check `value.is_nil()`.

Sounds like the reasonable solution, then.

epage: > I see, I didn't even think of that being exposed as a variable to the user, to be honest -- I just considered the Rust "layer" of things.

I'm a little rough on this code since looking at a bunch of other things since then but I believe this is populating the variables available for a page in liquid.  We just happen to reuse it within Rust.  The hint towards this is that it is using a `liquid::Value` rather than keeping things more rust-native (Option<String>).

rotty:
> > Hmm, it seems my commit message does not fit the style guidelines required by CI, but I cannot find anything in the contribution guidelines on that topic.
> 
> Sorry about that; I forgot to update the guide. In part, it is guidance only anyways. cobalt follows [conventional style](https://www.conventionalcommits.org/en/v1.0.0). If people aren't comfortable rewriting their history, I go ahead and do it by squashing the the change so I can use it for changelog generation.
>
I'm very comfortable with rewriting my history, but not used to that convention. From glancing at the spec, it seems my commit should now be valid, but CI still complains:

> Invalid commit format Cannot have sections past body+trailer: ```Fixes #724.```

From my understanding, taking a quick look at the spec, the `Fixes` line should match as footer:

> One or more footers MAY be provided one blank line after the body. Each footer MUST consist of a word token, followed by either a :<space> or <space># separator, followed by a string value.

BTW, the term "trailer", although used in the error message, is not even mentioned in the spec.

rotty:
> > Invalid commit format Cannot have sections past body+trailer: `Fixes #724.`
> 
> From my understanding, taking a quick look at the spec, the `Fixes` line should match as footer:
> 
> > One or more footers MAY be provided one blank line after the body. Each footer MUST consist of a word token, followed by either a `:<space>` or `<space>#` separator, followed by a string value.
> 
> BTW, the term "trailer", although used in the error message, is not even mentioned in the spec.
>
I've now taken an, admittedly quite brief, look at the code of `crate-ci/comitted`, and I suspect a bug: the code seems to introduce the concept of "sections", which does not exist in the spec. I suspect my second _paragraph_ is tripping it up; quoting the spec:

> 7. A commit body is free-form and MAY consist of any number of newline separated paragraphs.

I guess I should file a bug against `committed`. I must admit I'm a bit baffled, as the code as-is seems to prevent people from writing lengthier commit messages containing multiple paragraphs, which I deem detrimental to the purported job of `committed`, namely improving commit message quality.

epage:
committed is using an older verison of the spec. I've been waiting for the `conventional` crate to release to just switch to that.

epage:
Thanks! and sorry for the `committed` issues

rotty:
> committed is using an older verison of the spec. I've been waiting for the `conventional` crate to release to just switch to that.
>
I see -- well, I'm glad they changed the spec. For my personal projects, and at work, I'd find the policy currently implemented in `committed` (and I guess hence the older spec version) unbearable :wink: .

> Thanks! and sorry for the committed issues
>
No problemo -- thank you for your quick and helpful responses!

Issue #726, opened 2020-02-12
Suggestion: enable more clippy warnings


rotty:
When running my usual clippy invocation, as shown below, I noticed it spits out quite a few suggestions for improvements, which, at a first glance, would improve code quality a small bit, if applied. I would volunteer to provide a PR that changes the CI configuration accordingly, after fixing all warnings.

```
cargo clippy --all-targets --all-features -- -D warnings
```

epage:
That'd be great, thanks!

PR #727, opened 2020-02-12
Get the code clippy-clean

rotty:
Fixes: #726

----

Note that while the code is now warning-free with current beta clippy,
the CI is not switched to beta -- that is left as an exercise to the
reader, if deemed desirable,


epage:
Thanks!

PR #783, opened 2020-07-26
fix(serve): added content type to 404 response

JosephLing:
Added content type to 404 respsonse so that the html is rendered as previously it was just handling it as raw text.


epage:
Thanks!

PR #854, opened 2021-05-05
feat: Minify html,css,js

tglman:
- Minifies rendered documents
- Minifies js, and css assets

Minification is disabled by default, for backward and text compatibility.  This can be changed on an extension-by-extension basis.

This doesn't touch the unused config crate.

### DIFF
--- a/src/cobalt_model/assets.rs
+++ b/src/cobalt_model/assets.rs
@@ -69,7 +69,7 @@ impl Assets {
             let dest_path = dest.join(rel_src);
             if sass::is_sass_file(file_path.as_path()) {
                 self.sass
-                    .compile_file(self.source(), dest, file_path.as_path())?;
+                    .compile_file(self.source(), dest, file_path.as_path(), minify)?;
             } else if file_path.extension() == Some(OsStr::new("js")) {
                 copy_and_minify_js(file_path.as_path(), dest_path.as_path(), minify.js)?;
             } else if file_path.extension() == Some(OsStr::new("css")) {
@@ -111,10 +111,6 @@ fn copy_and_minify_css(src_file: &path::Path, dest_file: &path::Path, minify: bo
 
 #[cfg(feature = "html-minifier")]
 fn copy_and_minify_js(src_file: &path::Path, dest_file: &path::Path, minify: bool) -> Result<()> {
-    debug!(
-        "pre Copying and minifying {:?} to {:?}",
-        src_file, dest_file
-    );
     if minify {
         use html_minifier::js::minify;
         // create target directories if any exist
--- a/src/cobalt_model/assets.rs
+++ b/src/cobalt_model/assets.rs
@@ -4,6 +4,8 @@ use super::files;
 use super::sass;
 
 use crate::error::*;
+use failure::ResultExt;
+use std::ffi::OsStr;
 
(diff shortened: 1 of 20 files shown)


epage:
Thanks for doing this!

Looks like something needs to be resolved with the tests.

I see there are three commits.  Were these written for being preserved or would you want me to squash the PR?  

As for the `committed` failure, consider that more of a warning than an error.  I can help with resolving it, resolve it at merge, or we can leave it.

> Also i noticed that there is a nested crate that duplicate some configurations, is it needed to update that crate as well ?

... I guess I got interrupted in the middle of that work.  Don't worry about it.  I am taking a break from cobalt to wrap up another project so I can focus on it but after that, I'm working on a re-architecting of cobalt to make it easier to add features.

src/document.rs
```diff
@@ -18,8 +18,30 @@ use crate::cobalt_model;
...
+}
+
+#[cfg(not(feature = "html-minifier"))]
+fn minify_if_enabled(html: String, _context: &RenderContex) -> Result<String> {
+    Ok(html)
+}
+
+#[cfg(feature = "html-minifier")]
+fn minify_if_enabled(html: String, context: &RenderContex) -> Result<String> {
+    if context.minify.html {
```
epage, line 38: We might render other types of data than html

Ideas:
- Limit this it `.html` extension
- Switch control for this over to the frontmatter.  We'd then move the other minification options to be under `assets` in the config.

Thoughts?

src/cobalt_model/assets.rs
```diff
@@ -55,22 +57,89 @@ impl Assets {
...
+    fn populate_path(&self, dest: &path::Path, minify: &Minify) -> Result<()> {
         for file_path in self.files() {
+            let rel_src = file_path
+                .strip_prefix(self.source())
+                .expect("file was found under the root");
+            let dest_path = dest.join(rel_src);
             if sass::is_sass_file(file_path.as_path()) {
-                self.sass.compile_file(self.source(), dest, file_path)?;
+                self.sass
+                    .compile_file(self.source(), dest, file_path.as_path())?;
```
epage, line 72: Should we minify the SASS output?

epage:
Sorry I didn't notice these two questions before; it was when writing up a description for this that I double checked how things worked and had these questions.

src/document.rs
tglman: I see, yes i will try to add a double check that the final asset is actually an html file.

src/cobalt_model/assets.rs
tglman: Yes, will do

tglman:
Hi,

I fixed the issue i had in the tests, but the run keep failing on something that do not seems related to my changes: https://dev.azure.com/cobalt-org/cobalt-org/_build/results?buildId=749&view=logs&j=50044035-f94f-5642-804a-a0db99902e68&t=ab898a7a-9071-5c69-f447-a57931fa134d

For the commit, when I completed the rest i think is OK to squash them in a single commit

epage:
The failure looks like an issue in syntect.  I've opened https://github.com/trishume/syntect/issues/334

epage:
Thanks for the work on this! Do you need me to cut a release for you?

tglman:
Hi, 
No problem, don't need a release, feel free to do it when you feel is appropriate.

Have a good day.

PR #855, opened 2021-05-29
Fix infinite loop in serve command on Windows

AustinWise:
Before this change, the serve command would repeatedly rebuild the website after the first change. This was caused by the the rebuild being triggered when the output files changed. The reason the existing check did not notice that the change files were in the output folder is due to how Rust canonicalizes paths on Windows.

Most Windows are familiar with absolute paths that start with the drive letter. For example:

```
c:\windows\notepad.exe
```

The fully qualified version of this path, returned by `canonicalize()`, is:

```
\\?\c:\windows\notepad.exe
```

The destination path was not canonicalized, while the paths coming from notify.rs were. This prevented the starts_with check from detecting that the changed path was within the destination folder.

I also tested this change on Linux and confirmed that it still rebuilds when the input files change.

Some relevant documentation:
* [Rust issue about this behavior](https://github.com/rust-lang/rust/issues/42869)
* [Documentation about Windows fully qualified paths](https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats#dos-device-paths). These are also sometimes called "UNC-style" paths.

### DIFF
--- a/src/bin/cobalt/serve.rs
+++ b/src/bin/cobalt/serve.rs
@@ -179,6 +179,14 @@ fn watch(config: &cobalt_model::Config) -> Result<()> {
         .canonicalize()
         .with_context(|_| failure::err_msg("Failed in processing source"))?;
 
+    // Also canonicalize the destination folder. In particular for Windows, notify-rs
+    // generates the absolute path by prepending the above source path.
+    // On Windows canonicalize() adds a \\?\ to the start of the path.
+    let destination = config
+        .destination
+        .canonicalize()
+        .with_context(|_| failure::err_msg("Failed to canonicalize destination folder"))?;
+
     let (tx, rx) = channel();
     let mut watcher = notify::watcher(tx, time::Duration::from_secs(1))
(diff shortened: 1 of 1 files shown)


epage:
Thanks!

epage:
Would you like me to cut a release with this change or do you use cobalt from HEAD?

AustinWise:
@epage I'm already using my own build, so there is no hurry to cut a release.

Thanks for making this Cobalt program. It was pretty easy to migrate from Jekyll and it is so much faster. It is also much easier to install: no more fiddling with RVM or Docker.

epage:
> Thanks for making this Cobalt program. It was pretty easy to migrate from Jekyll and it is so much faster. It is also much easier to install: no more fiddling with RVM or Docker.

As A heads up, development slowed because of both real life and I was spread too thin.  For the last while, I was focused on wrapping up the distractions and I'm nearly complete.  I am turning my attention back to Cobalt.  My initial focus is on a re-architecture which should both make it easier to add new features (its a pain right now) and even faster, including `cobalt serve` doing partial rebuilds.

PR #856, opened 2021-05-29
Remove duplicate XML declaration in RSS file.

AustinWise:
The extra XML declaration prevents the XML from being parsed.

### DIFF
--- a/src/cobalt.rs
+++ b/src/cobalt.rs
@@ -433,7 +433,6 @@ fn create_rss(
     }
 
     let mut rss_file = fs::File::create(&rss_path)?;
-    rss_file.write_all(br#"<?xml version="1.0" encoding="UTF-8"?>"#)?;
     rss_file.write_all(&rss_string.into_bytes())?;
     rss_file.write_all(b"\n")?;
 


epage:
Huh, I wonder if the RSS crate used change its behavior and we missed that

AustinWise:
@epage It looks like it was rust-syndication/rss#113 which made it into version 1.10.0 of the crate.

Despite the [Cargo.toml having v1.8](https://github.com/cobalt-org/cobalt.rs/blob/d905914556db619e83f2779fcbcafd0e234a5de6/Cargo.toml#L43), the [Cargo.lock file has 1.10.0](https://github.com/cobalt-org/cobalt.rs/blob/d905914556db619e83f2779fcbcafd0e234a5de6/Cargo.lock#L1421).

epage:
> Despite the Cargo.toml having v1.8, the Cargo.lock file has 1.10.0.

Yeah, `Cargo.toml` just says "anything compatible with at least this version, so it turns into `>=1.8, <2.0`

Interesting that they considered adding that to be backwards compatible.  I wouldn't have...

PR #866, opened 2021-06-07
Add vimwiki support to cobalt

### LINKED ISSUE #817: Any interest in a contribution to add vimwiki as a source format?
As the title implies, curious if there's any interest in supporting other source formats other than markdown. I'd write the contribution myself, so no need to contribute cycles beyond review.

I saw https://github.com/cobalt-org/cobalt.rs/issues/272 from three years back mentioned the enum for source formats, which I see is used here for raw/markdown: https://github.com/cobalt-org/cobalt.rs/blob/master/src/document.rs#L301

### vimwiki

I've been working on [vimwiki-rs](https://github.com/chipsenkbeil/vimwiki-rs) (runtime & compile-time parsing of the vimwiki language) for a little while in conjunction with formally defining a specification for the language over at https://github.com/vimwiki/vimwiki/pull/1025, which just merged. The language has single and multi-line comments, which is where I'd stuff in metadata.

I've got on my list to support an export to typed html and/or html string, but not sure if there would be interest in supporting that with cobalt (maybe behind a feature gate).

My interest is enabling more users, including myself, to have the option to use vimwiki as a markup language given I already use it for note taking.

chipsenkbeil:
Resolves #817.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -40,6 +40,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "33954243bd79057c2de7338850b85983a44588021f8a5fee574a8888c6de4344"
 
 [[package]]
+name = "arrayvec"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "23b62fc65de8e4e7f52534fb52b0f3ed04746ae267519eef2a83941e8085068b"
+
+[[package]]
 name = "ascii"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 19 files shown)


chipsenkbeil:
@epage this worked until I extended the test to include a vimwiki code block:

```
{{{python
def my_func():
    print('hello world!')
}}}
```

This causes liquid to panic when parsing the wiki page. If you remove the code block from the test `index.wiki` file, it builds and outputs html just fine.

You'd mentioned there was a way to turn off liquid, but I couldn't figure it out from documentation or source code as it looks like the liquid parser is executed no matter what.

epage:
> This causes liquid to panic when parsing the wiki page. If you remove the code block from the test index.wiki file, it builds and outputs html just fine.

Can you open an issue in the `liquid` crate?

> You'd mentioned there was a way to turn off liquid, but I couldn't figure it out from documentation or source code as it looks like the liquid parser is executed no matter what.

Oh, I remembered you could control `md` and I planned to give control over `liquid`, but it looks like that hasn't happened yet.  I'll try to get tat implemented soon.

chipsenkbeil:
@epage the build is failing because of a job called "commit" with an invalid format. Is this related to my commit message? I don't see a recommendation of what to change to fix it. Tests, clippy, rustfmt, etc. all look fine, though 😄 

image

epage:
> @epage the build is failing because of a job called "commit" with an invalid format. Is this related to my commit message? I don't see a recommendation of what to change to fix it. Tests, clippy, rustfmt, etc. all look fine, though smile
> 
> image

Whoops, somehow cobalt never got my upgrade to the CI pipeline that explains everytjhing.

Basically, we follow Conventional Commit.  However, we do not want that to be a burden for contribution and can resolve things on your behalf.  I'll just resolve this now with a squash

PR #941, opened 2021-12-04
cobalt_model - Add a `time` to Site

elpiel:
This is the same option that is present in jekyll (as a global variable).

I'm rewriting the areweinspaceyet.org with rust and I want to keep it.
In a nutshell, it shows when was the last time the website was generated (updated).

https://github.com/AeroRust/are-we-in-space-yet/blob/933b45db16c2585f68d2ce0c9a03640ad6604366/_includes/footer.html#L19

### DIFF
--- a/src/cobalt_model/site.rs
+++ b/src/cobalt_model/site.rs
@@ -53,7 +53,7 @@ impl Site {
             sitemap,
             data,
             data_dir,
-            time: DateTime::now()
+            time: DateTime::now(),
         }
     }
 
@@ -77,10 +77,7 @@ impl Site {
                 liquid::model::Value::scalar(kstring::KString::from_ref(base_url)),
             );
         }
-        attributes.insert(
-            "time".into(),
-            liquid::model::Value::scalar(self.time),
-        );
+        attributes.insert("time".into(), liquid::model::Value::scalar(self.time));
 
         let mut data = self.data.clone().unwrap_or_default();
         let data_path = source.join(&self.data_dir);
--- a/src/cobalt_model/site.rs
+++ b/src/cobalt_model/site.rs
@@ -79,7 +79,7 @@ impl Site {
         }
         attributes.insert(
             "time".into(),
-            liquid::model::Value::scalar(kstring::KString::from_ref(&self.time.to_string())),
+            liquid::model::Value::scalar(self.time),
         );
 
         let mut data = self.data.clone().unwrap_or_default();
--- a/src/cobalt_model/site.rs
+++ b/src/cobalt_model/site.rs
@@ -2,6 +2,7 @@ use std::ffi::OsStr;
 use std::fs;
 use std::path;
 
+use cobalt_config::DateTime;
 use failure::ResultExt;
 use liquid;
 use serde_json;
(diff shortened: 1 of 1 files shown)


src/cobalt_model/site.rs
```diff
@@ -73,6 +77,11 @@ impl Site {
                 liquid::model::Value::scalar(kstring::KString::from_ref(base_url)),
             );
         }
+        attributes.insert(
+            "time".into(),
+            liquid::model::Value::scalar(kstring::KString::from_ref(&self.time.to_string())),
```
epage, line 82: There a reason you are inserting it as a string rather than doing `liquid::model::Value::scalar(self.time.clone())`?

src/cobalt_model/site.rs
```diff
@@ -73,6 +77,11 @@ impl Site {
                 liquid::model::Value::scalar(kstring::KString::from_ref(base_url)),
             );
         }
+        attributes.insert(
+            "time".into(),
```
epage, line 81: Could you expand in your PR on your use case for wanting this?

epage (changes requested):
Thanks for helping make cobalt better!  Sorry if it might seem quiet lately, I've been a bit busy with work on some other crates.

src/cobalt_model/site.rs
elpiel: Jekyll has this global variable.
I'm rewriting the areweinspaceyet.org with rust and I want to keep.
In a nutshell it shows when was the last time the website was generated (updated)

https://github.com/AeroRust/are-we-in-space-yet/blob/933b45db16c2585f68d2ce0c9a03640ad6604366/_includes/footer.html#L19

epage: Thanks!  I personally find it inaccurate, update is different than re-generate, but I can understand just wanting to carry something forward.

Could you put some of this in the PR description?

elpiel: I've added a description :+1:

src/cobalt_model/site.rs
elpiel: There isn't :raised_hands: 
Fixed.

elpiel:
Inside the template, you can use this snippet to show the date on which the site was built?
I haven't checked how `serve` behaves though and might be good to add this to the docs.

```
{{ site.time | date: '%Y-%m-%d'}}
```

Screenshot from 2021-12-04 21-45-00

epage:
Thanks!

If you want, [this](https://github.com/cobalt-org/cobalt-org.github.io/blob/source/docs/variables.md) would be the page to add it to

epage:
For the lint commit failure, as it says, don't sweat it.  Thats to encourage things in a format that makes maintenance more streamlined but isn't required.

epage:
(sorry, forgot to run the CI before)

epage (approved):

elpiel:
> For the lint commit failure, as it says, don't sweat it. Thats to encourage things in a format that makes maintenance more streamlined but isn't required.

It's fixed now

anowell:
@epage - [arewelearningyet.com](https://www.arewelearningyet.com/) just [switched to cobalt](https://github.com/anowell/are-we-learning-yet/pull/96) today and took a dependency on `site.time`. While not urgent, we'd love to see a release with this change so that we can [point to a specific release rather than build this particular git ref](https://github.com/anowell/are-we-learning-yet/blob/a388ca643771facb81e948244384a58cc42fe908/.github/workflows/build.yml#L101-L117).

epage:
v0.17.5 is now out!

PR #1051, opened 2022-08-23
Allow importing custom syntaxes

### LINKED ISSUE #200: Syntax highlight vimscript

rdimartino:
Fixes #200

I tried to follow the conversation in that issue about Mutexes vs dependency injection, but a lot has changed since that issue was created so I wasn't sure if that was still relevant.

Also not sure if adding a syntect feature to engarde is a breaking change, or how to handle that exactly.

Thanks!


crates/engarde/src/syntax.rs
```diff
@@ -19,6 +20,12 @@ impl Syntax {
         }
     }
 
+    pub fn load_custom_syntaxes(&mut self, syntaxes_path: &Path) {
```
epage, line 23: Raw needs this function as well so they can be used interchangeably

src/syntax_highlight.rs
```diff
@@ -19,7 +20,14 @@ use engarde::Raw as Highlight;
...
 lazy_static! {
-    static ref HIGHLIGHT: Highlight = Highlight::new();
+    static ref HIGHLIGHT: Highlight = {
+        let mut highlight = Highlight::new();
+        let syntaxes_path = Path::new("./_syntaxes");
+        if syntaxes_path.exists() {
+            highlight.load_custom_syntaxes(syntaxes_path);
+        }
+        highlight
+    };
```
epage, line 30: We're going to need to create this at runtime so we can load the correct root for `_syntaxes`.

Could you do that refactor in a separate PR?

tests/cmd/custom_syntax_highlighting.in/_dest/index.html
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
```
epage, line 1: I'm assuming the `_dest` folder shouldn't be included in the `.in/`

epage (commented):
Thanks for moving this forward!

crates/engarde/src/syntax.rs
rdimartino: Sure! Should that just be a no-op? Or should it log a warning or something?

epage: Let's leave it as a no-op

src/syntax_highlight.rs
rdimartino: Yeah, I'm happy to help out in any way to get this landed, but I'm not quite sure what you mean. I thought `lazy_static` would initialize this at runtime
> Using this macro, it is possible to have statics that require code to be executed at runtime in order to be initialized.

I could use a little more direction for what you're asking.

epage: I went ahead and did it real quick: https://github.com/cobalt-org/cobalt.rs/pull/1052

crates/engarde/src/syntax.rs
```diff
@@ -107,4 +114,24 @@ mod test {
...
+
+    const CUSTOM_CODEBLOCK_RENDERED: &str = "<pre style=\"background-color:#2b303b;\">\n\
+          <span style=\"color:#c0c5ce;\">[[[]]]</span>\
+          <span style=\"background-color:#bf616a;color:#2b303b;\">]</span>\
+          </pre>\n";
+
+    #[test]
+    fn highlight_custom_syntax() {
+        let mut syntax = Syntax::new();
+        let path = Path::new("./tests/fixtures/");
```
epage, line 128: We should probably have this in a folder with a little more context as we'll likely add custom syntax support in the future as well.

src/syntax_highlight.rs
epage: Merged

epage (approved):
Looks great, thanks!

rdimartino:
Not sure why the windows tests failed, but it looks like I forgot to skip the new test when the `syntax-highlight` feature is not enabled, like the others https://github.com/cobalt-org/cobalt.rs/blob/master/tests/cli_tests.rs#L5-L11

Sorry! Fixing now

epage:
> Not sure why the windows tests failed, but it looks like I forgot to skip the new test when the `syntax-highlight` feature is not enabled, like the others https://github.com/cobalt-org/cobalt.rs/blob/master/tests/cli_tests.rs#L5-L11
> 
> Sorry! Fixing now

Sorry I didn't say something before, windows is generally broken.  One of our dependencies makes assumption about the build machine but isn't tested with github actions and won't build.

epage:
Forgot to say, this has been released.

Issue #1081, opened 2022-10-24
Feature request: Atom 1.0 feeds


ChillyCider:
I think Atom is pretty popular, so it might be a cool feature. Maybe enabled on a per-collection basis (#368).

ChillyCider:
I'd be interested in trying my hand at implementing this. Never tried Rust before, but now's as good a time as any

epage:
Go for it!  The rss and jsonfeed implementations should provide a pretty good starting point for this.

PR #1089, opened 2022-11-02
Atom 1.0 support

ChillyCider:
This PR resolves issue #1081 by adding Atom 1.0 feed support.

In order to generate valid Atom feeds, I had to add 2 new frontmatter settings, `updated_date` and `authors`:

```yaml
---
# ...
updated_date: 2022-11-02 03:14:00 +0100
authors:
    - Charlie Murphy
---
```

The Atom spec demands the existence of an "updated" date, while the "published" date is optional -- this seems backwards to me but I won't argue with the spec 😄 . Since `published_date` is more common in cobalt, I've made Atom use it as a fallback. That way, most existing cobalt blogs should be fine.

However, Atom also demands the existence of "authors". For single-author blogs, you may be well-served to stick the author into your _cobalt.yml's "default" section.


crates/config/src/frontmatter.rs
```diff
@@ -21,12 +21,16 @@ pub struct Frontmatter {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub categories: Option<Vec<liquid_core::model::KString>>,
     #[serde(skip_serializing_if = "Option::is_none")]
+    pub authors: Option<Vec<liquid_core::model::KString>>,
```
epage, line 24: Could you split this out into its own Issue / PR?

crates/config/src/frontmatter.rs
```diff
@@ -21,12 +21,16 @@ pub struct Frontmatter {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub tags: Option<Vec<liquid_core::model::KString>>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub excerpt_separator: Option<liquid_core::model::KString>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub published_date: Option<DateTime>,
     #[serde(skip_serializing_if = "Option::is_none")]
+    pub updated_date: Option<DateTime>,
```
epage, line 32: Could you split this out into its own issue / PR?

#197 exists but that is more about automatically inferring it

src/cobalt.rs
```diff
@@ -506,6 +519,60 @@ fn create_jsonfeed(
...
+        }])
+        .title(atom_syndication::Text::plain(title.to_string()))
+        .entries(entries)
+        .build();
+
+    // create target directories if any exist
+    if let Some(parent) = path.parent() {
+        fs::create_dir_all(parent)
+            .with_context(|_| failure::format_err!("Could not create {}", parent.display()))?;
+    }
```
epage, line 568: The other ones don't have this.  Is there a reason to do it to one vs all?

src/cobalt.rs
```diff
@@ -2,13 +2,16 @@ use std::collections::HashMap;
...
+use std::time::SystemTime;
 
 use failure::ResultExt;
 use jsonfeed::Feed;
 use log::debug;
 use log::trace;
 use log::warn;
 use sitemap::writer::SiteMapWriter;
+use vimwiki::vendor::chrono::DateTime;
+use vimwiki::vendor::chrono::Utc;
```
epage, line 14: We shouldn't be relying on vimwiki like this.  If we need chrono, we should use chrono.

tests/cmd/atom.md
```diff
@@ -0,0 +1,16 @@
+```console
+$ cobalt -v build --destination _dest
```
epage, line 2: We need `.out` directories made so we can snapshot the filesystem.

`TRYCMD=dump` will dump to a directory.  You can then copy the output over into the test

epage (commented):
This includes some parts that are new features outside of atomic support.  We should make sure we are on the same page for the requirements in an issue and then do a PR for them.

For what atom support doesn't directly depend on, let's not have atom block on it but then have another PR to take advantage of it

ChillyCider:
@epage Hey thanks for reviewing my PR. I actually ended up using Jekyll to fix my issue, and thereby lost the driving need to add this to Cobalt. 😆 
I'll close this for now, good luck! I have high hopes for Cobalt.

PR #1092, opened 2022-11-06
fix: update dependencies and fixed relative compilation problems

tglman:


Cargo.lock
```diff
@@ -1012,18 +995,18 @@ dependencies = [
...
-checksum = "b8e7479fa1ef38eb49fb6a42c426be515df2d063f06cb8efd3e50af073dbc26c"
+checksum = "15315cfa9503e9aa85a477138eff76a1b203a430703548052c330b69d8d8c205"
 dependencies = [
  "utf8-width",
 ]
 
 [[package]]
 name = "html-minifier"
-version = "3.0.15"
+version = "3.0.17"
```
epage, line 1007: I'd like to hold off on doing this upgrade as it is a breaking change and should be yanked, see https://github.com/magiclen/html-minifier/issues/13

epage:
btw what was your goal in updating the dependencies?  Was there something you were wanting form one of the upgrades?

tglman:
Hi,

Nope I just saw the issue while installing and I fixed it, but I guess the right solution would be the yank that release.

Regards

epage:
`cargo install --locked` will ensure you use the versions we test

PR #1134, opened 2023-05-13
feat: allow multiple date formats in frontmatter published date

tglman:
Hi,

As the description say, this allow to have multiple date formats for the published_date in the front matter, and still use the the liquid date time parsing, I started this implementation because of a regression on the liquid serde implementation of DateTime, if you `cargo update` the dependencies existing sites may fail to parse because the serde deserialization of DateTime require now the milliseconds, I'm going to open an issue on liquid for this(if not already there).


epage:
Is there a reason we shouldn't fix this in liquid instead?

tglman:
Hi, 

I just opened a PR to fix this in liquid: https://github.com/cobalt-org/liquid-rust/pull/500, that will fix the regression anyway this PR also include the support of multiple date formats, like date formats without time, which I think it does make sense in term of blog posts.

Bye

PR #1142, opened 2023-08-20
feat(file-serve)!: Allow graceful shutdown of server

orzogc:
Assume that someone was developing an App which would serve some files temporarily, the server should be closed when it was no longer needed. But the server was just blocked for handling incoming requests, there were no valid methods to close it.

So a method for graceful shutdown of the server is introduced in this PR, also a method for checking whether a sever is running.

BREAKING CHANGE: `Server.serve()` returns `Result<(), Error>` instead of `Result<Infallible, Error>` because the server can be closed.
Remove `Debug` trait of `Server`.

### DIFF
--- a/crates/file-serve/src/lib.rs
+++ b/crates/file-serve/src/lib.rs
@@ -18,7 +18,10 @@
 
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
 
-use std::str::FromStr;
+use std::{
+    str::FromStr,
+    sync::{RwLock, TryLockError},
+};
 
 /// Custom server settings
 #[derive(Clone, Debug, PartialEq, Eq)]
@@ -66,19 +69,20 @@ impl ServerBuilder {
         Server {
             source,
             addr: format!("{}:{}", hostname, port),
+            server: RwLock::new(None),
         }
     }
 
     /// Start the webserver
-    pub fn serve(&self) -> Result<std::convert::Infallible, Error> {
+    pub fn serve(&self) -> Result<(), Error> {
         self.build().serve()
     }
 }
 
-#[derive(Debug)]
 pub struct Server {
     source: std::path::PathBuf,
     addr: String,
+    server: RwLock<Option<tiny_http::Server>>,
 }
 
 impl Server {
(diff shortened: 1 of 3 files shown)


crates/file-serve/src/lib.rs
```diff
@@ -100,17 +105,45 @@ impl Server {
...
+            return Err(Error::new("the server is running"));
+        }
+
         // attempts to create a server
         let server = tiny_http::Server::http(self.addr()).map_err(Error::new)?;
-
-        for request in server.incoming_requests() {
-            if let Err(e) = static_file_handler(self.source(), request) {
-                log::error!("{}", e);
+        self.running.store(true, Ordering::Release);
```
epage, line 121: Shouldn't these use a compare-exchange?

Technically, the new API allows for multiple threads can call `serve` and there will be a race that would cause multiple to get past the first check (and then fail when they try to bind).

crates/file-serve/src/lib.rs
```diff
@@ -100,17 +105,45 @@ impl Server {
...
     }
 
+    /// Whether the server is running
+    pub fn is_running(&self) -> bool {
+        self.running.load(Ordering::Acquire)
+    }
+
     /// Start the webserver
-    pub fn serve(&self) -> Result<std::convert::Infallible, Error> {
+    pub fn serve(&self) -> Result<(), Error> {
```
epage, line 114: The commit title has `!` but can you add a `BREAKING CHANGE:` section that summarizes what part of the API this breaks?

crates/file-serve/src/lib.rs
```diff
@@ -100,17 +105,45 @@ impl Server {
         self.addr.as_str()
     }
 
+    /// Whether the server is running
```
epage, line 108: "Whether the server was running at the instant the call happened"

crates/file-serve/src/lib.rs
epage, line 114: Could you add a comment in the PR/commit explaining the use case you have for this?

Generally, I even prefer the use cases to be discussed in an issue before we get to a PR

epage: > Assume that someone was developing an App

I can assume that from the code.  I was hoping for something more concrete but oh well

crates/file-serve/src/lib.rs
```diff
@@ -100,17 +105,54 @@ impl Server {
...
+        let server = match tiny_http::Server::http(self.addr()) {
+            Ok(server) => server,
+            Err(e) => {
+                self.running.store(false, Ordering::Release);
+                return Err(Error::new(e));
+            }
+        };
+
+        while self.running.load(Ordering::Acquire) {
+            match server.try_recv() {
```
epage, line 133: This is changing `serve` from letting the process sleep between invocations (I presume) to being in a busy-wait.

We could put sleeps in here but that just adds latency while still being in a slightly less bad busy-wait.

tiny_http has [unblock](https://docs.rs/tiny_http/latest/tiny_http/struct.Server.html#method.unblock)  to help with cancellations though that will likely require a bit bigger of an API change to support.

*(sorry, somehow overlooked this in the first pass)*

crates/file-serve/src/lib.rs
```diff
@@ -100,17 +105,54 @@ impl Server {
...
 
-        for request in server.incoming_requests() {
-            if let Err(e) = static_file_handler(self.source(), request) {
-                log::error!("{}", e);
+        // attempts to create a server
+        let server = match tiny_http::Server::http(self.addr()) {
+            Ok(server) => server,
+            Err(e) => {
+                self.running.store(false, Ordering::Release);
+                return Err(Error::new(e));
```
epage, line 128: nit: Since I'm already making another comment, I worry about keeping `self.running` valid through all of the error handling.

One way of dealing with that is to have a `serve_` function with the core logic and `serve` calls it and handles the `store(false)`.  This will ensure it will always be reverted, no matter the code path.  I guess an RAII object might also work

PR #1152, opened 2023-10-03
feat: New option date for new post cli command

tglman:
Hi,

this add a new option for provide a date when create a new post in the cli

Examples:
```
cobalt new "Last post" --date "2023-10-02 10:10:10"
```
```
cobalt new "Last post" --date "now"
```

bye


epage:
Could you expand on why vs just editing the frontmatter directly?

tglman:
Hi, 

yes just for the option `now` that is actually quite useful.

PR #1169, opened 2024-02-23
use / as base url instead of host addr in serve cmd

jcbhmr:
Why? Tunneling! For example GitHub Codespaces will tunnel your port 1234 on your codespace's localhost to https://sometunnel-1234.githubcodespaces.example.org/. if you try to get `http://localhost:1234/style.css` then it fails since its not on localhost; its on a completely separate domain

ex:
```html
<link rel=stylesheet href="{{ site.base_url }}/style.css">
```

This pr (hopefully) would fix that by turning the HTML from this:

```html
<link rel=stylesheet href="http://localhost:1234/style.css">
```
into this which works across localhost and tunnels
```html
<link rel=stylesheet href="/style.css">
```

### DIFF
--- a/src/bin/cobalt/serve.rs
+++ b/src/bin/cobalt/serve.rs
@@ -47,7 +47,7 @@ impl ServeArgs {
 
         let mut config = self.config.load_config()?;
         debug!("Overriding config `site.base_url` with `{}`", server.addr());
-        config.site.base_url = Some(format!("http://{}", server.addr()).into());
+        config.site.base_url = Some("/".into());
         let mut config = cobalt::cobalt_model::Config::from_config(config)?;
         debug!(
             "Overriding config `destination` with `{}`",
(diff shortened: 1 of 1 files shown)


src/bin/cobalt/serve.rs
```diff
@@ -47,7 +47,7 @@ impl ServeArgs {
 
         let mut config = self.config.load_config()?;
         debug!("Overriding config `site.base_url` with `{}`", server.addr());
-        config.site.base_url = Some(format!("http://{}", server.addr()).into());
+        config.site.base_url = Some("/".into());
```
epage, line 50: `debug` statement is no longer correct

epage:
As I don't deal with tunneling, I overlooked this.  As this should still work for local development, I'm fine with this.

epage:
Thanks!

Issue #1168, opened 2024-02-19
Cobalt cannot find layout no matter where it is if src and destination specified in _cobalt.yml


davidthewatson:
I'm using cobalt 0.19.0 built from cargo.

I have a cobalt site generated successfully with the default configuration of _cobalt.yml where src lives in the main project directory and destination is set to _site with layouts in _layouts as in _layouts/default.liquid. That works fine. 

However, if I set src and destination in _cobalt.yml, cobalt seems unable to find the frontmatter-specified layout, no matter where it lives: same current directory as the page or post, the _layouts directory, relative path set in the frontmatter layouts directive, nothing. 

No matter the configuration, I get an error such as:

    Error: Failed to render for file.html

    Caused by:
        Layout default.liquid does not exist (referenced in file.html).

I wish I could find a work-around but so far, no luck. I tried setting verbosity on cobalt build but none of the regex logs seemed causal with respect to what must-be malformed search-path for liquid templates. 

Thanks!

epage:
Have a reproduction case you can point me to?

matschundbrei:
There seems to be an issue with pathing in general. I just had a similar expierience while trying to use a secondary `_cobalt_build.yml` with the gitlab-cicd/pages example.

It does seem that I cannot convince the `source` configuration parameter in any yml outside of the default (`_cobalt.yml`) to just use "." or "this directory"

I am just doing this in order to be able to easily workaround the fact that `site.sitemap` set to anything will prevent `cobalt serve` to even come up.

You can reproduce this by:

```shell
$ cobalt init
Created new project at ./
$ ls
_cobalt.yml  _defaults  index.md  _layouts  posts
$ cp _cobalt.yml _cobalt_testing.yml
$ cobalt serve -c _cobalt_testing.yml 
Building from `` into `/tmp/.tmpvZWt5b`
Watching "" for changes
Serving /tmp/.tmpvZWt5b through static file server
Server Listening on http://localhost:1024
Ctrl-c to stop the server
Error: Failed in processing source

Caused by:
    No such file or directory (os error 2)

```

epage:
@matschundbrei v0.19.3 is out with a fix for that case.  However, without a reproduction case for @davidthewatson's problem, its hard to say if that fixed it or if there are other problems.

matschundbrei:
@epage hey, sorry for the very long delay, this fixed my issue, thanks a lot!

epage:
[...]

PR #1179, opened 2024-04-15
fix(serve): Work with custom config

epage:
This was brought up in #1168


matschundbrei:
Works on my machine ;)

PR #1200, opened 2024-07-24
fix: cobalt serve work correctly with sitemap

tglman:
Hi,

In the last release `cobalt serve` fail if sitemap is enabled, this is because sitemap require a full url and so the required base_url should be full, the serve replace the base_url with a '/' which cause the problem, this just replace the "/" with the current addr of the server.


epage:
Thanks!

PR #1243, opened 2025-04-06
Release AArch64 for macOS instead of x86-64

AustinWise:
For the last several release builds the macOS x86-64 build has failed. See this example:


epage:
Thanks for pointing this out!

I've instead done a separate push as I did a general update to my template for pre-built binaries.

AustinWise:
Thanks, looks good!

PR #1258, opened 2025-09-21
feat: Expose tags as categories in rss

vbfox:
Expose both the tags and categories defined in frontmater in the RSS feed.

Tries to follow [the specification](https://www.rssboard.org/rss-specification#ltcategorygtSubelementOfLtitemgt) by having a <category /> per tag in the frontmater and a single one separated by forward slashed for the hierarchical categories.

In:
```yaml
tags: [test, blogpost]
categories: [blog, test]
```

Out:
```xml
<item>
  <category>test</category>
  <category>blogpost</category>
  <category>blog/test</category>
</item>
```

I didn't make any of theses optional but as you wish.


epage:
Could you move the test additions into a commit before the other two, with the tests passing?  Then as you add support for tags and categories, the test should remain passing.  This makes the test changes in each commit demonstrate the behavior change they cause.

PR #1259, opened 2025-09-21
feat: Add a 'css' theme for syntax highlighting

vbfox:
This theme doesn't specify any color but uses the `ClassedHTMLGenerator` to generate CSS classes instead. This allow to have per-color-scheme themes and to easily edit themes manually.

This is similar to Zola [classed-highlighting](https://www.getzola.org/documentation/content/syntax-highlighting/#inline-vs-classed-highlighting) feature except with a `c-` prefix instead of a `z-` one.

It generates css classes like that (whitespace modified for readibility):

```html
<pre class="language-rust highlighter-syntect">
  <code class="highlight"><span class="c-source c-rust">
    <span class="c-support c-macro c-rust">println!</span>
    <span class="c-meta c-group c-rust">
      <span class="c-punctuation c-section c-group c-begin c-rust">(</span>
    </span>
    <span class="c-meta c-group c-rust">
      <span class="c-string c-quoted c-double c-rust">
        <span class="c-punctuation c-definition c-string c-begin c-rust">&quot;</span>
        Hello World!
        <span class="c-punctuation c-definition c-string c-end c-rust">&quot;</span>
      </span>
    </span>
    <span class="c-meta c-group c-rust">
      <span class="c-punctuation c-section c-group c-end c-rust">)</span>
    </span>
    <span class="c-punctuation c-terminator c-rust">;</span>
  </code>
</pre>
```

### DIFF
--- a/crates/engarde/src/syntax.rs
+++ b/crates/engarde/src/syntax.rs
@@ -3,7 +3,7 @@ use syntect::easy::HighlightLines;
 use syntect::highlighting::ThemeSet;
 use syntect::html::{
     ClassStyle, ClassedHTMLGenerator, IncludeBackground, append_highlighted_html_for_styled_line,
-    start_highlighted_html_snippet,
+    css_for_theme_with_class_style, start_highlighted_html_snippet,
 };
 use syntect::parsing::{SyntaxReference, SyntaxSet};
 use syntect::util::LinesWithEndings;
@@ -34,6 +34,10 @@ impl Syntax {
         self.syntax_set = builder.build();
     }
 
+    pub fn css_theme_name() -> &'static str {
+        CSS_THEME
+    }
+
     pub fn has_theme(&self, name: &str) -> bool {
         name == CSS_THEME || self.theme_set.themes.contains_key(name)
     }
@@ -44,6 +48,16 @@ impl Syntax {
         themes.sort_by_key(|a| a.to_ascii_lowercase());
 
         themes.into_iter()
+    }
+
+    fn css_class_style() -> ClassStyle {
+        ClassStyle::SpacedPrefixed { prefix: "c-" }
+    }
+
+    /// Get the content of a css file to apply the specified color theme when using the 'css' theme
+    pub fn css_for_theme(&self, name: &str) -> String {
+        let theme = &self.theme_set.themes[name];
+        css_for_theme_with_class_style(theme, Self::css_class_style()).unwrap()
     }
 
     pub fn syntaxes(&self) -> impl Iterator<Item = String> + '_ {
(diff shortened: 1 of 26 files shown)


tests/cmd/syntax_highlighting_css.in/_cobalt.yml
```diff
@@ -0,0 +1,2 @@
+syntax_highlight:
```
epage, line 1: Could you add this test case in a previous commit but use a regular theme for it?  Then this commit will switch to the `css` theme and we get to see how the behavior changes

vbfox: https://github.com/cobalt-org/cobalt.rs/commit/36fe05270d3063fe3f900ca0a49b33d19f1929e2#diff-3b6b0a70768bbae583638199ee4a7ff1cab39d68eeb4a75d9339fd2b86956673

crates/engarde/src/syntax.rs
```diff
@@ -46,6 +50,16 @@ impl Syntax {
         themes.into_iter()
     }
 
+    fn css_class_style() -> ClassStyle {
+        ClassStyle::SpacedPrefixed { prefix: "c-" }
+    }
+
+    /** Get the content of a css file to apply the specified color theme when using the 'css' theme */
```
epage, line 57: This should be `///` style comment

(sorry, overlooked this commit previously)

vbfox: Too much switching between languages, this is a Typescript doc comment and not a rust one XD

src/bin/cobalt/debug.rs
```diff
@@ -34,6 +36,19 @@ pub(crate) enum HighlightCommands {
         #[command(flatten, next_help_heading = "Config")]
         config: args::ConfigArgs,
     },
+
+    /// Save the css file for a theme
+    #[cfg(feature = "syntax-highlight")]
+    SaveThemeCss {
```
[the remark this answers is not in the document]
vbfox: I feel that it should be next to the other themes/syntaxes commands, but I would have put the whole `highlight` set of commands at the root as they are generally useful with the documentation even pointing to one of them

image

As for prior art:
* Zola function a bit differently as it has a `highlight_themes_css` that auto-generate the files input, but understand implicitly when edit them that it should not touch them so you can do that. https://www.getzola.org/documentation/content/syntax-highlighting/#inline-vs-classed-highlighting
* Jekyll rouge support custom css theme but doesn't have a command to generate the css
* Elevently and all others using prims use css colors as the base for their themes so there is no generation or conversion.
* Hugo has a `gen` command shared with `hugo gen chromastyles --style=monokai > monokai.css` (The `gen `command also generates man pages for the CLI) https://gohugo.io/commands/hugo_gen/
* Expressive code that has plugins for Astro and other SSGs support VSCode themes directly so they document to use the VSCode command to export a theme https://expressive-code.com/guides/themes/#using-themes-from-vs-code

Not much of a standard in there 🤷🏼‍♂️

